PR #80, opened 2017-07-12
RFC 80: Jsonnet based single file schemas

kevindew:
[This RFC](https://github.com/alphagov/govuk-rfcs/blob/edde212fba0fbe7f35cd57a24611a9fd5c1fc0ea/rfc-080-jsonnet-based-single-file-schemas.md) proposes the use of single file schema definitions that are authored with [Jsonnet](http://jsonnet.org).

A [working implementation](https://github.com/alphagov/govuk-content-schemas/pull/634) of this was developed during the GOV.UK firebreak last week - this proposal is borne out of that spike.

To give some further insight beyond the RFC there some further meta information on alternatives considered below.

I look forward to comments, questions and any suggestions to improve this - closing date is 21st July

# Alternatives to defining data with JSON Schema definitions

### Merging JSON Schema files further

Initially it was hoped that a system could be defined where a schema could be defined through the JSON schema system and extending a base object where required. This had the distinct advantage that a new syntax would not have to be learnt, although at the expense of the user understanding the merging process.

There were a number of reasons though which led to this being rejected:

- difficult to define merge points that didn’t introduce lots of repetition, or were verbose
- the required property within schemas is difficult to merge
- fields can be represented differently in different schemas

### Define a full syntax for schema definitions

Paul Bowsher defined an [approach](https://gist.github.com/boffbowsh/cebdce4e8cda210d348d) to how schemas could be defined. This syntax would then be transposed into JSON Schema.

This would not pursued due to:

- Concern over the scale of the task to sufficiently port features of JSON Schema
- Concern that a user has to understand more about the output of schema than necessary

## This approach compared to alternatives

The decision about how much of the schema definition should embrace JSON Schema is difficult. It feels a bit like you get caught between a rock and a hard place where defining anything distinct introduces a learning curve and using just JSON schema leads to problems.

It could be argued that having the hybrid syntax where JSON schema is embedded is a worst of both worlds scenario. However I think that it offers no greater learning curve than we currently have with JSON Schema and it allows us to embrace most aspects of JSON schema without having to have code written for all scenarios.

# Using YAML instead of Jsonnet

YAML is the most common configuration choice for ruby projects and is a format that is easier to read than Jsonnet. Had there been an import between files feature in YAML there’s a good chance we would have used that instead.

It was considered to try add in a syntax for doing importing with YAML

```
field:
	_import: “other-file.yml”
	field: value
```

This is relatively simple at a basic level. However it becomes very complicated once we consider what should happen if an import itself imports another file. Therefore YAML was only considered as a viable avenue if we were to no longer make usage of the import feature.


fofr:
Responding to @danielroseman and @timblair’s comments on frontend:

Regarding how schemas are used in the frontend I'd like to clarify that the presence of schemas has been beneficial. Having a definition of what a frontend can and can't receive keeps the frontends simple. We've built tooling around this in government-frontend for resiliency: https://github.com/alphagov/government-frontend/pull/294

The problems we encountered regarding schemas on the frontend are:

* Inconsistencies when writing schema, eg:
  - schemas have the same frontend feature but are built in isolation and end up using different definitions
  - schemas with properties of the same name but for different features
* Iterating schemas
* Keeping examples up to date and accurate
* Schemas not defining everything that the frontend needs to render

This was alleviated a little by:
https://github.com/alphagov/govuk-content-schemas/pull/509

kevindew:
There seems to be a general consensus that this proposal is accepted 🎉

@boffbowsh would you be able to review this?

PR #84, opened 2017-10-16
RFC 84: Replace Static by a gem

tijmenb:
👉 [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/bc8ffe85cdf5cdf5005502cba50d5b64237f1b71/rfc-084-frontend-in-a-gem.md)

### DIFF
--- a/rfc-084-frontend-in-a-gem.md
+++ b/rfc-084-frontend-in-a-gem.md
@@ -1,0 +1,81 @@
+# Replace Static by a gem
+
+Summary: This RFC proposes replacing the current system of sharing frontend code (via the Static application) with a system where we use a gem to do the same.
+
+## Background
+
+The system GOV.UK uses to share frontend code works like this: frontend applications use a [gem called Slimmer](https://github.com/alphagov/slimmer). Slimmer provides templates (layouts) and [GOV.UK components][cmps] to the Rails application by fetching a raw ERB template from the [Static application](https://github.com/alphagov/static). The template links to JS and CSS files that are also hosted by Static. This means that frontend changes can be rolled out across GOV.UK by changing and deploying Static. A number of caching layers make this approach performant. Slimmer makes changes to the rendered HTML like [adding meta tags][tags].
+
+[cmps]: https://github.com/alphagov/static/blob/master/doc/static-components.md
+
+## Problems with current approach
+
+1. We can't test if the application actually works correctly with updated layout or components. If a component in Static is updated, we have to test applications to see if they still work. This can be done by rolling out the change on integration and see if the Smokey tests pass, or perform manual testing. Smokey is very high level and doesn’t have much coverage and manual testing is time-consuming and imprecise.
+2. Doesn't allow interaction in tests. To prevent a dependency on a live service in tests, we stub out components in the test environment - they don’t render HTML, but dump the component input onto the page. Tests use this data to make sure that the input to the component is correct. This means we have to trust that the input to the component won’t change, there’s no explicit contract between the app and component.
+3. Caching. There are a number of caching layers for components. This means that when a new version of Static is released, sometimes the HTML changes are still cached, but the new CSS and JS are returned to the user. This can cause unexpected results.
+4. Forces awkward code in Static. Because we ship ERB templates, all logic needs to be embedded in the template. This forces us to [write code that we know goes against best practices][analytics-code] and style guides. To prevent this the [govuk_navigation_helpers gem](https://github.com/alphagov/govuk_navigation_helpers) was created to generate the input for the components.
+5. Considerable learning curve for the architecture. Shipping raw ERB templates is not a well known technique in the industry. This causes a steep learning curve for new developers on GOV.UK.
+
+[analytics-code]: https://github.com/alphagov/static/blob/master/app/views/govuk_component/analytics_meta_tags.raw.html.erb
+
+## Proposal
+
+Each frontend application will depend on 1 gem, `govuk_frontend_foo`. This gem:
+
+- Provides the components with the JS/CSS
+- Provides layouts (using components)
+
+## Trade-offs & costs
+
+There are 2 main costs associated with the proposal:
+
+### Deployment
+
+For each global change to the front end gem, we have to update and deploy all frontend applications.
+
+We can mitigate this by:
+
+- Implementing automated PRs for gem upgrades
+- Add monitoring to prevent version drift
+- Schedule regular weekly deploys
+- Making it easier to deploy
+- Making sure we only put global things into the gem. Things that are only used once can stay in the frontend application
+
+### Duplicated asset downloads
+
+Because there’s no more shared Javascript/CSS, each application will need to bundle everything it needs to render. For example, this means each application will ship their own version of jQuery to the user. This will make it slower and more data-expensive for a user to use the site.
+
+Possible mitigations:
+
+- "Deploy the gem". Every time a new version of the gem is released, we deploy the resulting bundle. The result will live on something like `assets.publishing.services/gem-foo-2.1.0/*`.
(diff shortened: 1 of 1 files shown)


h-lame:
One of the things we have to do a lot in the surveys and support team is deliver surveys that run across multiple pages and frontends on GOV.UK.  This is currently a JS file delivered as a shared asset by static which makes the process easy as it's used by every frontend.  We don't need to know which frontend will render a given page; we only need to make the change in one place and deploy one app to affect the entire site.  This is also something we do a lot, we probably do at least one survey related change (adding surveys, extending surveys, tweaking frequencies, etc...) most weeks.

The workflow for surveys is is a fairly minor consideration in the grand scheme of things, but it would be great if this rfc accommodated that workflow somehow.  Do we imagine keeping any mechanism for shared assets that can be used across all apps without doing an update and release dance on a gem + all the frontend apps?

tijmenb:
@h-lame That's a good point. This also applies to the emergency banner, which we haven't considered in this proposal. We'll have to come up with something for that.

nickcolley:
Can we reduce the scope of this RFC to things that change most often? If we reduced the scope to the components first it might be easier to push this forwards?

tijmenb:
Maybe something like static should stay around, but for very limited purposes. How about each frontend app has both a global & local bundle. A layout would look like:

```erb
<% if Rails.env.production? %>
  <%= javascript_include_tag "https://assets.publishing.service.gov.uk/survey.js" %>
  <%= stylesheet_link_tag "https://assets.publishing.service.gov.uk/survey.css" %>
  <%= javascript_include_tag "https://assets.publishing.service.gov.uk/jquery-and-such-3.2.1.js" %>
  <%= stylesheet_link_tag "https://assets.publishing.service.gov.uk/font-1.2.3.css" %>
<% else %>
  <%= javascript_include_tag "jquery-and-such-3.2.1.js" %>
  <%= stylesheet_link_tag "font-1.2.3.css" %>
<% end %>

<%= javascript_include_tag "application.js" %>
<%= stylesheet_link_tag "application.css" %>
```

This would allow us to keep using the survey as we do currently and solves a large part of the problem with bundle size.

nickcolley:
I feel really uncomfortable giving a 👍  to any approach that does not remove as much duplication as possible, if it's hard we should go the extra mile because it'll impact our users for our own gain.

fofr:
Following on from Nick's point, we have a story for looking at moving components from static to the govuk_publishing_components gem:
https://trello.com/c/q7x0jvzK/62-make-govukpublishingcomponents-the-home-of-components-shared-between-apps

> Put a single component in govuk_publishing_components so that it can be shared between apps. A good starting point would be the lead paragraph component.
>
> Components can already be shared using static but this has risks:
>
> * changes must be backwards compatible
> * caching layers mean that old HTML must be compatible with new CSS (https://github.com/alphagov/static/issues/1121)
> * component helpers can't easily be shared from static, leading to more logic in the view
> * components can't be easily iterated when they live in static

Even with this approach, we use the analytics component to role out small analytics tweaks across all apps, much like the survey changes. That said, this template is a good example of the limits of components without helpers in static:

https://github.com/alphagov/static/blob/master/app/views/govuk_component/analytics_meta_tags.raw.html.erb

andysellick:
Just to record a bit of technical debt... we've built a new component called a task list, which is designed to both sit on its own page in Collections but also in the sidebar of a wide variety of pages across GOV.UK. It was originally going to be a variation on the accordion component but due to this overall problem we decided to keep the accordion in Collections as is and copy and modify the code to a new task list component in static.

Although the task list component has considerable changes from the accordion, the basic expand/collapse functionality is common code that could be built and then reused in both components. The outcome of the task list work is still ongoing and further iterations of the component are possible but still worth recording here for future reference. PR for the component is here: https://github.com/alphagov/static/pull/1173

rubenarakelyan:
How about having new HTML/CSS/JS deployed to asset-manager as versioned files and then having a "slimmer/static"-equivalent gem for each app that points to a specific version of the files? Then you can have different apps with different gem versions that point at different versions of assets but you don't have to co-ordinate breaking changes across multiple frontend apps - their gem versions can be upgraded one-by-one. Apps that share the same gem version will point at the same files which will then be cached by browsers.

cbaines:
> Just to record a bit of technical debt... we've built a new component called a task list, which is designed to both sit on its own page in Collections but also in the sidebar of a wide variety of pages across GOV.UK. It was originally going to be a variation on the accordion component but due to this overall problem we decided to keep the accordion in Collections as is and copy and modify the code to a new task list component in static.

It would be good to know what specific issues the new task list component is running in to? I'm interested in understanding where the caching issues are coming from, as looking at [1], as far as I can see, the requested assets look to be content addressed, as the paths contain hashes.

1: https://www.gov.uk/learn-to-drive-a-car

andysellick:
> It would be good to know what specific issues the new task list component is running in to? I'm interested in understanding where the caching issues are coming from, as looking at [1], as far as I can see, the requested assets look to be content addressed, as the paths contain hashes.

Reuse of code is poor - the task list component has much code in common with the accordion component but they can't share it. In an ideal world we'd write a generic bit of accordion JS that could be used by both components. They also have similar GA tracking JS, which again could be reused rather than duplicated.

h-lame:
I'd like to unpick the:

> but they can't share it.

and 

> In an ideal world we'd write a generic bit of accordion JS that could be used by both components

statements, because nothing about static prevents us writing that JS and hosting it in static and then having Collections re-use that for it's accordion, and static re-use it for the task-list component.  Or, more correctly, nothing in _my understanding_ of how things hangs together prevents it.  Obviously though, something _did_ prevent it, so maybe a quick win is to write down the exact problem and we can fix that as a stepping stone towards the goal of this RFC.

andysellick:
It'll work in the browser but not in a test. A component JS test (in say, Collections) that relies upon JS in static can't currently access it. We can make up these gaps with smokey tests but they're not linked to PRs.

h-lame:
Thanks @andysellick - that's a really good point - I hadn't considered testing where the static JS isn't available.  I guess we could set up integration tests in the app to rely on running a local copy of static, but that definitely feels hard to orchestrate and I'm not sure any other parts of gov.uk do anything similar (other than the e2e tests or smokey which as you rightly say, aren't ideal because they're not attached to PRs).

cbaines:
> Thanks @andysellick - that's a really good point - I hadn't considered testing where the static JS isn't available. I guess we could set up integration tests in the app to rely on running a local copy of static, but that definitely feels hard to orchestrate and I'm not sure any other parts of gov.uk do anything similar (other than the e2e tests or smokey which as you rightly say, aren't ideal because they're not attached to PRs).

The tests for most applications already rely on services, e.g. the Publishing API relies on PostgreSQL.

You could deploy static to the ci-agents, like the other services used in tests?

tijmenb:
> You could deploy static to the ci-agents, like the other services used in tests?

Yes, but this doesn't solve the other problems described in the RFC, so it would be a partial solution at best.

nickcolley:
Sorry, I commented without reading, and closed this by mistake. Sorry all :)

tijmenb:
Thanks everyone for the comments! I'm going to pause this RFC.

I think there's broad agreement. The main blockers to getting this RFC accepted are answers to the following questions:

- How do we deal with duplication of asset downloads?
- Do we completely retire static? What do we do with the survey, emergency publishing, fonts and layouts?

Follow up actions before re-opening:

- [ ] Determine the user impact of duplicated assets in different scenarios, so that we can understand how much effort to put into a solution
- [ ] Following on from that, spike a system that [shares assets for users](https://github.com/alphagov/govuk-rfcs/pull/84#issuecomment-342460714)
- [ ] Come up with a vision on emergency publishing in this system  
- [ ] Come up with a vision on surveys in this system  
- [ ] Come up with a vision on layouts in this system

PR #85, opened 2017-10-18
RFC 85: Special Route Publisher app

brenetic:
[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/special-route-publisher/rfc-085-special-route-publisher.md)

### DIFF
--- a/rfc-085-special-route-publisher.md
+++ b/rfc-085-special-route-publisher.md
@@ -1,0 +1,32 @@
+# Special Route Publisher app
+
+## Summary
+
+This RFC proposes the creation of a Special Route Publisher app to cater for routes that require registration but have no content.
+
+## Problem
+
+There are examples of certain routes that have no direct content and therefore do not live in any particular publishing but still require a route to be registered via the the Publishing-Api -> Content Store. Examples of these can be seen in [Frontend](https://github.com/alphagov/frontend/blob/master/lib/special_route_publisher.rb) and [Rummager](https://github.com/alphagov/rummager/blob/master/lib/tasks/publishing_api.rake) where the former publishes many routes, such as site search, and the latter the site map. Each app then has a rake task which is triggered on deploy and updates the routes via the Publishing API.
+
+This creates some inconsistencies/issues:
+
+1. Frontend apps should ideally not talk directly to the Publishing API.
+   This has become more relevant recently whilst moving site search into Finder Frontend.
+
+2. No central location for the administration of special routes.
+
+3. No consensus as to which app should take responsibility for
+   publishing any future special routes.
+
+4. Each app that publishes these routes does so in slightly different ways.
+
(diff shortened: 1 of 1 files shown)


MatMoore:
I think this makes sense - writing code in every frontend app for building, validating, and publishing content items feels unnecessarily complex just to record some data about which routes go where.

What would be the MVP for this?

tijmenb:
Which apps would use this? I wonder if there are cases where we want to use content from the app in the payload to the publishing-api. 

In Whitehall we register some pages like a special route, but [use the HTML for the payload](https://github.com/alphagov/whitehall/blob/master/lib/publish_static_pages.rb#L20-L25). For calendars, we use the local data to [send to rummager & publishing-api](https://github.com/alphagov/calendars/blob/master/lib/registerable_calendar.rb). In those cases, would it still make sense to move the "publishing" to a separate app?

nickcolley:
Is there a way to name this so it cant be confused with 'specialist-publisher'?

MatMoore:
@tijmenb These are all content items that don't get rendered from the content store. So there is always a risk the publishing platform gets out of sync with the app, and we don't notice (for example, updating the html and not updating `indexable_content` in the publishing platform).

I think this change would make that scenario more likely, but the risk is still there at the moment. :man_shrugging:

tijmenb:
@MatMoore That won't happen if you use the actual HTML for `indexable_content`, like Whitehall does with the history pages.

I guess my bigger question is whether this can be used by all frontend apps. If it wouldn't, perhaps the it's easier to keep the (bit crappy) way we currently have of registering routes. If there were 2 ways of registering these routes we'd probably only add to the confusion. 

@brenetic can we make a list of frontend apps that would/wouldn't use this?

brenetic:
Going to close this until we feel we have a clearer reason for making it and what it would be responsible for, beyond the handful of registered routes in Frontend and Rummager (etc.)

rubenarakelyan:
Update: we have now revisited this RFC in light of the `/api/content` route that needs to be registered in order to remove `publicapi`, the nginx configuration that currently serves that route. The only app that it made sense to publish this route from was publishing-api, which then led to the confusing situation of publishing-api needing to talk to itself in order to publish a route.

In addition, the migration of our frontend apps to AWS has exposed the fact that having frontends talking to publishing-api means we need more infrastructure to handle the situation where they exist in different hosts, and this is non-optimal.

Due to these reasons, we have accepted the idea of a Special Route Publisher app (actually just a rake task with associated Jenkins job) that publishes all these routes from a single place. As work is done on apps that publish special routes, the idea is these routes will be moved to this rake task instead.

There will some special cases (like calendars) where this will be more difficult since more content than a special route is published by a frontend app, but for most cases, we should be able to remove special route publishing from most apps.

PR #86, opened 2017-11-03
RFC 86: Draft Stack Rummager

cbaines:
This RFC came out of a discussion about the needs of Content Tagger to
get per taxon metrics about tagged content, including draft taxons,
that led to a discussion about draft content in general, and how the
behaviour of rummager affects the draft stack.

### DIFF
--- a/rfc-086-draft-stack-rummager.md
+++ b/rfc-086-draft-stack-rummager.md
@@ -14,10 +14,14 @@ There is only one instance of this service, which is primarily used
 for the live stack, but is also currently used from some services on
 the draft stack (see [govuk-puppet][puppet-frontend-rummager-config]).
 
-Rummager may return some draft content, depending on the internal
-index which is used. However, the amount of draft content available
-through rummager is set to decrease, as more types of content are
-migrated to the `govuk` index (see [ADR 4][rummager-adr-4]).
+The current Rummager instance may expose some associations between
+live content and draft content (e.g. draft mainstream browse pages or
+taxons), depending on the internal index which is used for the live
+content. However, the current trend is that associations with draft
+content will be supported for less live content, as more types of
+content are migrated to the `govuk` index (see [ADR
+4][rummager-adr-4]), as the process of populating this index does not
+involve fetching draft content from the Publishing API.
 
 Missing draft content makes the draft stack less useful. Previewing
 pages that use rummager for some of the data won't give an accurate
--- a/rfc-086-draft-stack-rummager.md
+++ b/rfc-086-draft-stack-rummager.md
@@ -36,6 +36,25 @@ stack, is to setup a draft instance of rummager, populated with the
 appropriate content via the Publishing API, and then switch to using
 this from draft services.
 
+## Action Plan
+
+ 1. Deploy rummager to the draft stack
+    - Involves changing `govuk-puppet`
+    - Possibly requires some new machines
+ 2. Send draft content to a new `draft_documents` RabbitMQ Exchange
+    - From the Publishing API, for at least the PutContent, and
+      Unpublish commands
+    - Don't send any access limited content, to avoid Rummager
+      indexing it
+ 3. Subscribe the draft instance of Rummager to the `draft_documents`
+    exchange, as well as the existing `published_documents` exchange
+
+If the above work is done, at this point, there will be a draft
+Rummager service. To switch the draft frontend apps to use this
+instead, the next step is to make an assessment regarding whether it
+will work better than the live instance. This will probably depend on
+how much content has been migrated to the `govuk` index.
+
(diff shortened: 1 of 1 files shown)


rfc-086-draft-stack-rummager.md
```diff
@@ -0,0 +1,37 @@
...
+representation of what the page would look like if the relevant
+content is published.
+
+## Proposal
+
+The proposal made here to improve the user experience on the draft
+stack, is to setup a draft instance of rummager, populated with the
+appropriate content via the Publishing API, and then switch to using
+this from draft services.
+
```
tijmenb: Might be worth adding a section why this will not be implemented anytime soon:

- This depends on all content being indexed by rummager via the message queue
- Functionality needs to be added to the publishing-api to broadcast changes in drafts, including a hard think about access-limiting (we don't want sensitive documents sent to search)

cbaines: I was unsure whether it would be good to put some discussion in the RFC, or keep the discussion separate. I'm happy to put something in if you think it would help through?

tijmenb: I think it's good to give some context around it. Maybe rephrase it as "implementation notes" or something like that?

MatMoore: It's also worth noting that *publishing apps* have never sent to draft content to production rummager, and this has always been an unintentional thing. Rummager isn't able to distinguish between draft/live. The new indexing method is the same as the old in that respect.

Draft taxons (or links to them) are only returned by rummager now because of a stopgap measure we put in last year to keep links in sync with the publishing api, before we started working on indexing everything from the message queue. 

This is because it's using the HTTP API, which does return links to draft content, but  publishing api normally excludes these when pushing stuff to the content store/message queue.

rfc-086-draft-stack-rummager.md
```diff
@@ -0,0 +1,37 @@
...
+## Problem
+
+The [draft stack (content preview)][draft-stack-docs] part of GOV.UK
+relies in places on the [rummager service (Search API)][rummager].
+
+There is only one instance of this service, which is primarily used
+for the live stack, but is also currently used from some services on
+the draft stack (see [govuk-puppet][puppet-frontend-rummager-config]).
+
+Rummager may return some draft content, depending on the internal
```
MatMoore: This is not draft content itself, it's links to draft content (eg the mainstream_browse_pages, topics, organisations fields of other documents). This could still have unintended consequences for the live stack though.

cbaines: I've attempted to clarify this a bit in 6a9f6433e4f05e42a235695acb282f6ef03f7845

cbaines:
> I think it's good to give some context around it. Maybe rephrase it as "implementation notes" or something like that?

I've now added an "Action Plan" (naming from the README), which includes some thoughts @tomsabin and I had on the feasibility of doing this.

Going back to a point made about the dependency on the migration to the `govuk` index:

> This depends on all content being indexed by rummager via the message queue

In the plan, I've set this out as a dependency for switching the draft frontends to use the draft rummager. While this is written about the draft stack, I'm also thinking about other uses like tagging metrics, and that use case might not have a hard dependency on completing the migration to the `govuk` index.

Another idea that came up when discussing this with Tom, if you wanted to use the draft rummager before the `govuk` index migration is finished, you could copy over the other indexes (that wouldn't be updated in the draft stack) from the live stack. This isn't something to consider yet, but I thought I'd put the idea out there.

rfc-086-draft-stack-rummager.md
```diff
@@ -0,0 +1,61 @@
...
+## Action Plan
+
+ 1. Deploy rummager to the draft stack
+    - Involves changing `govuk-puppet`
+    - Possibly requires some new machines
+ 2. Send draft content to a new `draft_documents` RabbitMQ Exchange
+    - From the Publishing API, for at least the PutContent, and
+      Unpublish commands
+    - Don't send any access limited content, to avoid Rummager
+      indexing it
```
kevindew: Are you concerned about discarded drafts too?

cbaines: Hmm, that's a good point. I guess that there is no prior art for conveying a discarded draft down the message queue, although the result should be the same as a unpublish with type vanish.

cbaines:
I'm planning to close this in just over 2 weeks on the 5th of January.

This is mostly a reminder to me, but if anyone has any comments that they haven't given yet, it would be useful to comment over the next couple of weeks. Thanks :)

cbaines:
Awesome, thanks for reviewing Mat.

I'll merge this now, as I'm not sure there is anything to be gained by leaving it open for longer, but obviously the actual execution of this remains uncertain.

PR #87, opened 2017-11-06
RFC 87: Dealing with errors

tijmenb:
Attempting to write down how we should treat different types of errors on GOV.UK.

👉 [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/errors/rfc-087-dealing-with-errors.md)

https://trello.com/c/wX66HhtR

### DIFF
--- a/rfc-087-dealing-with-errors.md
+++ b/rfc-087-dealing-with-errors.md
@@ -30,7 +30,9 @@ Desired behaviour: error is sent to Sentry, developers are notified and fix the 
 
 ### Intermittent errors without user impact
 
-Frontend applications often see timeouts when talking to the content-store or rummager.
+Frontend applications often see timeouts when talking to the content-store.
+
+There's no or little user impact because the request will be answered by the caching layer.
 
 Example: <https://sentry.io/govuk/app-finder-frontend/issues/352985400>
 
--- a/rfc-087-dealing-with-errors.md
+++ b/rfc-087-dealing-with-errors.md
@@ -77,3 +77,9 @@ Rails reports `ActionDispatch::RemoteIp::IpSpoofAttackError`.
 [Example](https://sentry.io/govuk/app-service-manual-frontend/issues/365951370)
 
 Desired behaviour: HTTP 400 is returned, error is not reported to Sentry.
+
+### Database entry not found
+
+Often a controller will do something like `Thing.find(params[:id])` and rely on Rails to show a 404 page for the `ActiveRecord::RecordNotFound` it raises ([context](https://stackoverflow.com/questions/27925282/activerecordrecordnotfound-raises-404-instead-of-500)).
+
+Desired behaviour: errors are not reported to Sentry
--- a/rfc-087-dealing-with-errors.md
+++ b/rfc-087-dealing-with-errors.md
@@ -28,13 +28,21 @@ A code change makes the application crash.
 
 Desired behaviour: error is sent to Sentry, developers are notified and fix the error.
 
-### Expected intermittent errors
+### Intermittent errors without user impact
 
 Frontend applications often see timeouts when talking to the content-store or rummager.
 
 Example: <https://sentry.io/govuk/app-finder-frontend/issues/352985400>
 
 Desired behaviour: error is not sent to Sentry. Instead, we rely on Smokey and Icinga checks to make sure we the site functions.
+
+### Intermittent errors with user impact
+
+Publishing applications sometimes see timeouts when talking to publishing-api. This results in the publisher seeing an error page and possibly losing data.
+
+Example: <https://sentry.io/govuk/app-content-tagger/issues/367277928>
+
+Desired behaviour: TBD
 
 ### Expected environment-based errors
 
--- a/rfc-087-dealing-with-errors.md
+++ b/rfc-087-dealing-with-errors.md
@@ -76,7 +76,9 @@ Desired behaviour: backing app returns a 4XX status code, response is fed back t
 
 Something goes wrong and we need to let developers know.
 
-Desired behaviour: developers do not use Sentry for logging.
+Example: [Slimmer's old behaviour](https://github.com/alphagov/slimmer/pull/203/files#diff-e5615a250f587cf4e2147f6163616a1a)
+
+Desired behaviour: developers do not use Sentry for logging. The app either raises the actual error (which causes the user to see the error) or logs the error to Kibana.
 
 ### Intermittent retryable errors
 
--- a/rfc-087-dealing-with-errors.md
+++ b/rfc-087-dealing-with-errors.md
@@ -16,7 +16,11 @@ There are 2 principles:
 
 Applications should report exceptions to Sentry. Applications must not swallow errors.
 
-### 2. Applications should not error
+### 2. Notifications should be actionable
+
+Sentry notifications should be something that requires a developer of the app to do something about it. Not just a piece of information.
+
+### 3. Applications should not error
 
 The goal of GOV.UK is that applications should not error. When something goes wrong it should be fixed.
 
--- a/rfc-087-dealing-with-errors.md
+++ b/rfc-087-dealing-with-errors.md
@@ -1,0 +1,53 @@
+# Dealing with errors
+
+## Summary
+
+This describes how we treat errors on GOV.UK
+
+## Problem
+
+We've recently migrated to a new error tracking service, Sentry. This provides us an opportunity to rethink how we treat errors.
+
+## Proposal
+
+There are 2 principles:
+
+### 1. When something goes wrong, we should be notified
+
+Applications should report exceptions to Sentry. Applications must not swallow errors.
+
+### 2. Applications should not error
+
+The goal of GOV.UK is that applications should not error. When something goes wrong it should be fixed.
+
+## Classifying errors
+
+### Bug
+
+A code change makes the application crash.
+
+Desired behaviour: error is sent to Sentry, developers are notified and fix the error.
+
+### Expected intermittent
+
+Frontend applications often see timeouts when talking to the content-store or rummager.
+
+Desired behaviour: error is not sent to Sentry. Instead, we rely on Smokey and Icinga checks to make sure we the site functions.
+
+### Expected environment-based
+
+MySQL errors on integration while data sync happens.
+
+Desired behaviour: our environment is set up such that these errors do not occur.
+
+### Incorrect treatment of input
+
+Rummager crashing on incorrect dates, passed on directly from finder-frontend.
+
+Desired behaviour: backing app returns a 4XX status code, response is fed back to user. Nothing is ever logged or sent to Sentry.
+
+### Manually logged errors
+
+Something goes wrong and we need to let developers know.
+
+Desired behaviour: developers do not use Sentry for logging. 
--- a/rfc-087-dealing-with-errors.md
+++ b/rfc-087-dealing-with-errors.md
@@ -50,4 +50,10 @@ Desired behaviour: backing app returns a 4XX status code, response is fed back t
 
 Something goes wrong and we need to let developers know.
 
-Desired behaviour: developers do not use Sentry for logging. 
+Desired behaviour: developers do not use Sentry for logging.
+
+### Intermittent retryable errors
+
+Sidekiq worker sends something to the publishing-api, which times out. Sidekiq retries, the next time it works.
+
+Desired behaviour: errors are not reported to Sentry until retries are exhausted. See [this PR for an example](https://github.com/alphagov/content-performance-manager/pull/353).  
--- a/rfc-087-dealing-with-errors.md
+++ b/rfc-087-dealing-with-errors.md
@@ -28,21 +28,29 @@ A code change makes the application crash.
 
 Desired behaviour: error is sent to Sentry, developers are notified and fix the error.
 
-### Expected intermittent
+### Expected intermittent errors
 
 Frontend applications often see timeouts when talking to the content-store or rummager.
 
 Desired behaviour: error is not sent to Sentry. Instead, we rely on Smokey and Icinga checks to make sure we the site functions.
 
-### Expected environment-based
+### Expected environment-based errors
 
-MySQL errors on integration while data sync happens.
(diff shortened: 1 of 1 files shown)


MatMoore:
I broadly agree with this but am cautious of trying to classify everything at once, when we've just changed our tooling.

What if we made the classification part a living google doc, and then this RFC is just "we will decide what does and doesn't belong in sentry and get the error count to 0"?

tijmenb:
@MatMoore good point, we could move the classification list to the docs after this gets agreed, so we can make amendments more easily. Or amend the RFC when appropriate. An RFC is the correct place to discuss this though - any particular thing you have comments about?

MatMoore:
I have opinions on everything, but I don't want to derail too much 😀 

In particular I think https://github.com/alphagov/govuk-rfcs/blob/errors/rfc-087-dealing-with-errors.md#manually-logged-errors is really a subset of a broader problem with application logging on GOV.UK.

When the application has got into a state where it has to give up on something, there are things the developer should know. But as a developer debugging a problem, I want to know what the application did *before* it reached that critical point, and even if the application was able to respond I still want to see what it did. Only showing diagnostic information when something has definitely gone wrong makes it harder to debug, because I can't tell if I'm not seeing logs because they didn't get logged, or because I'm using Kibana wrong.

We [deliberately built diagnostic logging into the work we're doing for search indexing](https://github.com/alphagov/rummager/blob/master/lib/govuk_index/publishing_event_worker.rb#L53), because there is a lot of complexity in the system at the moment, and there are a lot of reasons why content can not be updated. Debugging it is hard, but reading these logs is often enough to quickly isolate where the problem is, and verify that yes, this message was received by Rummager and the system isn't magic.

So I agree this is not an appropriate thing to put in sentry and raise alerts for, but I think we could do a lot more to improve application logging to Kibana and would like to discuss it more in general.

Availability errors like timeouts can cause problems for users, but I don't think they are very actionable in sentry. In particular, for Rummager, it's often hard to distinguish between timeouts in our API (which can interfere with rendering) and timeouts that take place within different contexts like rake tasks or sidekiq workers.

We should absolutely be tracking when this happens (for example by storing metrics in graphite) but raising sentry errors doesn't seem helpful unless there is an obvious way to intervene. 

I think that responding to failure of dependent services is a concern that should be handled in shared libraries, rather than applications themselves, because it's tricky to get right and we should have common approaches to handling it. Have a shared solution for in band HTTP requests, a shared solution for background workers etc.

Regarding https://github.com/alphagov/govuk-rfcs/blob/errors/rfc-087-dealing-with-errors.md#incorrect-bubbling-up-of-errors, Rummager is absolutely doing the wrong thing in this case. The error should be returned as 4xx, and the upstream app should either deal with it or not generate invalid requests. But I think there is another issue here...
 
If an upstream app can't do something because of a failure in a dependent service, that is a critical piece of information to have in the application logs. By not reporting this as an error with the upstream app, I am more likely to see that the actual broken app is broken. But I still might want to investigate what the impact was on upstream applications, in which case I want to see something in kibana. On the other hand if I start investigating a problem by looking at the upstream app, seeing that log message signposts the root cause and might prompt me to go look at sentry for that app in the first place. So I think this is another case where we should not take the information out of sentry without first making sure we have good metrics or logging in other places.

MatMoore:
Another thing that possibly ties in with this RFC, is our workflow for sentry. It's possible to [resolve and silence](https://blog.sentry.io/2017/06/30/workflow-improvements) errors in sentry and it's also possible to customise the conditions when an event triggers a notification. By default, we're only notifying on the first occurrence of an error, and after that the only way it gets fixed is if people look at sentry and see it.

This is something we're experimenting with a bit on the search team for rummager, but non-actionable errors are the big problem at the moment, and some of the changes in this RFC will definitely help with this.

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Incorrect bubbling up of errors
+
+Rummager crashes on date parsing, returns 500, which is passed on directly in finder-frontend.
+
+Example: <https://sentry.io/govuk/app-finder-frontend/issues/400074507>
+
+Desired behaviour: backing app returns a 4XX status code, response is fed back to user. Nothing is ever logged or sent to Sentry.
+
+### Manually logged errors
```
MatMoore: I think this is really a subset of a broader problem with application logging on GOV.UK.

When the application has got into a state where it has to give up on something, there are things the developer should know. But as a developer debugging a problem, I want to know what the application did *before* it reached that critical point, and even if the application was able to respond I still want to see what it did. Only showing diagnostic information when something has definitely gone wrong makes it harder to debug, because I can't tell if I'm not seeing logs because they didn't get logged, or because I'm using Kibana wrong.

We [deliberately built diagnostic logging into the work we're doing for search indexing](https://github.com/alphagov/rummager/blob/master/lib/govuk_index/publishing_event_worker.rb#L53), because there is a lot of complexity in the system at the moment, and there are a lot of reasons why content can not be updated. Debugging it is hard, but reading these logs is often enough to quickly isolate where the problem is, and verify that yes, this message was received by Rummager and the system isn't magic.

So I agree this is not an appropriate thing to put in sentry and raise alerts for, but I think we could do a lot more to improve application logging to Kibana. I'd like to discuss it more in general.

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+
+Frontend applications often see timeouts when talking to the content-store.
+
+There's no or little user impact because the request will be answered by the caching layer.
+
+Example: <https://sentry.io/govuk/app-finder-frontend/issues/352985400>
+
+Desired behaviour: error is not sent to Sentry. Instead, we rely on Smokey and Icinga checks to make sure we the site functions.
+
+### Intermittent errors with user impact
```
MatMoore: Availability errors like timeouts can cause problems for users, but I don't think they are very actionable in sentry. In particular, for Rummager, it's often hard to distinguish between timeouts in our API (which can interfere with rendering) and timeouts that take place within different contexts like rake tasks or sidekiq workers.

We should absolutely be tracking when this happens (for example by storing metrics in graphite) but raising sentry errors doesn't seem helpful unless there is an obvious way to intervene.

MatMoore: I think that responding to failure of dependent services is a concern that should be handled in shared libraries, rather than applications themselves, because it's tricky to get right and we should have common approaches to handling it. Have a shared solution for in band HTTP requests, a shared solution for background workers etc.

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+
+Often happens in [security checks](https://sentry.io/govuk/app-frontend/issues/400074979).
+
+Example: <https://sentry.io/govuk/app-frontend/issues/400074979>
+
+Desired behaviour: user gets feedback, error is not reported to Sentry
+
+[bad-request]: https://sentry.io/govuk/app-service-manual-frontend/issues/400074003
+
+### Incorrect bubbling up of errors
```
MatMoore: Rummager is absolutely doing the wrong thing in this case. The error should be returned as 4xx, and the upstream app should either deal with it or not generate invalid requests. But I think there is another issue here...
 
If an upstream app can't do something because of a failure in a dependent service, that is a critical piece of information to have in the application logs. By not reporting this as an error with the upstream app, I am more likely to see that the actual broken app is broken. But I still might want to investigate what the impact was on upstream applications, in which case I want to see something in kibana. On the other hand if I start investigating a problem by looking at the upstream app, seeing that log message signposts the root cause and might prompt me to go look at sentry for that app in the first place.

So I think this is another case where we should not take the information out of sentry without first making sure we have good metrics or logging in other places.

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+There are 2 principles:
+
+### 1. When something goes wrong, we should be notified
+
+Applications should report exceptions to Sentry. Applications must not swallow errors.
+
+### 2. Applications should not error
+
+The goal of GOV.UK is that applications should not error. When something goes wrong it should be fixed.
+
```
MatMoore: Perhaps another principle:

Sentry notifications should be something that requires a developer of the app to do something about it. Not just a piece of information.

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+Example: <https://sentry.io/govuk/app-frontend/issues/400074979>
+
+Desired behaviour: user gets feedback, error is not reported to Sentry
+
+[bad-request]: https://sentry.io/govuk/app-service-manual-frontend/issues/400074003
+
+### Incorrect bubbling up of errors
+
+Rummager crashes on date parsing, returns 500, which is passed on directly in finder-frontend.
```
MatMoore: Rummager is absolutely doing the wrong thing in this case. Bad rummager.

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+
+Example: <https://sentry.io/govuk/app-content-tagger/issues/367277928>
+
+Desired behaviour: TBD
+
+### Intermittent retryable errors
+
+Sidekiq worker sends something to the publishing-api, which times out. Sidekiq retries, the next time it works.
+
+Desired behaviour: errors are not reported to Sentry until retries are exhausted. See [this PR for an example](https://github.com/alphagov/content-performance-manager/pull/353).
```
Davidslv: This is not desirable, unless we change the default Sidekiq behaviour, otherwise we will only be notified of any bug 21 days later https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry
We should probably define a new default, 21 days is too long

rfc-087-dealing-with-errors.md
kevindew: Agree with these 👍

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Intermittent errors without user impact
+
+Frontend applications often see timeouts when talking to the content-store.
+
+There's no or little user impact because the request will be answered by the caching layer.
+
+Example: <https://sentry.io/govuk/app-finder-frontend/issues/352985400>
+
+Desired behaviour: error is not sent to Sentry. Instead, we rely on Smokey and Icinga checks to make sure we the site functions.
```
kevindew: I wonder if some of this behaviour should be handled at GDS API Adapters or another appropriate layer? It seems like a bit of a problem if each of the Frontend apps all have the same timeout handling logic?

rfc-087-dealing-with-errors.md
kevindew: @Davidslv this should only be the case with errors that are already marked as retryable, so they shouldn't be ones that are surprise - but it does raise the question of how we find out if every job is failing with a retryable errors and nothing succeeding.

I do wonder if we want to introduce some sidekiq middleware that makes our retrying logic consistent.

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Bad request errors
+
+User makes a request the application can't handle ([example][bad-request]).
+
+Often happens in [security checks](https://sentry.io/govuk/app-frontend/issues/400074979).
+
+Example: <https://sentry.io/govuk/app-frontend/issues/400074979>
+
+Desired behaviour: user gets feedback, error is not reported to Sentry
```
kevindew: Seems reasonable though I'd not that your security checks example is something that should be fixed as it looks like a params issue.

rfc-087-dealing-with-errors.md
kevindew: Is this the sort of thing where we can tag the error with something in sentry so it can be isolated?

rfc-087-dealing-with-errors.md
```diff
@@ -0,0 +1,105 @@
...
+Desired behaviour: backing app returns a 4XX status code, response is fed back to user. Nothing is ever logged or sent to Sentry.
+
+### Manually logged errors
+
+Something goes wrong and we need to let developers know.
+
+Example: [Slimmer's old behaviour](https://github.com/alphagov/slimmer/pull/203/files#diff-e5615a250f587cf4e2147f6163616a1a)
+
+Desired behaviour: developers do not use Sentry for logging. The app either raises the actual error (which causes the user to see the error) or logs the error to Kibana.
+
```
kevindew: Are we going to try avoid warnings in Sentry via this? I feel like this still offers us some useful advantages than Kibana since the logs are retained longer.

rfc-087-dealing-with-errors.md
tijmenb: Added

rfc-087-dealing-with-errors.md
tijmenb: Updated in https://github.com/alphagov/govuk-rfcs/pull/87/commits/494975f9ce8e43e89b028413cd35c267f3e9d268. Do you agree?

rfc-087-dealing-with-errors.md
tijmenb: I think I had this one wrong. Rummager does the right thing by returning a `422`. Fixed in https://github.com/alphagov/govuk-rfcs/pull/87/commits/f78767c0d339fdcdde0083b5537784a6ad81d566.

rfc-087-dealing-with-errors.md
tijmenb: Good point - added in https://github.com/alphagov/gds-api-adapters/issues/776.

rfc-087-dealing-with-errors.md
tijmenb: I suppose an error like https://sentry.io/govuk/app-frontend/issues/400074979 can be prevented by checking that the params are in the correct format (an integer in this case). Is that what you mean?

rfc-087-dealing-with-errors.md
tijmenb: Do you have an example of a thing you'd use this for?

tijmenb:
@MatMoore @kevindew can you check if there are open questions. Otherwise I'll merge this on Friday.

PR #88, opened 2017-12-12
RFC 88: External content

MatMoore:
[Rendered markdown](https://github.com/alphagov/govuk-rfcs/blob/459208e61c7661ecc0a569c4bc31e7bcdeaf09fd/rfc-000-external-route.md)

I'm aiming to make a decision on this RFC by the end of the week, unless there are strong objections.

## Changes
- Dec 12 11:33 Initial draft
- Dec 12 14:42 Renamed `external_route` -> `external_content` and added reference to RFC 43
- Dec 14 15:24 Removed `comment` field from details hash.


boffbowsh:
I agree in principal, but I think maybe `external_link` or `external_content`? It doesn't feel like `_route` makes sense in this context, as we'd never route to it in the traditional sense. 

Sorry for bikeshedding.

rfc-000-external-route.md
```diff
@@ -0,0 +1,45 @@
...
+## Proposal
+
+Create an `external_route` document type so we can store these links in the publishing API.
+
+Create an `external_route` schema.
+
+The schema MUST store in the details hash:
+- `hidden_search_terms` - a set of search keywords/phrases that should route users to the page. [This field has already been added for smart answers](https://github.com/alphagov/govuk-content-schemas/pull/685/files).
+-  `url` - a URL for the external resource
+- `comment` - an internal description of why the link is relevant to GOV.UK users
```
tijmenb: Is this the same as the field in search-admin? Maybe we don't want to display this publicly.

rfc-000-external-route.md
```diff
@@ -0,0 +1,45 @@
...
+
+The schema MAY store information about change history.
+
+The standard fields `title` and `description` MUST be set.
+The `base_path` MUST NOT be set.
```
tijmenb: And probably the other fields that aren't relevant to items w/o base_path: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-043-content-items-without-a-base-path.md (note that that RFC hasn't been fully implemented I think).

tijmenb:
👍 Looks sensible, some minor comments.

rfc-000-external-route.md
MatMoore: Yes it is. It's fine to take this out, but I don't think there's a strong reason it can't be public. We would have to clean up the values a bit first though.

rfc-000-external-route.md
MatMoore: Ah thanks, I'll update the RFC to be consistent with this.

MatMoore:
@boffbowsh no problem, I'm fine with changing the name. I avoided `link` deliberately to avoid confusion with the publishing platform concept of links between two content items. I like `external_content` though.

tijmenb:
+1 for `external_content`

MatMoore:
external_content it is 👍

rfc-000-external-route.md
MatMoore: I've updated the RFC to take this out as we don't need it.

PR #90, opened 2018-01-18
Removing the /government prefix route

cbaines:
The /government prefix route is a large source of uncertainty, and
because of this uncertainty it is quite difficult to work out what
specific content still uses it. Changing the way the traffic is
handled could offer a methodical way of tracking what the remaining
work is to make the /government prefix route redundant.

https://github.com/alphagov/govuk-rfcs/blob/rfc-89/rfc-089-removing-the-government-prefix-route.md

### DIFF
--- a/rfc-089-removing-the-government-prefix-route.md
+++ b/rfc-089-removing-the-government-prefix-route.md
@@ -1,0 +1,54 @@
+# Removing the /government prefix route
+
+## Summary
+
+The /government prefix route is a large source of uncertainty, and
+because of this uncertainty it is quite difficult to work out what
+specific content still uses it. Changing the way the traffic is
+handled could offer a methodical way of tracking what the remaining
+work is to make the /government prefix route redundant.
+
+## Problem
+
+The /government prefix route is a large source of uncertainty in the
+context of the rest of the routing on GOV.UK. It's hard to tell what
+it's responsible for, and thus what work needs doing to get Whitehall
+content to manage it's own routes.
+
+Also, while I think it would be possible in the router to have
+separate handlers for /governement as an exact route, and /government
+as a prefix route, this isn't supported in either the Publishing API
+or Router API. This means that /government (the exact path) is
+currently stuck being rendered by whitehall-frontend.
+
+## Proposal
+
+There are some known groups of content rendered by whitehall-frontend
+for which the routing information isn't explicitly set in the
+Publishing API.
+
+However, even if that were fixed, due to the nature of the problem,
+the uncertainty of what this prefix route is responsible for, it would
+be hard to tell what routing information is missing from what content.
+
+### Action Plan
+
+To make it possible to measure and track what requests are still being
+managed by the /government prefix route, the traffic to
+whitehall-frontend could be split at the router.
+
+Instead of all traffic going directly to whitehall-frontend, a new
+virtual host, say whitehall-frontend-prefix-route could be created,
+which like the whitehall-frontend virtual host would reverse proxy to
+the whitehall-frontend. The rendering app for the special-route at
+/government would then be changed to whitehall-frontend-special-route,
+to direct traffic handled by the prefix route to this new virtual
+host.
+
+This wouldn't change the experience for users, but would allow
+tracking what successful requests are being made through the
+whitehall-frontend prefix route. If there comes a time when all the
+recent requests haven't been successfull (e.g. all responded to with a
+404 status), then this would suggest that the routing information for
+content rendered by whitehall-frontend is complete enough to do
+without the /government prefix route.


rfc-089-removing-the-government-prefix-route.md
```diff
@@ -0,0 +1,54 @@
...
+
+There are some known groups of content rendered by whitehall-frontend
+for which the routing information isn't explicitly set in the
+Publishing API.
+
+However, even if that were fixed, due to the nature of the problem,
+the uncertainty of what this prefix route is responsible for, it would
+be hard to tell what routing information is missing from what content.
+
+### Action Plan
```
tijmenb: I wonder if we can do this too by getting a months worth of requests from the CDN logs, filtering by `/government` prefix and checking if they have exact matches in the router. That would save us from having to alter the infrastructure. What do you think?

rubenarakelyan:
Couple of things:

1. We should also do this with the `/world` prefix route since that is also rendered mostly by whitehall with some exact routes under it being rendered by collections.

2. Part of this will involve changing placeholders to register routes, or changing formats that use placeholders to use fully-featured content items with associated routes.

rfc-089-removing-the-government-prefix-route.md
cbaines: > I wonder if we can do this too by getting a months worth of requests from the CDN logs, filtering by /government prefix and checking if they have exact matches in the router. That would save us from having to alter the infrastructure. What do you think?

I think that approach would work, however if putting in the effort of writing something to check for the existence of routes is worth it, it would be more useful to check against the routes in the Publishing API, as that would give better data (as it would avoid routes that have crept in to the router outside of the Publishing API).

Overall, while doing the route checking in code rather than using the running router service would avoid changing the infrastructure, my feeling is that it would take more effort to setup (although using data from the Publishing API could have an added benefit).

cbaines:
> We should also do this with the /world prefix route since that is also rendered mostly by whitehall with some exact routes under it being rendered by collections.

Sure, yeah, if something is put in place for /government, hopefully it would be easy enough to get some data on /world as well.

> Part of this will involve changing placeholders to register routes, or changing formats that use placeholders to use fully-featured content items with associated routes.

Yes, while the title of this RFC is about removing the route, I didn't really go in to detail about what changes would need to be made to the content and the respective routes, but the placeholders issue might be a big one. I've also got one Pull Request on this subject open if you're interested in reviewing [1]?

1: https://github.com/alphagov/whitehall/pull/3696

rfc-089-removing-the-government-prefix-route.md
tijmenb: > my feeling is that it would take more effort to setup 

I think we could loop through the URLs and query the router directly? There would be no new code to write.

https://github.com/alphagov/router-api#getting-details-of-a-route

cbaines: > I think we could loop through the URLs and query the router directly? There would be no new code to write.

I was including that loop in the code that would need to be written and that considering prefix routes under /government as a "must have". Adding more specific prefix routes might be a practical way of getting rid of the /government prefix route and by my count there are quite a few prefix routes within /government already [1] that may be handling some requests.

1: 
```
irb(main):014:0> Route.where(route_type: "prefix", incoming_path: { '$regex': /^\/government\/.*/ }).count
=> 733
```

boffbowsh:
I'll have a proper look at this soon, but FYI this RFC would be RFC 90, not 89 as in the filename.

cbaines:
> I'll have a proper look at this soon, but FYI this RFC would be RFC 90, not 89 as in the filename.

I'm now a little confused, as from reading the README again, it should actually be 000, although it doesn't say when to change the name... I went for 89, as the last RFC I could see was number 88, but I could have missed one?

boffbowsh:
This is Pull Request 90, so it's RFC 90. We number them after the PR to avoid clashes when people are working on two RFCs at the same time.

boffbowsh:
I feel like this is adding complexity to an already-complicated part of the stack in order to provide measurement of something we can observe already from logs, as Tijmen suggests. 

I agree that removing this prefix route is something we should look to do (it is [tracked on the Tech Debt board](https://trello.com/c/74KTVXBs/52-whitehall-relies-on-prefix-government-and-world-routes)), but in reality ensuring that every URL that is served by Whitehall is in the Publishing API is not a small amount of work. There is currently a Q4 mission for a new Publishing Workflow alpha, one of the outcomes of which is expected to be a recommendation of what to do with Whitehall. Spending significant migration effort before the outcome of that work is known might be a waste.

tl;dr yes we should measure this but in a less obtrusive way, but we probably won't be able to do anything about it for a little while.

cbaines:
> I feel like this is adding complexity to an already-complicated part of the stack in order to provide measurement of something we can observe already from logs, as Tijmen suggests.

> tl;dr yes we should measure this but in a less obtrusive way, but we probably won't be able to do anything about it for a little while.

My thinking on this is also that logs are the way to go, just that it's going to be easier and less costly both in the short and long term to put the data in the logs by using a different application name, rather than trying to annotate the logs after the fact, I guess by performing the same process as the router (either reimplementing the same logic, or using the same code).

While the approach suggested here is around making changes to some configuration, I don't see it as obtrusive especially in a negative way, but instead as taking advantage of the existing architecture (logging and otherwise).

boffbowsh:
We already have all the data needed though, we don't need to annotate things at source. This does add a lot of complexity and confusion around an area that is a) not often worked on directly and b) frequently debugged. That means that various people rolling through Platform Support would need to be aware of the differences, which would hopefully only be short-lived but with no guarantee. 

Another reason is that as things stand we can look at logs now for up to 30 days and observe what we need, but with this method we'd need to wait 30 days for the same amount of data.

tijmenb:
@cbaines I'm going to close this in the interest of tidiness and the lack of activity. Feel free to reopen if there's more to discuss!

PR #91, opened 2018-02-08
RFC91: Sharing assets

tijmenb:
👉 [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/sharing-assets/rfc-091-sharing-assets.md)

Closing Friday 16 February.

https://trello.com/c/dCXLbdBH

### DIFF
--- a/rfc-091-sharing-assets.md
+++ b/rfc-091-sharing-assets.md
@@ -1,0 +1,43 @@
+# Sharing assets
+
+## Summary
+
+We need to share assets to make sure users only have to download a thing once. To do this, we'll upload each application's assets to the same location.
+
+## Problem
+
+In [RFC84 we proposed replacing Static with a gem](https://github.com/alphagov/govuk-rfcs/pull/84). One of the architectural challenges with this approach is that by moving assets (like CSS, JS and images) into the gem, the assets won't be shared for users. This means that users visiting GOV.UK would have to download assets multiple times, depending on what app renders the page.
+
+## Proposal
+
+We upload the assets to a central location. For example, if both `collections` and `government-frontend` have a file named `foo.css`, they will upload the file to `assets.publishing.service.gov.uk/shared/foo.css`. Because [Rails fingerprints assets](http://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark), the asset actually uploaded will be `assets.publishing.service.gov.uk/shared/foo-8d811b8c3badbc0b0e2f6e25d3660a96cc0cca7993e6f32e98785f205fc40907.css`, unless the files are different, in which case they'll have different fingerprints and won't overwrite.
+
+How to achieve this:
+
+* In govuk-app-deployment, we configure capistrano to precompile the assets and then upload the resulting files to a central location.
+* Proxy assets.publishing.service.gov.uk/shared to this location
+* Point applications to use this URL as asset host so that it's used in production
+
+Upsides
+
+* We remove Nginx config to proxy to individual applications, perhaps removing some hops
+* Whenever there's an asset to share, it will be shared automatically. If an application copies an image from another app, on production they'll have the same URL.
+* Each application ensures that its own assets are present on the shared bucket.
+* Applications don't serve files anymore.
+
+Downsides
+
+* We'll end up with a directory full of assets which are of unknown origin.
+
+## Alternatives
+
+### 1. Do nothing
+
+The simplest solution is to not share assets between applications. This would have significant impact on the user, especially if the font will be distributed in the gem.
+
+### 2. Deploy the gem
+
+This would work roughly that every time we publish a new gem version, we'll deploy the resulting assets to a central place. In production apps point to that shared asset URL instead of
+
+* Disparity between development and production
+* Inevitable deploy dependency - the gem deploy needs to happen before the application is deployed
--- a/rfc-091-sharing-assets.md
+++ b/rfc-091-sharing-assets.md
@@ -10,7 +10,11 @@ In [RFC84 we proposed replacing Static with a gem](https://github.com/alphagov/g
 
 ## Proposal
 
-We upload the assets to a central location. For example, if both `collections` and `government-frontend` have a file named `foo.css`, they will upload the file to `assets.publishing.service.gov.uk/shared/foo.css`. Because [Rails fingerprints assets](http://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark), the asset actually uploaded will be `assets.publishing.service.gov.uk/shared/foo-8d811b8c3badbc0b0e2f6e25d3660a96cc0cca7993e6f32e98785f205fc40907.css`, unless the files are different, in which case they'll have different fingerprints and won't overwrite.
+We upload the assets to a central location. For example, if both `collections` and `government-frontend` have a file named `foo.css`, they will upload the file to `assets.publishing.service.gov.uk/shared/foo.css`.
+
+Because [Rails fingerprints assets](http://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark), the asset actually uploaded will be `assets.publishing.service.gov.uk/shared/foo-8d811b8c3badbc0b0e2f6e25d3660a96cc0cca7993e6f32e98785f205fc40907.css`, unless the files are different, in which case they'll have different fingerprints and won't overwrite.
+
+As part of this solution we'd need a way to ensure or encourage different frontend apps to use assets in a way that is shareable. For example, applications could include a separate stylesheet with all the component styles (`components.css`), which would be shared across apps.
 
 How to achieve this:
 


sihugh:
A problem that have in terms of measuring the benefit of any of this is that we don't know the impact to users of any of these sorts of decisions.  I wrote [a card for our backlog](https://trello.com/c/wPcGWJbi/386-set-up-performance-testing-metrics-for-govuk-pages) that suggests running [some web page tests](https://www.webpagetest.org/) across various sorts of pages and uploading metrics to graphite, but I don't know when we're going to get to it.

rfc-091-sharing-assets.md
```diff
@@ -0,0 +1,43 @@
...
+Upsides
+
+* We remove Nginx config to proxy to individual applications, perhaps removing some hops
+* Whenever there's an asset to share, it will be shared automatically. If an application copies an image from another app, on production they'll have the same URL.
+* Each application ensures that its own assets are present on the shared bucket.
+* Applications don't serve files anymore.
+
+Downsides
+
+* We'll end up with a directory full of assets which are of unknown origin.
```
sihugh: We could (for example) get the script to write a log of assets to a Github repo.  We could then have an icinga check to alert us if anything else was in the bucket unexpectedly (if we thought we needed it)

rfc-091-sharing-assets.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Alternatives
+
+### 1. Do nothing
+
+The simplest solution is to not share assets between applications. This would have significant impact on the user, especially if the font will be distributed in the gem.
```
sihugh: A problem that have in terms of measuring the benefit of any of this is that we don't know the impact to users of any of these sorts of decisions.  I wrote [a card for our backlog](https://trello.com/c/wPcGWJbi/386-set-up-performance-testing-metrics-for-govuk-pages) that suggests running [some web page tests](https://www.webpagetest.org/) across various sorts of pages and uploading metrics to graphite, but I don't know when we're going to get to it.

rfc-091-sharing-assets.md
```diff
@@ -0,0 +1,43 @@
...
+
+We need to share assets to make sure users only have to download a thing once. To do this, we'll upload each application's assets to the same location.
+
+## Problem
+
+In [RFC84 we proposed replacing Static with a gem](https://github.com/alphagov/govuk-rfcs/pull/84). One of the architectural challenges with this approach is that by moving assets (like CSS, JS and images) into the gem, the assets won't be shared for users. This means that users visiting GOV.UK would have to download assets multiple times, depending on what app renders the page.
+
+## Proposal
+
+We upload the assets to a central location. For example, if both `collections` and `government-frontend` have a file named `foo.css`, they will upload the file to `assets.publishing.service.gov.uk/shared/foo.css`. Because [Rails fingerprints assets](http://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark), the asset actually uploaded will be `assets.publishing.service.gov.uk/shared/foo-8d811b8c3badbc0b0e2f6e25d3660a96cc0cca7993e6f32e98785f205fc40907.css`, unless the files are different, in which case they'll have different fingerprints and won't overwrite.
```
fofr: > For example, if both `collections` and `government-frontend` have a file named `foo.css`

As part of this solution we'd need a way to ensure or encourage different frontend apps to use assets in a way that is shareable.

For example, `application.scss` in government-frontend might include a component with some application styles, `application.scss` in collections, could do the same – the same component and its own application styles. That leads to duplication of CSS that users would download twice.

rfc-091-sharing-assets.md
nickcolley: I can help put these measurements together, let me know, I'll try to find some time to do help.

rfc-091-sharing-assets.md
tijmenb: I'm not too worried about this anymore actually. We could easily nuke the bucket and start over by deploying all the applications, so cleaning up would be possible if we ever need it (we don't I think).

rfc-091-sharing-assets.md
tijmenb: While it's true that we don't know exactly how much impact this will have, it's certain that if we were to move all assets to per-app distribution there would be more things to download.

If we look at someone who goes from the homepage to a browse page:

- https://www.gov.uk without cache is 458KB
- https://www.gov.uk/browse/disabilities without cache is 351KB
- https://www.gov.uk/browse/disabilities after visiting the homepage is 80KB

So at the very least this journey would mean 271KB more data traffic for this particular journey. Most of it is the font of course (~260KB) but the principle probably stands.

rfc-091-sharing-assets.md
```diff
@@ -0,0 +1,47 @@
...
+
+We need to share assets to make sure users only have to download a thing once. To do this, we'll upload each application's assets to the same location.
+
+## Problem
+
+In [RFC84 we proposed replacing Static with a gem](https://github.com/alphagov/govuk-rfcs/pull/84). One of the architectural challenges with this approach is that by moving assets (like CSS, JS and images) into the gem, the assets won't be shared for users. This means that users visiting GOV.UK would have to download assets multiple times, depending on what app renders the page.
+
+## Proposal
+
+We upload the assets to a central location. For example, if both `collections` and `government-frontend` have a file named `foo.css`, they will upload the file to `assets.publishing.service.gov.uk/shared/foo.css`.
```
kevindew: The bit that makes me a bit uncomfortable with this is having it be very hard to determine the source of an asset based on the filename.

Eg we'd likely have:
`https://assets.publishing.service.gov.uk/government-frontend/application-00d6906ee26428034a3cc61e0498475b38903f895aeaf5f53222efba375a20c7.css` => `https://assets.publishing.service.gov.uk/shared/application-00d6906ee26428034a3cc61e0498475b38903f895aeaf5f53222efba375a20c7.css`
and
`https://assets.publishing.service.gov.uk/frontend/application-acf275543318182565f911d30e18a81240b22ba94fac1a5b19f9f3c3697b922f.css`
=>
`https://assets.publishing.service.gov.uk/shared/application-acf275543318182565f911d30e18a81240b22ba94fac1a5b19f9f3c3697b922f.css`

but both of those application.css files are very different.

I can't really think of a good centralised way that can be resolved, but it might be nice to have guidelines on the format assets that are publicly exposed are stored so that their paths have more answers.

E.g. we could suggest that assets directories all contain the name of the app (which wouldn't be a concern of a shared resource via a gem) which would then provide clues from the filename of the source file.

nickcolley: For application level styles I'd expect them to always be unique and therefore wouldn't need to be hoisted up to a shared layer?

This RFC only needs to apply to shared styles between applications.

kevindew: I'm not aware that this RFC discriminates between a shared or application specific asset - but probably better for @tijmenb to clarify that.

This line seems to imply all assets though: https://github.com/alphagov/govuk-rfcs/pull/91/files#diff-c53d3b658ba990b1192236d1f1fccfe8R23

nickcolley: Ah okay, then if all assets are hoisted I agree application bundles might need an exception for clarity.

h-lame: In my head I've assumed that we are talking about keeping some separation between the unique to the app assets and the shared ones that previously came from static.  In order to avoid making users download the same assets multiple times depending on the rendering app for some content we should try to make sure that the assets that come from static aren't asset-pipeline/webpack/whatever smooshed together into one file with the assets that come directly from the rendering app.  So that the browser will ask for `/app-name/application.css` (this could be `/shared/application.css and we rely on the fingerprint to distinguish the versions for each application, but I like the idea of having the app name in there somehow to help with identification - and clearing out old ones) and `/shared/shared.css`.  These `/app-name/` assets will likely always be unique to the app, but the `/shared/`ones shouldn't be (depending on the version, but as we upgrade all apps will converge on the same version) so we get a benefit from sharing them.  Of course, I've only mentioned css files here, but I think the same should apply to js, font, images, etc... too.

tijmenb: To be clear, I'm indeed proposing it to work in the way @kevindew describes.

The way I envision this working at the moment is that after Rails precompiles the assets, we will upload the directory (`/public/assets`) to an S3 bucket. We won't touch the source directory. Maybe it feels less uncomfortable to think of the bucket as a kind of cache shared between apps? I'm not sure I have many problems with a bucket full of `application-*.css` files.

h-lame: Ah, ok, so we'll get the benefit of shared versions for things like images and fonts and such, but probably not for JS and CSS bundles, because they'll all be smooshed together by asset-pipeline and each app will probably have their own unique bit of CSS or JS that means their `application.[js|css]` has a different fingerprint, even though 90% of it will be the same as all the other apps?

Actually, I don't know how true that is?  Maybe there aren't many apps that have their own bits of JS + CSS beyond what they currently get from static?  Have we checked?

tijmenb: Ha yes - this is what Paul alluded to in https://github.com/alphagov/govuk-rfcs/pull/91#discussion_r166927276. For this to work you we'll need to put only app-specific things in `application.css`. Shared things need to be compiled separately (this already happens with the font, jquery and govuk-template assets).

rfc-091-sharing-assets.md
MatMoore: Could we send the Capistrano logs to kibana? You should be able to search a version hash and find the thing that created it.

tijmenb: Yes, good idea!

PR #92, opened 2018-06-18
RFC for mandatory primary publishing organisation

MatMoore:
[Rendered markdown](https://github.com/alphagov/govuk-rfcs/blob/ee3260811db459b17aa7360e3df4ec039ebd6e0b/rfc-000-mandatory-primary-publishing-organisation.md)

TLDR:
> Data on content ownership is patchy. We are gradually adding it to the publishing API. We should make sure that new content always has this metadata set.

> If this metadata is not enforced, we can expect data quality to degrade over time as we make changes to publishing apps/content types and forget to set stuff. Incomplete ownership data would affect future work on content management.

Closing Monday 25th June.

### DIFF
--- a/rfc-000-mandatory-primary-publishing-organisation.md
+++ b/rfc-000-mandatory-primary-publishing-organisation.md
@@ -1,0 +1,45 @@
+# Make primary publishing organisation mandatory in content schemas
+
+## Summary
+
+Data on content ownership is patchy. We are gradually adding it to the publishing API.
+We should make sure that new content always has this metadata set.
+
+If this metadata is not enforced, we can expect data quality to degrade over time as we make changes to publishing apps/content types and forget to set stuff. Incomplete ownership data would affect future work on content management.
+
+This RFC proposes to make this field mandatory in the schemas. It does not address validation of the schemas by the publishing API.
+
+## Problem
+
+The data-informed-content team is gradually setting the primary organisation link type for all content on GOV.UK, so that we can help publishers manage their content over time.
+
+Organisations should be able to use GOV.UK publishing tools to manage their "content estates" and effectively prioritise content improvement work. But
+to do that we need to know who owns what.
+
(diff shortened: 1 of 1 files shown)


kevindew:
I think we have a challenge on how to make this mandatory at Publishing API level with schemas

It can be mandatory if it's part of the PutContent endpoint as a required link - but with PatchLinkSet (I'm assuming your planning to use that based on the slack conversation following this) you can't make any links required as in it's patch nature it expects to be able to just change a couple and not the whole set.

MatMoore:
Ah, good point, damn.

Maybe we should make it part of put content then. Can you think of any drawbacks to this approach? Unfortunately we've just gone an added it to a bunch of apps using the patch links method. ☹️

kevindew:
Only drawback with PutContent if I remember correctly is that they don't support child links / hierarchies.

With any mandatory link you always have the problem that it only exists if the document to link it to exists, there are no ways to ensure a link will definitely exist unfortunately.

MatMoore:
I think that's fine, since the link only makes sense if the document exists.

But the downside of this is if there are machinery of government changes, then this data would need to be updated, even though the content itself hasn't changed.

And unfortunately we've also already added a bunch of code to publishing applications that uses the patch links method. 🤔

cbaines:
> Only drawback with PutContent if I remember correctly is that they don't support child links / hierarchies.

... and possibly that changing the link requires publishing a new edition of the document.

MatMoore:
Yeah, this is part of the reason we stuck with patch links when we started this work.

It seems wrong that changing metadata about a document changes the history of the document itself.

MatMoore:
I'm going to reject this RFC in its current form, as I overlooked the 'patch' nature of these links, which is incompatible with making them mandatory at publish time. I'm not convinced that edition links are a perfect fit either, so I'm not going to amend the proposal now.

I still think this thing *should* be mandatory, but I don't have a simple solution that I can make happen this quarter. Next quarter I'm not going to be on GOV.UK, so if we do continue with this idea, someone else will need to take it over. 🏃

PR #93, opened 2018-06-19
Retire GOV.UK CDN Logs Monitor

kevindew:
👉[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/4df73d305361e24d6523e24037bd2886398cbb6a/rfc-093-retire-govuk-cdn-logs-monitor.md) 👈 

A proposal on why we should remove this application and the process we
should undertake.

Closing date: Wednesday 27th June

### DIFF
--- a/rfc-093-retire-govuk-cdn-logs-monitor.md
+++ b/rfc-093-retire-govuk-cdn-logs-monitor.md
@@ -1,0 +1,106 @@
+# Retire GOV.UK CDN Logs Monitor
+
+## Summary
+
+This RFC proposes the retiring of the application
+[GOV.UK CDN Logs Monitor][cdn-logs-repo] for the following reasons:
+
+- We do not make use of the data it generates;
+- It is a complicated tool that few GOV.UK developers understand the purpose of;
+- Similarly, due to lack of knowledge we don't know if the data is accurate
+  when problems occur;
+- Since there is a very infrequent need to use the application there isn't
+  much justification to invest in developers learning it;
+- It uses a significant amount of disk space that requires maintenance.
+
+The purpose of this RFC is to make the case that we should retire it based off
+the knowledge we have. We are hoping that by circulating this suggestion
+across the wider GOV.UK technical community we can identify any issues that
+we haven't yet identified.
+
+## Background
+
+We understand that the original motivation for creating GOV.UK CDN Logs Monitor
+was in response to an incident and it was intended as a means to monitor
+when URLs on GOV.UK change from responding with a success (2xx) status code to a
+different one.
+
+A summary of the responsibilities of the application is as follows:
+
+- Monitor the log files that Fastly sends directly to the box via syslog.
+- Increments statsd counters for the amount of responses the Fastly is serving
+  with a particular status code.
+  E.g. stats.govuk.app.govuk-cdn-logs-monitor.logs-cdn-1.status.200
+- Increments statsd counters for which backend (e.g. origin or mirror) are
+  serving a request.
+- It outputs data to stdout - which subsequently goes to logit - of any
+  requests that are not served by origin or the CDN itself.
+- On a nightly basis various log files are assembled which:
+  - Count how many times a path was accessed via a particular method and
+    backend in an hour.
+  - Store a list of all paths that were accessed successfully that day.
+
+There is more in-depth documentation in the [repo][repo-docs].
+
+## Comparative data
+
+There does not appear to be any tools that are monitoring the graphite data
+that statsd populates. This was checked by searching govuk_puppet for any
+references to the govuk-cdn-logs-monitor namespaces.
+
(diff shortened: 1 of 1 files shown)


sihugh:
I'm trying to remember whether transition logs are processed on logs-cdn-1 as I think they were moved there last year.  I'll have a look in the morning to verify, but if so, we'd need to keep the machine or move them somewhere else.  (They're quite big too).  cc @issyl0

kevindew:
@sihugh Thanks - I hadn't seen those. Yeah I think that might mean I need to ramp down the ideas on this to just retiring the application but not the machine. 

They seem to be using about 8GB so a good half terrabyte less than CDN Logs Monitor.

sihugh:
Yeah maybe they're not _that_ big.  I just remember having to move them because they were too big for the previous thing.  There was a single point of Issy that had to keep zipping them up occasionally

MatMoore:
I agree with the RFC, but mostly out of complete ignorance about what this thing is for and how it's supposed to be used.

I don't think I've ever looked at the metrics this thing creates. The 5xx/2xx counters icinga and grafana use seem to come from a collectd plugin. The limitation of those metrics is you can't easily break the numbers down in any way.

Being able to see whether errors are coming from the mirror seems useful, but only in very rare cases (origin being broken is far more likely and we have other diagnostic info for this).

It would be more useful if we could break down edge errors by application, but the CDN doesn't know what's ultimately going to handle the request. If we could tie together edge and origin requests by govuk_request_id (within kibana), then maybe there's value in capturing the logs. But as currently implemented, I've never been able to get any useful information from them.

> We understand that the original motivation for creating GOV.UK CDN Logs Monitor was in response to an incident and it was intended as a means to monitor when URLs on GOV.UK change from responding with a success (2xx) status code to a different one.

What was the impact of the incident? How do you think we'd respond to it if it happened today?

issyl0:
@sihugh Yes, transition logs are processed on logs-cdn-1. I really quite miss the SPOI zipping them up when they were on transition-logs-1, it was relaxing.

rfc-093-retire-govuk-cdn-logs-monitor.md
```diff
@@ -0,0 +1,106 @@
...
+- It uses a significant amount of disk space that requires maintenance.
+
+The purpose of this RFC is to make the case that we should retire it based off
+the knowledge we have. We are hoping that by circulating this suggestion
+across the wider GOV.UK technical community we can identify any issues that
+we haven't yet identified.
+
+## Background
+
+We understand that the original motivation for creating GOV.UK CDN Logs Monitor
```
tijmenb: These reports are relevant I think:

- https://docs.google.com/document/d/1rNLjPb5mYYJoURTDoDlQQ9-DM1TilKEm58zbBP8sBnY/edit
- https://docs.google.com/document/d/1Ev_axmMdvsg3WTnYpdYVBciSAzM3q64QklB8MOZl7Go/edit

rfc-093-retire-govuk-cdn-logs-monitor.md
```diff
@@ -0,0 +1,106 @@
...
+The [Future steps](#future-steps) section of this document explores using
+[AWS Athena][] as a means to query for the data sources that are lost.
+
+## Data usage
+
+This application currently uses 413GB on `logs-cdn-1` and stores > 100GB of
+Graphite databases on `graphite-1`. A significant portion of the graphite storage
+is due to unnoticed misconfiguration.
+
+## Proposal
```
tijmenb: I wonder if we can split this up into 2 things:

1. Retire the cdn-logs-monitor
2. Move CDN logs to S3

I think moving to S3 might need more thought, I've found it useful at times to analyse or tail the cdn logs using grep.

kevindew:
Great thanks for the comments so far.

Given the comments so far it does seem too ambitious to try to remove the machine given it's transition usage.

I'm going to edit the scope of this RFC to reflect this and aim to archive the application. I was wrong in my earlier assessment of disk usage of transition as they use many GB of files in the cdn-logs directory. I think what might be a good idea is to start sending the logs to S3 and continue to send them to the cdn box, but make the time the logs are stored lowered on this box which can act as a means to transition to a s3 / athena world without losing current practices.

rfc-093-retire-govuk-cdn-logs-monitor.md
kevindew: Awesome thanks 👍

kevindew:
I've updated this RFC now to reflect feedback.

boffbowsh:
I agree, and I think we need to set up Glue / Athena ASAP and start people using it and getting familiar with it. It can be more powerful than grep/awk etc (and I say that as a proper awk nerd) if configured correctly, especially using a columnar format such as Parquet. 

Interestingly, with Athena we can easily generate a nightly report of the same format the logs monitor already gave us :)

PR #96, opened 2018-10-29
Scheduled publication functionality in Publishing API

thomasleese:
👉 [Rendered Version](https://github.com/alphagov/govuk-rfcs/blob/scheduled-publication-functionality-to-publishing-api/rfc-096-publishing-api-scheduled-publications.md)

### DIFF
--- a/rfc-096-publishing-api-scheduled-publications.md
+++ b/rfc-096-publishing-api-scheduled-publications.md
@@ -1,0 +1,45 @@
+# Scheduled publication functionality in Publishing API
+
+## Summary
+
+Add scheduled publication functionality to the Publishing API.
+
+## Problem
+
+We have certain scheduled publication formats that we must represent live on GOV.UK within a minute of the scheduled publication time (such as statistical publications).
+
+Scheduled publications are not currently a feature supported by the Publishing API. This means each publishing application that wishes to implement scheduled publishing must do so independently. This leads to duplicate functionality being developed in each publishing application that required the functionality. At the same time, although conceptually each application would develop similar functionality, the code surrounding scheduled publications might vary between applications.
+
+This has several drawbacks:
+
+  * Developers must become familiar with multiple implementations that solve the same problem.
+  * Duplicate functionality across multiple applications violates a Don't Repeat Yourself philosophy.
+  * Debugging problems with scheduled publications across multiple micro service layers is difficult.
+  * There are multiple places where content may get "stuck" and debugging requires knowledge of individual publishing apps.
+
+Whitehall is one of the publishing applications that implements scheduled publishing. In Whitehall's case, the code that interacts with the Publishing API is often difficult to reason about and hard to debug.
+
+## Proposal
+
+Create scheduled publishing functionality in the Publishing API.
+
+- Publishing API MUST provide the ability for publishing apps to schedule a publish of draft content.
+- Publishing API MAY provide a `/v2/content/:content_id/schedule` endpoint.
+- Publishing API MAY introduce a 'scheduled at' date field to editions.
+- Publishing API MAY implement callbacks or push on to message queue.
+- Publishing API MAY implement an endpoint to poll for an item's state (e.g. when in Whitehall admin, as you hit an edition, it polls the Publishing API to check whether it's been published).
+- Publishing API MUST implement a queue of scheduled content (probably use Sidekiq `perform_at` ? as in Whitehall).
+- Publishing API MUST provide the ability to cancel a scheduled publish.
+- Publishing API SHOULD store scheduled requests in the actions log (e.g. `Schedule` and `CancelSchedule`).
+- Publishing API SHOULD allow subsequent requests to overwrite existing scheduled publish requests.
+- Publishing API MUST cancel any scheduled publish requests if the draft edition is discarded.
+- Publishing API MUST cancel any scheduled publish requests if the draft edition is published directly.
+- Whitehall MUST be migrated to use the new Publishing API functionality.
+- Publisher MUST be migrated to use the new Publishing API functionality.
+- Specialist Publisher MAY get scheduled publishing functionality added using the Publishing API.
+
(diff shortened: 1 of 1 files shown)


rfc-096-publishing-api-scheduled-publications.md
```diff
@@ -0,0 +1,45 @@
...
+- Publishing API MAY introduce a 'scheduled at' date field to editions.
+- Publishing API MAY implement callbacks or push on to message queue.
+- Publishing API MAY implement an endpoint to poll for an item's state (e.g. when in Whitehall admin, as you hit an edition, it polls the Publishing API to check whether it's been published).
+- Publishing API MUST implement a queue of scheduled content (probably use Sidekiq `perform_at` ? as in Whitehall).
+- Publishing API MUST provide the ability to cancel a scheduled publish.
+- Publishing API SHOULD store scheduled requests in the actions log (e.g. `Schedule` and `CancelSchedule`).
+- Publishing API SHOULD allow subsequent requests to overwrite existing scheduled publish requests.
+- Publishing API MUST cancel any scheduled publish requests if the draft edition is discarded.
+- Publishing API MUST cancel any scheduled publish requests if the draft edition is published directly.
+- Whitehall MUST be migrated to use the new Publishing API functionality.
```
cbaines: Note that the way in which Whitehall handles change notes for drafts currently blocks this.

I tripped over this recently, so there's a comment in the code now describing this: https://github.com/alphagov/whitehall/blob/516c8154f19af2a68815cb10cae07b7f87847128/lib/whitehall/publishing_api.rb#L16-L25

kevindew: It's probably also blocked by links in Whitehall too if I remember correctly (as Whitehall uses patch links). So if you were to change links on a document on Whitehall it doesn't apply links until the point of publishing when there is an existed published item otherwise those link changes would affect the published item.

rfc-096-publishing-api-scheduled-publications.md
```diff
@@ -0,0 +1,45 @@
...
+- Publisher MUST be migrated to use the new Publishing API functionality.
+- Specialist Publisher MAY get scheduled publishing functionality added using the Publishing API.
+
+Whitehall currently doesn't store the full state history of content that has been published automatically after being scheduled. We also think it won't be necessary to store this in the Publishing API. A piece of content that has a draft edition can be scheduled for publish, and that same edition may continue to be updated up until it gets published. Whether or not an edition has a scheduled publish request attached to it, doesn't affect the state of the edition until it gets published.
+
+With data in the Publishing API, we might be able to improve the [Scheduled Publishing dashboard][dashboard] to include visibility on the queues, the number of items scheduled for publishing and when the content is going to be published.
+
```
kevindew: It might be useful to have a section on how this resolves the current drawbacks (which I guess is mostly why it would be simpler in Publishing API than how it is in Whitehall - given Whitehall is the primary case) and also the drawbacks of this approach (which most notably is that state can change in Publishing API without direct application interaction)

tijmenb:
Thanks for this RFC! 🤘

I've pondered on this since last week, and I think I'm not yet sold on the approach. The current problems are evident, but I'm not sure if this solves them. The questions I have are mainly inspired by the work we've been doing on Content Publisher.

## Does it solve duplication?

The 'Problems' section describes the problem this RFC is trying to solve mainly in terms of preventing duplication.

- Have other approaches been considered? There could be other ways to solve this problem, for example requiring all applications to implement the functionality in the same way (a standard), or sharing code using a shared library (a gem like govuk_sidekiq).
- It seems like significant amount of code will be to required communicate with the publishing-api. This code will also be duplicated, so would this approach exchange one type of duplication with another?
- If removing potential duplication is the main goal, then we have to convert both Publisher and Whitehall, requiring significant effort. Is this worth it, as the long term goal is to get rid of these applications? If we're not converting all existing applications, then is it still worth it?

## Does it make things simpler?

We've been moving away from centralising functionality in the publishing-api. As [ADR 2 in Content publisher](https://github.com/alphagov/content-publisher/blob/master/docs/adr/0002-use-local-datastore.md) explains, content publisher uses publishing-api only to publish content.

- The polling endpoint is probably not enough, we'll need a more proactive notification when something is scheduled, like a webhook system that notifies publishing applications. Content publisher needs this to create a history entry and update the internal state to make searches work. This will add a significant amount of code, and also increases the risk that something goes wrong.
- The suggested approach adds complexity to testing. Currently, testing scheduled publishing can be done mostly in the application using integration and unit tests. If it lives in the publishing-api, we have to drop down to API-stubbing, which is easy to do wrong. This is one of the reasons for content publisher to use a local data store. Additionally I'm unsure about implementing end-to-end tests for this feature, because of the requirement to travel through time.
- A local solution limits the number of calls between publishing-api and application, limiting the number of things that can go wrong. Applications will have to consider failures in the `/schedule` endpoint, the polling endpoint and cancel endpoint.  
- The concept of scheduled publishing will live in two places, even if the actual code that does the publishing lives in the publishing-api. This means developers will have to understand both the publishing-api and the application they're working with.

## Does it solve the right problem?

I wonder if we know enough about the current problems to know what the correct solution is.

- There are 2 implementation of scheduled publishing, in Publisher and Whitehall. As far as I know, the Publisher implementation has significantly less problems than Whitehall. Why is this?
- I'm keen to understand the actual problems experienced in Whitehall. Do we have a list of incidents and root causes that would have been prevented by the approach suggested in this RFC?
- Conceptually scheduled publishing isn't difficult (it's emulating a button click at a certain time). For [content publisher we prototyped this approach](https://github.com/alphagov/content-publisher/compare/add-scheduling?expand=1) to scheduled publishing. I wonder if that would have the same problems?

## Will it work for existing apps and content publisher?

- As @cbaines mentions, some [pre-publish code is run in Whitehall](https://github.com/alphagov/whitehall/blob/516c8154f19af2a68815cb10cae07b7f87847128/lib/whitehall/publishing_api.rb#L16-L25) that we have to take into account. In Publisher, [some link patching happens](https://github.com/alphagov/publisher/blob/93cf94480c42ea89b7095790a29384d9d5642127/app/services/publish_service.rb#L20-L25) at publish time. 
- When publishing something in content publisher it [first publishes the assets](https://github.com/alphagov/content-publisher/blob/dcc33d8cc24e92b428b632ee830c1d8d711055f8/app/services/document_publishing_service.rb#L16-L17). How can this be implemented with scheduling in the publishing-api?

PR #97, opened 2018-11-14
HUB-341: Start pages for GOV.UK Verify

jakubmiarka:
👉 [Rendered Version](https://github.com/alphagov/govuk-rfcs/blob/hub-341-start-pages-for-verify/rfc-097-verify-specific-start-pages.md)

This document describes an approach for providing alternative service start pages
that will be displayed to users on a GOV.UK Verify single IDP journey.

Ported from `verify-architecture` repo, authored by Chris Clayson.

### DIFF
--- a/rfc-097-verify-specific-start-pages.md
+++ b/rfc-097-verify-specific-start-pages.md
@@ -28,114 +28,25 @@ Verify and, hence, their chosen IDP.
 
 #### Discoverability
 
-Any customised content is designed to be reachable solely via hyperlinks from IDPs, which will redirect via the Verify hub.
+Any customised content is designed to be solely reachable, indirectly via the hub, using hyperlinks IDPs publish to their users.
+
 Therefore the new content MUST not be indexed and reachable by performing, for example, a Google search or GOV.UK search. 
 
+#### Content Schema
 
-#### Content Schemas
-
-There are two GOV.UK content schemas used for service start pages. Most services use the `transaction` schema but some use the 
-`guide` schema. Any solution MUST work for, at least, both these content schemas.
-
+The GOV.UK content schema used for service start pages is the `transaction` schema, the changes proposed will require changes to this schema which will, potentially, require changes to be made to existing data that conforms to this schema.
 
 ## Proposal
+  
+The solution would be for GOV.UK to support 'variants' of content in the transaction content schema. This should allow a variant of the page to be requested in the URL, as shown, `http://www.gov.uk/<base_url>/<variant-name>`. For example, `http://www.gov.uk/my-rp-transaction-page/verify`.
 
-### Short term/interim solution - Alternate Content
+This would require multiple changes to GOV.UK including schemas, publisher and government frontend.
 
-In the short term, and in the interests of getting up and running quickly, we can simply take a copy of the existing content in
-GOV.UK's content publisher service, alter it, and publish it with a different URL slug (as specified in the `base_url` property 
-of the content item. A suggestion would be to append `-verify` to the the end of the existing slug.
-
-For example:
-```
-// Original Content Item
-{
-    "analytics_identifier": null,
-    "base_path": "/my-rp-transaction-page",
-    "content_id": "abcd1234-5678-90ab-cdef-0123456789ab",
-    ...
-}
-```
-
-```
-// Alternate Content Item
-{
-    "analytics_identifier": null,
-    "base_path": "/my-rp-transaction-page-verify",
-    "content_id": "abcd1234-5678-90ab-cdef-abcdef123456",
-    ...
-}
-```
-
-[**needs confirmation**] This solution works well for `transaction` content schema, and looks like it should work with `guide` although more care will 
-need to be taken the latter to ensure all slugs are updated and any explicit hyperlinks updated also.
-
-#### Searching
- 
-[**for discussion**] The GOV.UK team state there is an existing method for ensuring that these alternate content pages are not searchable
-within the GOV.UK website, a change will need to be implemented on GOV.UK to ensure the these pages are not indexable by 
-search engine web crawlers. Even though the new alternate content will not be linked from anywhere, so would not be discoverable by crawlers, 
-GOV.UK does publish a `sitemap.xml`. We need to ensure these alternate content pages are not published in `sitemap.xml` and/or include the
- `ROBOTS` meta tag with a `NOINDEX` value.
- 
-### Long term solution - Content Variations
-
-A long term solution would be for GOV.UK to support 'variants' of content in the content schemas. This should allow a variant of
-the page to be requested in the URL, for example, `http://www.gov.uk/<base_url>/<variant-name>`
-
-For example, `http://www.gov.uk/my-rp-transaction-page/verify`.
-
-An example of how the content schema for `transaction` could be updated is shown here:
-
-```
-{
-  // Existing
-  ...
-  "details": {
-    "introductory_paragraph": "<p>The services introduction</p>\n",
-    "start_button_text": "Start now",
-    "will_continue_on": "the RP start page",
-    "transaction_start_link": "https://www.my-service.gov.uk/start",
-    "more_information": "<p>Additional Info</p>\n",
-    "external_related_links": [],
-    "department_analytics_profile": ""
-  }
-  ...
-}
-```
-
-```
-{
-  // Including a variant
-  ...
-  "details": {
-    "introductory_paragraph": "<p>The services introduction</p>\n",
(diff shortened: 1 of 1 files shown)


kevindew:
Quick question from skimming the RFC - what is an IDP? I guessed at Identity Provider?

rfc-097-verify-specific-start-pages.md
```diff
@@ -0,0 +1,136 @@
...
+[**for discussion**] The GOV.UK team state there is an existing method for ensuring that these alternate content pages are not searchable
+within the GOV.UK website, a change will need to be implemented on GOV.UK to ensure the these pages are not indexable by 
+search engine web crawlers. Even though the new alternate content will not be linked from anywhere, so would not be discoverable by crawlers, 
+GOV.UK does publish a `sitemap.xml`. We need to ensure these alternate content pages are not published in `sitemap.xml` and/or include the
+ `ROBOTS` meta tag with a `NOINDEX` value.
+ 
+### Long term solution - Content Variations
+
+A long term solution would be for GOV.UK to support 'variants' of content in the content schemas. This should allow a variant of
+the page to be requested in the URL, for example, `http://www.gov.uk/<base_url>/<variant-name>`
```
cbaines: Content on GOV.UK can already register routes for multiple paths. Rendering apps can then process this however they want.

A common example is the Travel Advice pages, these two pages relate to the same content item:
 - https://www.gov.uk/foreign-travel-advice/france
 - https://www.gov.uk/foreign-travel-advice/france/safety-and-security

I'm not sure how well this fits with what you're trying to do though...

erino:
Also following on from @kevindew's comment what's an RP?

jakubmiarka:
@kevindew @erino: Apologies for not being clearer with the SAML abbreviations. It was originally written in the Verify repo.

- IDP is Identity Provider
- RP is Relying Party (a service in our context)

rfc-097-verify-specific-start-pages.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Summary
+
+This document describes an approach for providing alternative service start pages 
+that will be displayed to users on a GOV.UK Verify single IDP journey.
+
+## Problem
+
+When following a single IDP journey, the user needs to be redirected via the RP (service) in order
+to pick up a valid SAML request. Initial thoughts were to direct the RP headless start page,
```
bevanloon: I wonder if you could expand on this a little (maybe with some more specific examples spelled out); It strikes me that there's quite a lot implied in this sentence. I'm not really familiar with user journeys or the meanings of the specific vocabulary in Verify (e.g. what is an Identity Provider or Relaying Party?) and so I think I'm missing some important context about the actual need/problem we are trying to address with this RFC. 

It would also be great if you could pop the full version of those acronyms into the text as well please.

chrisclayson: I've push a revised version with additional text to help explain this. I'm to come over and show a demo if it makes it clearer.

bevanloon: 👍 thanks that seems clearer now

rfc-097-verify-specific-start-pages.md
rubenarakelyan: A version of this is how I envisage these start pages working in practice. Basically, get Mainstream Publisher to register the "variant" routes at the same time and route them to Government Frontend as usual where they will be picked up and rendered.

The part that needs work is storing the variant content somewhere in the original content item so Government Frontend can render it without too much custom/hardcoded logic (just enough to understand how to get and display the variant content).

rubenarakelyan:
tl;dr: the "long-term" solution makes most sense to me and I think it's the right thing to do for what you're trying to achieve.

My preference here is what you're terming the "long-term solution", but I don't think it will take as much time as you imagine.

As Chris mentioned in his comment above, there is already precedent for publishing apps to register more than one route for the same content. In effect, this is what you're referring to as "variants". The content for all the pages is in one content item, but that content item can be accessed via multiple base paths, each one of which picks which part of the content item to render.

By taking this route, you'd just be extending what is already an established pattern, just for a different type of content. In fact, the JSON schema for this, which currently refers to "parts", could probably be somewhat reused, maybe with a few more fields that are specific to start pages.

You'd have to build functionality into Mainstream Publisher to be able to publish such multi-part content items for start pages. Government Frontend will also need some changes, but since it already knows about multi-part pages, there work here is probably much smaller in scope. Finally, you need to add this into the content schemas, but again the work there shouldn't be too much.

kevindew:
Both approaches here sound pretty reasonable, where I'm a bit unclear is how many data is shared between these content items/variants.

Your first approach sounds good if the amount of data shared between the two content items is low to zero. If these can affectively live as two separate things then that's great and is the approach you should use. Whereas if there's a bunch of data to be shared between them meaning you have to co-ordinate publishing both at the same time, then that's an area of concern.

With the approach you've suggested with variants, like others have pointed out there's prior art in having multiple pages with a single content item - you wouldn't need to create a whole new section of a schema for variants it could go within details (which would need to be customised if you have any additional custom data). Though you should only go down this path if there is a reasonable amount of shared data as content items that serve multiple paths are arguably not ideal and are a source of confusion. Life is simpler the closer we are to single URL -> single content item mapping.

Finally the other option that hasn't been discussed here is the use of linking to share data. If there is a clear main page and then other sub-pages, then the sub pages could have a link to the master page and you could specify which data fields you want to be included in the content item - See: https://github.com/alphagov/publishing-api/blob/57628918fbb2572d37bd85bdbe0e3d9a04e235bb/lib/expansion_rules.rb#L58-L77. This is probably the most GOV.UK architecturally sound way to share data between pages but I'm not sure it's right for your scenario.

rubenarakelyan:
I think the best way to think of this is not as a parent page with a set of sub pages, but multiple parent pages with a high level of shared content and some (minor) differences. They each cater to a slightly different audience but convey mostly the same information.

In my view the single content item makes sense because the vast majority of the content will be identical, and the variants mostly deal with where the user goes next in their journey. I don't think we currently model this kind of relationship with links and so would probably have to define a new link type.

rfc-097-verify-specific-start-pages.md
cbaines: > The part that needs work is storing the variant content somewhere in the original content item so Government Frontend can render it without too much custom/hardcoded logic (just enough to understand how to get and display the variant content).

Well, crossing what's described here with what Government Frontend/Travel Advice Publisher do for travel advice pages seems sensible. So having something in the details, but maybe calling it variants rather than parts.

cbaines:
> The consensus is that we go with the single content item --> multiple base paths route with variants as required.

Given the RFC currently doesn't say this, or at least differs in the technical details from what's been discussed in the comments, I'm a bit uncertain about it being merged?

rubenarakelyan:
@cbaines I've asked the Verify team to clean up the RFC before merging it. The "long-term" solution currently in the document is basically what's been agreed on, so they'll remove the "short-term" one and clean up the discussion items.

rfc-097-verify-specific-start-pages.md
```diff
@@ -0,0 +1,51 @@
...
+
+The approach described in this document is to implement alternate service start pages, that
+will display all important information to the user but only give them the option of using
+Verify and, hence, their chosen IDP. 
+
+### Considerations
+
+#### Discoverability
+
+Any customised content is designed to be reachable solely via hyperlinks from IDPs, which will redirect via the Verify hub.
```
jakubmiarka: is this accurate? the customised content will be only reached via redirects from hub. Only the hub knows the URLs.

rfc-097-verify-specific-start-pages.md
```diff
@@ -0,0 +1,51 @@
...
+
+#### Content Schema
+
+The GOV.UK content schema used for service start pages is the `transaction` schema, the changes proposed will require changes to this schema which will, potentially, require changes to be made to existing data that conforms to this schema.
+
+## Proposal
+  
+The solution would be for GOV.UK to support 'variants' of content in the transaction content schema. This should allow a variant of the page to be requested in the URL, as shown, `http://www.gov.uk/<base_url>/<variant-name>`. For example, `http://www.gov.uk/my-rp-transaction-page/verify`.
+
+This would require multiple changes to GOV.UK including schemas, publisher and government front end.
```
jakubmiarka: *frontend

rfc-097-verify-specific-start-pages.md
rubenarakelyan: We can handle this in the frontend by serving robots tags for search engines. For internal search, this is only one content item so only the "parent" start page would be indexed by default.

PR #98, opened 2019-02-07
RFC 98: Implement Content Security Policy in apps

tijmenb:
The purpose of this RFC is to document why we're going to add a Content Security Policy, and to get consensus on how to implement it.

👉[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/csp-on-cdn/rfc-098-csp.md)

### DIFF
--- a/rfc-098-csp-on-cdn.md
+++ b/rfc-098-csp-on-cdn.md
@@ -1,0 +1,59 @@
+# RFC 98: Implement Content Security Policy on the CDN
+
+## Summary
+
+This RFC proposes to use our CDN to configure a Content Security Policy for GOV.UK.
+
+## Background
+
+We'd like to implement a Content Security Policy (CSP) on www.gov.uk.
+
+> The HTTP Content-Security-Policy response header allows web site administrators to control resources the user agent is allowed to load for a given page. With a few exceptions, policies mostly involve specifying server origins and script endpoints. This helps guard against cross-site scripting attacks (XSS).
+
+https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
+
+CSP works by sending a header with each HTTP response. It looks something like this:
+
+```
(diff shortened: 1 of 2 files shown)


kevindew:
> * Does this RFC make clear why we want to implement CSP?

Yes, though it doesn't make clear exactly what we want to apply it to (just www.gov.uk or beyond)

> * Are there any missing trade-offs in the implementation options?

It doesn't make it too clear that the configuration will only apply to the www.gov.uk and not draft stack or any of the admin apps.
Nginx would also need local configuration to support dev VM.

> * Do you agree with making the trade off in configuring it at the CDN level?

It seems pragmatic but I think it'd be more consistent for GOV.UK to be done at the Nginx level and managed via Puppet, and that'd allow us to apply it to Non-CDN apps without two approaches.


Other things I wondered where this could break things: embedding of youtube videos, and since our markdown supports basic image syntax do we know if any publishers use it? (I guess we're hoping not)

rubenarakelyan:
I also think it would make things more consistent to have this in nginx config (akin to how we do STS).

YouTube video embedding should be simple to do since it all comes from one domain (that we can whitelist). Images, however, would be very difficult if we allow people to include them from any domain, unless we whitelisted everything (which would make that part of the CSP redundant). It could work if we proxied such images through our own servers.

tijmenb:
I'm starting to lean towards adding the CSP on the app level. The experiment with CSP on the CDN yielded a absolute ton of violations, most of them caused by browser extensions.

image

It might be easier to implement CSP app-by-app, so that we can more easily test it. I've raised https://github.com/alphagov/government-frontend/pull/1250 as an experiment. I'll update the RFC is that is successful.

rubenarakelyan:
I found this blog post (https://www.troyhunt.com/add-ons-extensions-and-csp-violations-playing-nice-with-content-security-policies/) that also talks about browser extensions causing loads of CSP violations. He's filtering them out and also actually contacting the extension developers to ask them to fix their extensions.

PR #99, opened 2019-02-07
RFC 099: Q&A Forum for GDS developers

bilbof:
The purpose of this RFC is to gauge interest in trying out [StackOverflow for teams](https://stackoverflow.com/teams).

[Read the rendered version](https://github.com/alphagov/govuk-rfcs/blob/473d014b1ebace8da651906d50a44a387252adba/rfc-098-stack-overflow-for-gds.md)

### DIFF
--- a/rfc-099-stack-overflow-for-gds.md
+++ b/rfc-099-stack-overflow-for-gds.md
@@ -1,0 +1,41 @@
+# Q&A Forum for GDS developers
+
+## Summary
+
+A searchable, long-lived body of developer knowledge would make finding
+answers easier.
+
+StackOverflow is a Q&A forum that all developers at GDS are familiar with.
+A private instance of StackOverflow would ensure that valuable knowledge
+is retained and more easily found.
+
+## Problem
+
+When faced with a problem we're not able to solve, we ask for help from
+other developers in person or over Slack.
+
+Both of these methods mean that developer knowledge is quite often lost,
+and time is spent finding answers to the same problems. The solutions
+remain in Slack, and aren't easily retrievable.
+
+## Proposal
+
+I propose GDS uses [StackOverflow for Teams](https://stackoverflow.com/teams).
+
+As developers, we all know the value of StackOverflow. I've used their Teams
+product before and have found it very useful.
+
+StackOverflow allows for org-specific tagging (e.g. `publishing-api`).
+
+Developers can gain karma, and be more easily recognised for helping others.
+
+This should result in more documentation. Questions and answers will be
+preserved, and frequently asked questions will be more visible and can
+be written up into documentation.
+
+It's possible to export data from StackOverflow for Teams, even after we
+stop paying for it, which will help to prevent lock-in.
+
+You can integrate StackOverflow with Slack, and post new questions into channels.
+
+There is a 14 day free trial so we could try out it out.


rfc-099-stack-overflow-for-gds.md
```diff
@@ -0,0 +1,41 @@
...
+## Proposal
+
+I propose GDS uses [StackOverflow for Teams](https://stackoverflow.com/teams).
+
+As developers, we all know the value of StackOverflow. I've used their Teams
+product before and have found it very useful.
+
+StackOverflow allows for org-specific tagging (e.g. `publishing-api`).
+
+Developers can gain karma, and be more easily recognised for helping others.
```
elliotcm: I'm a little bit wary of how this leads to in-group/out-group stuff on SO proper, so as long as we can avoid importing that culture it's probably fine to have the karma system.

rfc-099-stack-overflow-for-gds.md
```diff
@@ -0,0 +1,41 @@
...
+A searchable, long-lived body of developer knowledge would make finding
+answers easier.
+
+StackOverflow is a Q&A forum that all developers at GDS are familiar with.
+A private instance of StackOverflow would ensure that valuable knowledge
+is retained and more easily found.
+
+## Problem
+
+When faced with a problem we're not able to solve, we ask for help from
```
elliotcm: We (in govuk) also maintain [significant developer docs](https://docs.publishing.service.gov.uk/), however it's an intermingling of background documentation and answers to specific issues, so it would probably be of benefit to have a different system for the latter type.

It also has bad / weird search.

rfc-099-stack-overflow-for-gds.md
bilbof: I agree! Hopefully people will ask/answer questions in the same manner as they do on Slack.

emmabeynon:
On the whole, I'm in to this.  I really hope that this will help gather knowledge that is stored away in people's minds.  To @elliotcm's point, I think we'll need some clear guidance as to what should live in the dev docs and what should live in Stack Overflow to avoid any muddying of the waters and confusion over where a developer should go to find something out.  I've not used the teams version before, so would be curious to know from your experience @bilbof how upvoting/downvoting was managed?  I think we've got a really nice, supportive culture among GOV.UK developers and I'd be wary of introducing a system that might undermine that.

There's a possible alternative to using Stack Overflow teams which I've seen in the past that could help solve the problem at hand.  It's essentially a private repo where people can open up questions as issues and others can respond using comments.  It can make use of tagging of course and there's an element of searchability.  It does however probably lack the social networking element of Stack Overflow e.g. karma.

tijmenb:
Thanks @bilbof for proposing! I think this raises a lot of interesting questions. 

As @elliotcm mentions, the developer docs is currently the place for long-lived developer knowledge. Some of the upsides are that it's public, it's programmable (it pulls in docs from other repos), and it uses an existing system that is usable by all developers. Incidentally, the bad/weird search was replaced last week, so hopefully that's better/nonweirder now.

I really like @emmabeynon's suggestion. Having an GitHub-issue-based Q&A should solve some of the problems this RFC raises, and we can adopt it fairly easily. It suffers less from network effects than StackOverflow, which might only be useful once there's a lot of content on it (which will take more than 14 days).

We could put the questions-issues in the main developer docs repo (then if a particular issue has useful answers, we can create a page for it in the docs).

If we opt to do this, we'll have to:

- Make it clear to developers to *only* ask questions in https://github.com/alphagov/govuk-developer-docs/issues
- Use 👍 to vote for the "correct" answer (if applicable)
- Set up a Slack integration that posts dev docs issues into #govuk-developers for visibility
- Encourage people to answer

elliotcm:
The new dev docs search is definitely better so far.

bilbof:
Thank you for your comments! I am keen to try out SO, whether for the 14 day free trial or for a paid trial if we need a longer period to get it going.

> I've not used the teams version before, so would be curious to know from your experience @bilbof how upvoting/downvoting was managed?

@emmabeynon I don't think the voting feature can be disabled; I didn't notice voting being an issue, though. I think if our Github/Slack Q&A community is good natured, then I imagine this would continue if we adopted SO.

> a private repo where people can open up questions as issues and others can respond using comments

@emmabeynon @tijmenb I can see the value in have a private Github repo for Q&A.



I think we could take even more value from StackOverflow, as beside feature parity (slack integration, labels, voting, comments, notifications, markdown support), it comes with features that a Github issue doesn't have. For the most part this is because SO is designed for the Q&A format, including:

- SO permits comments on questions and answers, whereas a Github issue comment is not necessarily an answer.
- In SO when drafting a question one is helped to write a good question and apply helpful labels, and automatically prompted to check if the same question has already been asked. This will prevent people solving the same issue twice (which is also a good reason SO could be better than Slack as a place to ask questions).
- Good search based around questions and answers (e.g. searching by unanswered and so on, but also a good search syntax like url:"*.data.gov.uk" infavorites:mine, is:question hasaccepted:yes)
- A browse page designed for questions, whereas Github issue browsing is designed around project management and reviews 
screen shot 2019-02-21 at 14 09 28
- I also think that SO's features designed around rewarding people who both answer and ask questions would encourage people to help each other.

timblair:
I like the concept of having somewhere to go to ask questions, but I don't like the idea of adding yet _another_ disparate source of documentation, at least without the value being proved first.

My issues with SO for Teams specifically:

* There's a monetary cost to this: quick calculations would be between £2–3k/yr for GOV.UK devs only.
* It's locked down to just users with accounts, and more users == more £. We see pain with lock-out of non-GOV.UK people trying to find valuable information that's in the GOV.UK wiki, which is similarly locked down. It also means that people outside of GOV.UK can't answer questions on which they might be best-placed to provide input.
* It's not currently an IA-approved tool, so would need to go through an approval process before being used.

I think trialling a Q&A-style system is a good idea, but we should do it using the tools we already have access to to prove the value.

rfc-099-stack-overflow-for-gds.md
nickcolley: Agree with Elliot, not sure we should be adopting something built by people who do not value healthy diverse communities.

bilbof:
Thanks, @timblair. It sounds like there is interest in having a Q&A forum, and that a Github repository would be the right thing to use for this.

That is probably worth discussing at the next tech fortnightly meeting, so I'll close this RFC down for now.

Thanks everyone for your comments!

chao-xian:
Apologies for not commenting before the close date, but I just wanted to add that for me the kudos system is going to be detrimental to people who have imposter syndrome as it will discourage them from ever contributing. The kudos system to me is a detractor from the willingness to just help each other. The context of the kudos system for SO in the outside world is as signal of trustworthiness and authority on the subjects, but within the context of our workplace and with our colleagues there is no need for it. It should always be encouraged that every contribution is valid and valuable, and not gamified for points. :v:

bilbof:
Thank @chao-xian, that makes sense and I agree that kudos isn't necessary for an internal Q&A. This sounds like a plus for using a Github repo as a forum, as there's no way to accumulate points like SO kudos :)

PR #100, opened 2019-02-20
RFC 100: Improve linting

tijmenb:
This RFC proposes a change our linting systems.

It is based on work by @theseanything to implement linting with Rubocop directly (https://github.com/alphagov/content-performance-manager/pull/1082), @alex-ju's work to implement frontend linting in the components (https://github.com/alphagov/govuk_publishing_components/pull/758) and @cbaines work to [properly lint all the repos](https://github.com/pulls?q=is%3Apr+author%3Acbaines+archived%3Afalse+linting+is%3Aclosed).   

👉[View it rendered](https://github.com/alphagov/govuk-rfcs/blob/linting/rfc-100-linting.md)

### DIFF
--- a/rfc-100-linting.md
+++ b/rfc-100-linting.md
@@ -1,0 +1,53 @@
+# RFC 💯: Update linting system
+
+## Summary
+
+Update the way we do linting by using Rubocop directly and using NPM-packages to lint CSS and JS.
+
+## Background
+
+At present, we use a gem called [govuk-lint][] for linting in our projects.
+
+- It's a wrapper around [rubocop], a community driven styleguide for Ruby, so that we can have [central configuration files][rules] that specify our linting rules
+- It also adds the [--diff --cached commands][commands] to rubocop, which allow us to lint only the changed files
+- It provides a wrapper for [scss_lint][] to provide [shared rules][css-rules] for CSS. These tools are used by the Design System team, are well maintained and have an autocorrect feature.
+
+We [wrote about introducing linting](https://gdstechnology.blog.gov.uk/2016/09/30/easing-the-process-of-pull-request-reviews/) in 2016.
+
+[govuk-lint]: https://github.com/alphagov/govuk-lint
+[rubocop]: https://github.com/rubocop-hq/rubocop
+[scss_lint]: https://github.com/brigade/scss-lint
+[rules]: https://github.com/alphagov/govuk-lint/tree/master/configs/rubocop
+[commands]: https://github.com/alphagov/govuk-lint#ruby
+[css-rules]: https://github.com/alphagov/govuk-lint/blob/9c501a15824a156718d58e7e1a107a7d78171c5f/configs/scss_lint/gds-sass-styleguide.yml
+
+## Issues
+
+1. Rubocop now has a native way of sharing rules: you include rules by specifing `inherit_gem` in `.rubocop.yml`. This has been successfully [trialed in content-performance-manager](https://github.com/alphagov/content-performance-manager/pull/1082). As mentioned in that PR, this makes the linting faster and makes it compatible with developer environments (see [this issue from 2016](https://github.com/alphagov/govuk-lint/issues/61) for evidence of a need).
+1. A lot of hard work has gone into fixing older Ruby linting violations ([for example in finder-frontend](https://github.com/alphagov/finder-frontend/pull/579)). This has allowed us to [turn off the "diffing" behaviour](https://github.com/alphagov/finder-frontend/pull/581) for most projects - we now lint all of the code all of the time.
+1. We currently do not lint our Javascript. This is a gap first raised [in 2016](https://github.com/alphagov/govuk-lint/issues/51) and further discussed in an [issue talking about Standard.js](https://github.com/alphagov/govuk-lint/issues/63).
+1. The [scss_lint tool is being deprecated](https://github.com/alphagov/govuk-lint/issues/70). The authors suggest using a different library.
+1. The CSS linting hasn't been fully adopted in GOV.UK projects - only a few frontend applications have it enabled. Others [explicitly disable it](https://github.com/search?q=org%3Aalphagov+sassLint%3A+false&type=Code). The main reason for this is that `scss_lint` does not have an autocorrect feature, so it's a significant (and boring) investment to adopt the linting for older projects.
+1. While govuk-lint has been adopted by [a lot of Ruby projects in GDS and wider government](https://github.com/alphagov/govuk-lint/network/dependents), the CSS linting feature isn't always used. Projects that don't use CSS still need to pull in the `scss_lint` dependency.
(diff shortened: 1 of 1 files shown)


rubenarakelyan:
Can this be called something like `govuk_rubocop` or similar to fit in with our naming rules for gems?

rfc-100-linting.md
```diff
@@ -0,0 +1,55 @@
...
+1. Last year, [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) (part of the [Design System](https://design-system.service.gov.uk/)) was officially launched. Since this is the defacto standard for building frontend things in government, GOV.UK should be adopting the same tools it uses. This will allow us to easily push things upstream and re-use GOV.UK Frontend patterns.
+1. There's been a shift in the Ruby community regarding the use of Javascript packages. Where previously it was preferred to use tools written in Ruby like `scss_lint`, Rails now [ships with Yarn](https://guides.rubyonrails.org/5_1_release_notes.html#yarn-support) and supports [Webpack](https://guides.rubyonrails.org/5_1_release_notes.html#optional-webpack-support).
+
+## Proposal
+
+1. Deprecate the `govuk-lint` gem
+2. Create a new gem called `govuk_rubocop` that includes all rules configuration and a dependency on the `rubocop` gem
+3. Adopt NPM modules [standard](https://www.npmjs.com/package/standard) for Javascript linting and [sass-lint](https://www.npmjs.com/package/sass-lint) for CSS linting
```
tijmenb: The open question here is whether or not we need to share config for sass-lint. GOV.UK Frontend [uses a YAML file to configure it](https://github.com/alphagov/govuk-frontend/blob/master/config/.sass-lint.yml). The preference would be to have these rules defined in a single place. @alex-ju is going to investigate this.

36degrees: At one point we were definitely talking about shipping our linter config either as part of GOV.UK Frontend or as a separate package – if this would be useful it _might_ be something the team can look into. (/cc @nickcolley @hannalaakso)

nickcolley: Yeah, providing they're not too big we could ship them as part of GOV.UK Frontend main package which would be relatively easy.

Alternatively we could consider publishing a separate package, which may be OK since they dont get updated very often, and easier to contribute to since it'd be a small repository.

alex-ju: @36degrees that would be really helpful as we won't need to create a separate repo/place to keep these sort of config files.

rubenarakelyan:
This RFC will remain open until 6 March so please get your comments in before then.

rfc-100-linting.md
kevindew: Are the rules in GOV.UK frontend sass-lint consistent with the ones used in the scss_lint tool?

If not do we know whether we're going to switch to GOV.UK Frontend rules or maintain the scss_lint rules?

rfc-100-linting.md
```diff
@@ -0,0 +1,53 @@
...
+## Consequences
+
+### Impact
+
+- All applications that provide a frontend will have a development / CI dependency on Yarn.
+
+### Implementation
+
+1. Create new `govuk_rubocop` gem
+2. Update all GOV.UK repos to switch out `govuk-lint` for `govuk_rubocop`,  add `inherit_gem` to `.rubocop.yml`, and add a `package.json` with standard and sass-lint and tasks defined. We would probably be able to automate this.
```
kevindew: Not my area of expertise but I wonder if we might need to wrap sass-lint in an NPM module as a way to share configuration rules around?

This wouldn't be a problem with standard as per it's lack of config

rfc-100-linting.md
```diff
@@ -0,0 +1,53 @@
...
+3. Update [govuk-jenkinslib](https://github.com/alphagov/govuk-jenkinslib) to automatically `yarn install` and run the linting tasks
+4. Update documentation
+5. Enable Dependabot for NPM modules in repos
```
kevindew: Have we trialed this anywhere? If not I'd suggest we try turning it on soon in some of our apps that already use NPM. The nature of NPM seems to mean there are many more packages installed that most dependency management systems, so I wonder if this could potentially be overwhelming. Hopefully not though as it'd be good to have it.

rfc-100-linting.md
alex-ju: We were thinking that `govuk-frontend` can share the lint config along with its code (or in a separate repo - up to be decided - [as per comments above](https://github.com/alphagov/govuk-rfcs/pull/100#discussion_r258418745)) and we can use that for the `sass-lint` task.

rfc-100-linting.md
alex-ju: Good point about trialing dependabot with npm. govuk_publishing_components watching for `govuk-frontend` might be a good candidate.

rfc-100-linting.md
alex-ju: @kevindew not the same rules they are different and GOV.UK frontend sass-lint are more constraining; `*.scss` need to be [updated](https://github.com/alphagov/govuk_publishing_components/pull/758/commits/2959153de1604cb995a2b11d8c7c66363a91cd4d) in each project before moving to `sass-lint`. Good thing is that they both work in parallel for a while - no clashing rules.

andysellick:
Only one question - will the new linting run locally when we run tests?

kevindew:
> Only one question - will the new linting run locally when we run tests? (e.g. bundle exec rake)

That's an interesting question and may span broader than this RFC.

The short answer is likely to be no.

This is because the govuk-jenkinslib script expects the default task to be just to run the tests: https://github.com/alphagov/govuk-jenkinslib/blob/master/vars/govuk.groovy#L651-L655 and aside from rake default we've not got an agreed task that would just run tests (though arguably `test` would be fine)

We could make the move that the groovy script runs the whole build in one task a la rake default, but that would mean that it all happens in one stage. Which seems to be moving in a different direction than what Tijmen was aiming for with: https://github.com/alphagov/govuk-jenkinslib/pull/46

We have a setup on Content Publisher where `rake default` does all the linting and build: https://github.com/alphagov/content-publisher/blob/20bad2a09363b42b9cc4b4c02252b7792723f346/Rakefile#L9 and https://github.com/alphagov/content-publisher/blob/20bad2a09363b42b9cc4b4c02252b7792723f346/lib/tasks/lint.rake#L3-L9 
but then we've ended up with a less conventional Jenkinsfile https://github.com/alphagov/content-publisher/blob/20bad2a09363b42b9cc4b4c02252b7792723f346/Jenkinsfile

andysellick:
I think initially it could cause a lot of frustration as we adjust to the new linting, so having it run locally would be helpful. That said, if it's too complicated I'm sure we'll adapt quickly (and we can all set up the linting in our editors, right?)

rubenarakelyan:
I also feel the eventual goal is to split up tests and linting into separate rake tasks that are run as separate steps by Jenkins and reported separately to GitHub, but can also be run together as one task for development. That sounds like the most sensible option but it's a separate discussion to this one.

PR #102, opened 2019-03-29
RFC 102 Allow notification workflow step in publishing pipeline

steventux:
Email notifications are currently sent to relevant subscribers for a major update.

We've identified a use case where the significance of _organising_ content after it has been published **may** warrant an email notification.

[View RFC](https://github.com/alphagov/govuk-rfcs/blob/2cc32ae534d0f30f9702401a31ccd69f73bc4706/rfc-102-add-notification-workflow.md)

### DIFF
--- a/rfc-102-add-notification-workflow.md
+++ b/rfc-102-add-notification-workflow.md
@@ -1,0 +1,26 @@
+# A 'notify' workflow step in publishing pipeline
+
+## Summary
+
+Email notifications are currently sent to relevant subscribers only when content is published with a major update.
+
+We've identified a use case where the significance of _organising_ content after it has been published may warrant an email notification.
+
+## Problem
+
+We have a high profile use case with the [business readiness finder](https://www.gov.uk/find-eu-exit-guidance-business) where the facet-tagging workflow step is deemed significant enough to warrant notifying subscribers via an email alert.
+
+Departments wish to notify users that new items of content from across GOV.UK have been included in the results for this finder. This is partly a consequence of tagging content after it has been published, but also in part due to the process of writing and reorganising guidance at pace.
+
+Republishing content for the purposes of notifying users is possible but this seems pollutive given the workflow event is (re)categorisation.
+
+The current (temporary) [tagging notification mechanism uses a direct HTTP request between Rummager and Email Alert API](https://github.com/alphagov/rummager/blob/master/lib/indexer/workers/metadata_tagger_notification_worker.rb#L11-L20).
+
+The current (temporary) [metadata tagging implementation is moving to Content Tagger](https://github.com/alphagov/content-tagger/pull/884), the publishing pipeline will handle [tagging via links](https://github.com/alphagov/govuk-content-schemas/commit/071731cb08e8c9c8956a70769748c979f762cf44).
+The current email tagging notifications will not be triggered once this has moved.
+
+## Proposal
+
+- Publishing API supports a 'notify' endpoint and sends a distinct message type downstream for Email Alert Service/API consumption.
+- Content Tagger UI supports an explicit 'notify users of these changes' workflow choice, defaulting to a no-op.
+- Users can opt out of this type of notification as they may only wish to receive updates for publishing events.


barrucadu:
So, the email goes out to users who have a subscription which covers that content item.  Would it be possible for a user to opt out of these notifications but keep major update email alerts?

steventux:
Adding the ability to opt out of this type of notification would be a useful feature, that's not an option we currently support with the MVP tagging email alert we issue but it would provide a better user experience.

rfc-102-add-notification-workflow.md
```diff
@@ -0,0 +1,26 @@
...
+Republishing content for the purposes of notifying users is possible but this seems pollutive given the workflow event is (re)categorisation.
+
+The current (temporary) [tagging notification mechanism uses a direct HTTP request between Rummager and Email Alert API](https://github.com/alphagov/rummager/blob/master/lib/indexer/workers/metadata_tagger_notification_worker.rb#L11-L20).
+
+The current (temporary) [metadata tagging implementation is moving to Content Tagger](https://github.com/alphagov/content-tagger/pull/884), the publishing pipeline will handle [tagging via links](https://github.com/alphagov/govuk-content-schemas/commit/071731cb08e8c9c8956a70769748c979f762cf44).
+The current email tagging notifications will not be triggered once this has moved.
+
+## Proposal
+
+- Publishing API supports a 'notify' endpoint and sends a distinct message type downstream for Email Alert Service/API consumption.
```
cbaines: If I understand the user need, publishers want to be able to tag content, and have the subscribers to the finder be notified that the content of the finder has changed?

I guess one alternative to a separate, explicit "notify" endpoint would be to support something like this as a option when patching links.

I'm not sure that approach will work for finders though, I can see how it could work for Topic pages, as the content links specifically to the Topic page, but it doesn't sound like there's a direct link in the Publishing API sense between the content in the finder and the finder. Does that make sense?

bilbof:
Forgive me for my ignorance of content tagger. I'm not sure if versioning documents is already a part of this application.

If an editor is asked if they are making a major/minor change when changing the tags on a document, could we use the existing publishing workflow?

rfc-102-add-notification-workflow.md
```diff
@@ -0,0 +1,26 @@
...
+- Content Tagger UI supports an explicit 'notify users of these changes' workflow choice, defaulting to a no-op.
```
cbaines: It would be interested to know if you've considered having Content Tagger do what Rummager is doing now, and talking to the Email Alert API directly.

Or to put it another way, what's the benefit of involving the Publishing API and RabbitMQ.

steventux: I did consider this, it felt out of place as it was another bypass of the publishing pipeline to get to a downstream API.
The current tagging email mechanism was a very quick hack in Rummager, I think there are sufficient benefits of predictability and visibility in using the publishing pipeline.

steventux: @bilbof I've considered/discussed just using a major change as a way to send an additional notification to subscribers. It's quite pollutive of Edition change history though, each major change is recorded and in many formats it's also rendered. It feels like an awkward abuse of change history in order to generate a notification.
Ideally in the future facet tagging will happen as part of the draft-to-published workflow and this endpoint may well be deprecated and dropped. Some discussions I've had with developers identified a need to defer notifications though, for instance when dependency resolution is slow to update all related items, it may be desirable to have greater control over when a notification goes out. This RFC doesn't fully address this as we are working with the EU Exit finder tagging use case but it's an example of how this endpoint might be evolved.

rfc-102-add-notification-workflow.md
steventux: @cbaines it could well be integrated into another endpoint, we have one very awkward use case right now so keeping a clean separation seemed the pragmatic choice, if the need for this goes away I won't be sorry, by making a distinct thing it might be easier to deprecate and delete if it's ultimately superseded by better tagging workflow (ie as part of drafting and publishing) .

rfc-102-add-notification-workflow.md
cbaines: > I did consider this, it felt out of place as it was another bypass of the publishing pipeline to get to a downstream API.

I guess to me, talking to the Email Alert API directly seems a bit neater, as it involves less services, and I don't see what would be bypassed by not going through the Publishing API. I've only skimmed the relevant Publishing API Pull Request, but the only extra stateful change that occurs is the event in the events table right?

> The current tagging email mechanism was a very quick hack in Rummager, I think there are sufficient benefits of predictability and visibility in using the publishing pipeline.

As I mentioned above, the entry in the events table adds to the audit trail at least, which might be what you meant by visibility.

Anyway, adding support in to the Publishing API to talk indirectly to the Email Alert API doesn't preclude talking directly to the Email Alert API now, or in the future, so this is not a crucial point of discussion.

rfc-102-add-notification-workflow.md
cbaines: Ok, sure :+1:

rfc-102-add-notification-workflow.md
steventux: @cbaines I'll go with consensus here. I don't particularly mind talking directly to the email alert api if that's agreeable to the tech community, thanks for the helpful info and discussion.

kevindew:
Good work trying to dig our way out of this hole :+1:

One of the things that seems a bit fishy and unresolved is the communication with Email Alert API. At the moment we’re having to fib to it with a different change note and public updated at so it doesn't consider it a conflict. This feels like something that should be resolved if this functionality were to be more than temporary. Have any changes to Email Alert API been considered? 

It feels wrong that we tell Email Alert API about these as a content change notification. I feel like the way to legitimise this would be for Email Alert API to know about these in a different way and be able to work out the email and tie it to the last content change. This suggestion feels like it increases the scope though so understand it may not be do-able.

Otherwise I think Chris’ suggestion is a good one of having Content Tagger talk directly to Email Alert API. That seems to reduce the footprint of applications/concepts affected by this and that seems pragmatic while we lack a nice architectural vision for the long term.

steventux:
Thanks @kevindew for the comment. I think there's sufficient agreement both within our team and the commenters here that talking directly to the email-alert-api is viable for now.
We only have one fairly anomalous use case so perhaps a bespoke endpoint is overly complex and will take longer to unpick if this remains the only use.
I'm happy to follow @cbaines suggestion that we send alerts directly from content-tagger to email-alert-api, having chatted with @leenagupte we think it's the pragmatic thing to do as it doesn't prevent us from sending the tagging notifications we need to send.

PR #103, opened 2019-04-09
RFC 103: Merge Dependabot Pull Requests with a single review

cbaines:
# [View the RFC](https://github.com/alphagov/govuk-rfcs/blob/merge-dependabot-pull-requests-with-a-single-review/rfc-103-merge-dependabot-pull-requests-with-a-single-review.md)

Closes on Wednesday the 17th of April.

### DIFF
--- a/rfc-103-merge-dependabot-pull-requests-with-a-single-review.md
+++ b/rfc-103-merge-dependabot-pull-requests-with-a-single-review.md
@@ -1,0 +1,45 @@
+## Problem
+
+I think there are two somewhat related problems that I'd like to see
+addressed.
+
+Pull Requests by Dependabot, that the guidance describes as requiring
+a review from two people are being merged with a single review.
+
+Pull Requests by Dependabot are up for long amounts of time, without
+being merged, including those that could be time sensitive like
+security fixes.
+
+## Proposal
+
+I think a step forward in addressing both problems would be to not
+treat Dependabot as an external contributor when reviewing Pull
+Requests, and amend the guidance on reviewing and merging Pull
+Requests to permit merging Pull Requests by Dependabot with a single
+review.
+
+## Rationale
+
+Currently, Pull Requests by Dependabot are viewed as coming from an
+external contributor, and as such, should be reviewed by two people
+employed by GDS, working on GOV.UK.
+
+However, Dependabot is already a special case in the following ways:
+
+ - It's an automated service, not a person raising the Pull Requests
+ - Dependabot pushes changes directly to the repositories hosted on
+   the alphagov GitHub organisation, rather than using a fork of the
+   repository
+ - It should only be changing files in the repositories relating to
+   the versions of dependencies (Gemfile and Gemfile.lock in the case
+   of Rubygems/Bundler)
+
+Therefore, to try and decrease the amount of time Pull Requests remain
+open, and reduce the amount of time spent reviewing them, the proposal
+is to amend the guidance to only require a single review on GitHub.
+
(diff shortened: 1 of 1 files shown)


kevindew:
I'm for them just having a single review, that's effectively what happens most of the time in practice anyway.

karlbaker02:
I have mixed opinions on this. I agree that potentially needing two reviews on a PR raised by Dependabot, depending on what gems are being updated, can slow down the process; likewise, if this process is not enforced, it means that our guidance in the developer docs does not match reality and creates confusion as to what the right approach is. At the same time, I don't think that this has been the main reason historically why Dependabot PRs tend to hang around for a long time.

One of the ways that I think we can improve, whatever we decide, is to be more proactive on reviewing Dependabot PRs, testing them to ensure that there are no unexpected side-effects, then merging and deploying them all the way through to production. It _might_ actually help to have two people review these PRs, as this can give more confidence if a single reviewer doesn't have the experience the changes in said PRs as mentioned in a comment above.

There's also a more remote and potential danger with a one-person-approve-and-merge that doing this becomes routine and happens oblivious to the contents of the change within the PR, which is a risk hopefully reduced with a two person model.

sihugh:
I agree in general because of the opinions stated above.  I do wonder whether we should get our heads together and see if there's any different mitigation steps we could take to be less vulnerable to things like https://snyk.io/blog/malicious-remote-code-execution-backdoor-discovered-in-the-popular-bootstrap-sass-ruby-gem/ (which we weren't affected by, probably *because* we didn't have dependabot at the time and therefore didn't bump to that particular version)

Dependabot are [looking at ways](https://dependabot.com/blog/stopping-dependencies-getting-hijacked-2/) to mitigate this sort of attack, but it doesn't look super easy for them to do.

rubenarakelyan:
I understand all the concerns in the comments about making sure we know what we're merging, but then again, none of this is any different from a developer on GOV.UK making gem bumps which are then approved by a single other developer before being merged. As long as reviewers check that nothing else apart from gem bumps are contained in Dependabot PRs, then I think they should be treated the same way as any other PR.

karlbaker02:
I think the key difference is that a developer on GOV.UK making gem bumps can be assumed to do so knowing that they are making the gem bumps, before another developer then reviews (so it has had 2i in a sense). Dependabot doing the same and then receiving a single review, whilst it is similar, means only one developer will review in the current scheme of things. This is not to say that after seeing the comments above we shouldn't use a single review approach, but I want to add this in for completeness.

cbaines:
> One of the ways that I think we can improve, whatever we decide, is to be more proactive on reviewing Dependabot PRs, testing them to ensure that there are no unexpected side-effects, then merging and deploying them all the way through to production. It _might_ actually help to have two people review these PRs, as this can give more confidence if a single reviewer doesn't have the experience the changes in said PRs as mentioned in a comment above.

I'm definitely not against multiple people reviewing Pull Requests by Dependabot, and actually, for some applications/gems (e.g. Rails 6 for Whitehall...) you probably want multiple people to review Pull Request. While permitting merging with only a single review might lessen this, I'm still hopeful that this will be an improvement overall.

> There's also a more remote and potential danger with a one-person-approve-and-merge that doing this becomes routine and happens oblivious to the contents of the change within the PR, which is a risk hopefully reduced with a two person model.

Yep, however, I'm all for emphasising the importance of actually looking at the diff, and maybe the changelog/commits of the relevant gem.

cbaines:
I think the general consensus is towards accepting this, and Tom has approved the Pull Request, so I'm going to merge this now. Thanks everyone for getting involved :+1:

PR #104, opened 2019-04-24
RFC 104 Service API responses should provide classes

bilbof:
This RFC will close on Wed 8th May.

[VIEW THE RFC](https://github.com/alphagov/govuk-rfcs/blob/3e6279739ebe0c1f5da56a7caa57ac4cf8e633bb/rfc-104-shared-classes.md)

Example:

```
content_item = GdsApiAdapters::ContentStore::ContentItem.find('/government/organisations/hm-treasury')
content_item.kind_of? GdsApiAdapters::ContentStore::ContentItem
=> true
content_item.instance_of? GdsApiAdapters::ContentStore::OrganisationContentItem
=> true
content_item.title
=> 'HM Treasury'
content_item.is_live?
=> true
content_item.first_published_at.class
=> DateTime
content_item.links.ordered_child_organisations.first.class
=> GdsApiAdapters::ContentStore::OrganisationContentItem (lazy loaded)
```

### DIFF
--- a/rfc-104-shared-classes.md
+++ b/rfc-104-shared-classes.md
@@ -1,0 +1,85 @@
+# Service API responses should provide classes
+
+## Summary
+
+When frontend service make a request to a publishing service, they should,
+where useful, receive classes in response.
+
+For example, when finder-frontend requests a finder content item from
+the content store, a FinderContentItem might be returned, rather than JSON.
+
+```
+content_item = GdsApiAdapters::ContentStore::ContentItem.find('/government/organisations/hm-treasury')
+content_item.kind_of? GdsApiAdapters::ContentStore::ContentItem
+=> true
+content_item.instance_of? GdsApiAdapters::ContentStore::OrganisationContentItem
+=> true
+content_item.title
+=> 'HM Treasury'
+content_item.is_live?
+=> true
+content_item.first_published_at.class
+=> DateTime
+content_item.links.ordered_child_organisations.first.class
+=> GdsApiAdapters::ContentStore::OrganisationContentItem (lazy loaded)
+```
+
+This will bring a number of benefits, including application consistency,
+keeping things DRY, easier communication between services, and making API responses immutable.
+
+## Problem
+
+Applications repeat code to handle GdsApiAdapters responses. For example, the ContentItem class appears in a number of different apps. I found some similar implementations:
+
+- [collections](https://github.com/alphagov/collections/blob/master/app/models/content_item.rb),
+- [finder-frontend](https://github.com/alphagov/finder-frontend/blob/master/app/models/content_item.rb)
+- [email-alert-api](https://github.com/alphagov/email-alert-api/blob/master/app/models/content_item.rb)
+- [content-tagger](https://github.com/alphagov/content-tagger/blob/master/app/models/content_item.rb)
+- [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/app_helpers/taxon_breadcrumbs.rb#L47)
+- [support-api](https://github.com/alphagov/support-api/blob/master/app/models/content_item.rb)
+- [content-store](https://github.com/alphagov/content-store/blob/master/app/models/content_item.rb)
+- [govuk-content-navigation](https://github.com/alphagov/govuk-content-navigation/blob/master/lib/ruby/content_item.rb)
+- [govuk_navigation_helpers](https://github.com/alphagov/govuk_navigation_helpers/blob/master/lib/govuk_navigation_helpers/content_item.rb)
+- [govuk-services-prototype](https://github.com/alphagov/govuk-services-prototype/blob/master/app/models/content_item.rb)
+
+Benefits:
+
+- Apps would be more consistent - easier to work with.
+- Keeps apps DRY and more focused on a specific purpose.
+- Would make API responses effectively immutable (dissuades manipulation of hash responses).
+- Would establish a more explicit contract between services.
+- Classes that mirror the schemas, like FinderContentItem or PeopleContentItem,
+  would make it easier to implement specialist behaviour across services.
+- We could add ActiveRecord-like behaviour to classes returned by GDS API,
+  like `ContentItem.find(:content_id)`.
+- Repeated logic, like cacheing and statsd timers. E.g. finder-frontend caches content items,
+  and monitors response times for content items but we want this implemented in
+  more places (needed to do this in each app).
+- Easier monitoring of performance of requests (as above).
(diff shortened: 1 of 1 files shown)


elliotcm:
Any thoughts on whether we're looking to break this repo up?

thomasleese:
@elliotcm There was a firebreak idea to do it once but I don't think it got anywhere: https://trello.com/c/Ik8s0rvh/115-reorganise-gds-api-adapters-to-make-updating-easier

rubenarakelyan:
@elliotcm I tried it a few firebreaks ago but got stuck with running out of time as well as stumbling across a few things I didn't know about in there, like automatic pagination for whitehall API responses.

For the RFC itself: is there a way we can build the specific classes and their available methods based on the content schemas? That would make it more maintainable.

bilbof:
@rubenarakelyan we could potentially autogenerate classes. 

I think @koetsier had an idea on a way to do this using https://github.com/trailblazer/representable?

I suppose if we set more specific types on fields in schemas, like using `datetime` instead of `string`, we could do things like returning a `DateTime`.

We could make it always possible to access the underlying data from the content item (as [here](https://github.com/alphagov/govuk-services-prototype/blob/master/app/models/content_item.rb#L13-L17)), and then override those keys with custom methods where appropriate (e.g. `public_updated_at` should return a `DateTime`).

```
  # dynamically create methods using the schema keys
  schema_keys.each do |key|
    define_method key do
      @content_item_data[key.to_s]
    end
  end
```

To ensure the schemas and classes stay consistent, we could use the schemas to test the classes. E.g. examples from content schemas could be used to test the API classes, and we'd run this check whenever either repo was changed. We could also add tests that force a developer to change both the content schemas and the class.

If anyone has ideas on automating the creation/updating of API classes I'd be interested to hear them.

kevindew:
We already have some examples of particular api's wrapping API responses in classes. This can be seen in Link Checker API: https://github.com/alphagov/gds-api-adapters/blob/166793c645628657726af1ae87db345c8c0a7e83/lib/gds_api/link_checker_api.rb#L153-L173 so it's not actually a given expectation that everything returns a `GdsApi::Response`.

I notice this RFC only focuses on Content Items in examples and most of the discussion, but the scope seems a bit broader than that? Could you clarify if it would be to change everything or just Content Store responses (for what it's worth I think it'd be fine to be just Content Store).

The other thing I was intrigued about with regards to scope is the suggestion that each type of ContentItem seems to have it's own class (organisation, finder etc) and so a call to get content items would return a subclass of `GdsApiAdapters::ContentStore::ContentItem`. Would individual classes be required or would a generic `ContentItem` handle most things?

I'm a bit concerned about the lazy load suggestion, as one of the frontend architectural goals  is that we can build a page with a single request to the Content Store (sorry I can't find the source but @timblair has a doc I believe). Therefore we shouldn't be making any additional requests to content store opaque.

Another thing to bear in mind is what a migration path would look like for apps that use this. I remember when we switched GdsApiAdapters from using open struct to hashes, which proved to be a rather painful transition as: a) it was a GdsApiAdapters major version change that suddenly mattered and b) it spread far and wide across all apps.

Also it might be quite challenging working out the degree of consistency with content schemas because of the unusual way content schemas is deployed as a dependency. That might lead to it being more prudent to determine some things at run time?

timblair:
The relevant principle that @kevindew referred to is "[Front-end applications render main content bodies with a single API call.](https://docs.google.com/document/d/1Oft4akc6dZfhhOjosNPbFpcLUOUjz7YG7QPcVZi8hww/edit#heading=h.h8e5mqu62b4b)" While lazy-loading can be a good thing to prevent loading of additional data when not required, it can also result in performance problems over time if people rely on it and don't consider the data access that's happening under the hood. The Content Store is not an attached database with a persistent connection, and API calls are expensive.

My main concern about this approach is about the delivery of it. In your cons you say "It's a lot of setup work." Right now, our approach is at least consistent, even if there could be improvements. Is it realistic that we would update every consuming application to use this new approach, or are we likely to end up with a hybrid approach where some applications never quite get to the top of the to-do list? Is it worth the investment?

bilbof:
Thank you for your comments!

@kevindew 

I used ContentItem as the example, and actually I'd be happy for the scope to be limited to this type (perhaps with subclasses like FinderContentItem). If the approach is helpful we could expand it to other API responses?

> That might lead to it being more prudent to determine some things at run time?

Yes potentially! We could create classes for each content item using the schemas at runtime. We could have methods like `is_live?` and return types like `DateTime` for some fields. It would also be nice to have behaviour which is custom to a content item type (but this could come later).

If we auto generate a class using the schemas for each response this might be good enough, and would lead to some of the benefits I listed:

- consistency between apps
- keeping apps DRY
- immutable api responses
- explicit contracts between services.

The other things (repeated custom behaviour like cacheing and statsd timing) could be added later.

@timblair you're right, this does risk consistency. Migrating apps to requesting a class rather than a JSON response in stages could lead to less consistency if we don't complete the migration quickly.

If we try it out with just base content items, this would be fairly quick to do, and would let us prove/disprove the benefits more easily. If it does work, then we'd have to spend time migrating the other apps to use content items, which would require an investment of time (but I'm not sure how much).

kevindew:
Perhaps an idea for this would be to create this separate from GDS API Adapters as a means to start the splitting up of this tool. 

That way we could then deprecate the content store aspect of GDS API Adapter and have a govuk_api_content_store gem that performs the ideas suggested?

bilbof:
@kevindew that's a good idea. It might be a decent firebreak project to prototype a gem that uses GdsApiAdapters to make requests and returns classes generated from the schemas.

As the 2 weeks are up for this RFC I will close it now. Please feel free to add more thoughts on this. Thank you all for your comments!

tijmenb:
Bit late to the party, but here's my 2 cents! My main issue is that we'd likely duplicate the knowledge about the API response.

Current workflow for adding a field in a publisher for display on the frontend:

1. Add field to content schemas
2. Send field to publishing-api from the publisher
3. Use the field in the frontend app

If we'd have these classes exposed from gds-api-adapters or a separate gem, we would require a rebuild, version bump and release before being able to update the frontend.

That said, it's totally reasonable to make classes for API responses. But I think it's actually more useful to create these inside the application, where they can be tailor made to the context we're in.

For example, content tagger has a [Taxon model](https://github.com/alphagov/content-tagger/blob/master/app/models/taxon.rb) that is populated with a content item, but exposes specific functionality. Collections [also has a Taxon model](https://github.com/alphagov/collections/blob/master/app/models/taxon.rb), but with different methods. This makes it possible to see which fields from a taxon content item are used in an application, and to easily add context-specific behaviour. It makes development easier, because you don't have to understand the behaviour of yet another library - everything you need is defined in the application.

PR #106, opened 2019-06-17
RFC 106: Use Docker for local development 🐳

tijmenb:
In the next firebreak, we're thinking to overhaul GOV.UK's development environment. Before doing this, we'll have to choose an approach.

This RFC proposes to use Docker. But we know Docker isn't without problems, so this RFC also serves to collect feedback on the types of problems we're likely to see.

👉[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/docker-for-local-development/rfc-106-docker-for-local-development.md)

### DIFF
--- a/rfc-106-docker-for-local-development.md
+++ b/rfc-106-docker-for-local-development.md
@@ -1,0 +1,50 @@
+# RFC 106: Use Docker for local development
+
+## Summary
+
+Adopt a Docker-based approach for local development instead of the Vagrant VM.
+
+## Problem
+
+We currently rely on the Vagrant development VM for the day to day dev cycle.
+
+- It has [many documented issues](https://docs.publishing.service.gov.uk/manual.html#development-vm)
+- Changes to puppet often result in a broken VM, because the infrastructure team don't use it.
+- We are adopting more cloud-based services into our stack, like S3 and Amazon's Elasticsearch. This means that our development environment doesn't look like production anymore, but we still need code in govuk-puppet.
+- It's hard to keep updated. Running the puppet command will often error and take a long time.
+- GOV.UK is moving to a containerised infrastructure (likely the [GDS Supported Platform](https://github.com/alphagov/gsp) (GSP))
+
+## Possible approaches to local development
+
+- An approach based on **Docker and docker-compose**. Ben Thorner has a working prototype called [govuk-docker](https://github.com/benthorner/govuk-docker) and we have the [end-to-end tests](https://github.com/alphagov/publishing-e2e-tests) running in the same. Other GDS programmes like GOV.UK Pay use this approach.
+- An approach based on **GNU Guix** package manager, which Chris Baines uses via govuk-guix.
+- Because they are read-only, frontend applications on GOV.UK can be **run against production** using the startup.sh --live flag.
+- Developers on the GOV.UK PaaS work by developing against a **remote set of services** - we’d run many versions of GOV.UK in the cloud, which developers can use in development.
+- Applications can be run locally by **installing dependencies manually**. This is what some of GOV.UK’s Linux users do.
+- The RE team has been working on local development tooling called **[gsp-local](https://github.com/alphagov/gsp/blob/master/docs/gds-supported-platform/getting-started-gsp-local.md)**.
+
+Approach | Pro | Cons
+-- | -- | --
+Vagrant VM | Can look like production | Resource intensive, brittle
+Docker | Aligns with future hosting. Highly reproducible. | Slow because on a Mac it needs a VM. Development won’t match production/CI exactly until we migrate onto a container-based platform.
+Guix | Huge amount of work put in | Technology unknown in GOV.UK
+Against production | Low setup, fast | Only works for frontend applications. Still needs local install. Limits you to testing one application at a time.
+Remote dependencies | Low setup. Can match Production exactly in terms of technologies used | Expensive, no offline development
+Manual install | Fast in development | Lots of setup, brittle
+gsp-local | GOV.UK will move to the GSP at some point | Multiple new layers of technology, quite new
+
+## Proposal
+
+We'll adopt a Docker based local development environment. Docker is widely supported in the community and is most aligned with our tech strategy. In the long term we might replace it with a local environment of the GDS Supported Platform.
+
+## Impact and follow up work
+
+We'll commit to doing these things:
+
+- Document how to test puppet locally using Vagrant now that the development VM is going away.
+- Come up with a process of ownership so that the tooling improves over time.
+
+Things we'll need to look out for:
+
+- Make sure we can limit the number of dependencies we run, to avoid having the environment being too resource intensive.
+- Our local development will diverge more from our production environment. This might lead to more testing on integration. We should be sure this is viable and doesn't become a blocker if lots of people are testing things at the same time.
--- a/rfc-106-docker-for-local-development.md
+++ b/rfc-106-docker-for-local-development.md
@@ -26,8 +26,8 @@ We currently rely on the Vagrant development VM for the day to day dev cycle.
 Approach | Pro | Cons
 -- | -- | --
 Vagrant VM | Can look like production | Resource intensive, brittle
-Docker | Aligns with future hosting. Highly reproducible. | Slow because on a Mac it needs a VM. Development won’t match production/CI exactly until we migrate onto a container-based platform.
-Guix | Huge amount of work put in | Technology unknown in GOV.UK
+Docker | Aligns with future hosting. Reproducible (with caveats). | Slow because on a Mac it needs a VM. Development won’t match production/CI exactly until we migrate onto a container-based platform.
+Guix | Is almost feature complete (runs Signon, TLS) | Technology unknown in GOV.UK
 Against production | Low setup, fast | Only works for frontend applications. Still needs local install. Limits you to testing one application at a time.
 Remote dependencies | Low setup. Can match Production exactly in terms of technologies used | Expensive, no offline development
 Manual install | Fast in development | Lots of setup, brittle


rfc-106-docker-for-local-development.md
```diff
@@ -0,0 +1,37 @@
...
+- Developers on the GOV.UK PaaS work by developing against a **remote set of services** - we’d run many versions of GOV.UK in the cloud, which developers can use in development.
+- Applications can be run locally by **installing dependencies manually**. This is what some of GOV.UK’s Linux users do.
+
+
+Approach | Pro | Cons
+-- | -- | --
+Vagrant VM | Can look like production | Resource intensive, brittle
+Docker | Aligns with future hosting. Highly reproducible. | Slow because on a Mac it needs a VM. Development won’t match production exactly until we migrate Production onto a container-based platform.
+Guix | Huge amount of work put in | Technology unknown in GOV.UK
+Against production | Low setup, fast | Only works for frontend applications. Still needs local install
```
ChrisBAshton: I'd add that it also limits you to testing one application at a time. Often a feature will span multiple repositories and it's useful to be able to test it all working together, but the 'against production' approach only lets you test one part of the system at a time and it can be difficult to co-ordinate.

rfc-106-docker-for-local-development.md
```diff
@@ -0,0 +1,37 @@
...
+Remote dependencies | Low setup. Can match Production exactly in terms of technologies used | Expensive, no offline development
+Manual install | Fast in development | Lots of setup
```
thomasleese: I think we shouldn't be too quick to dismiss this option, particular for our more simple backend applications. This is generally the method I use to work and it's often the quickest and easier way for me to run tests.

```
$ brew install postgres redis
$ brew services start postgres
$ brew services start redis
$ bundle exec rake
```

This method certainly isn't appropriate for all our applications (particular ones with large numbers of dependencies) or for everyone. It can be quite daunting and sometimes causes problems with it not being well documented, but for a lot of our apps, I think it's often the simplest option.

barrucadu:
I try not to use the dev VM for running apps any more because it's a pain, so the main thing I use the dev VM for is for testing puppet changes.  It can be awkward to test those elsewhere:

- You can't test puppet locally outside the VM
- If you mess up a puppet change the easiest fix is often to rebuild the VM, which makes it a bit of a pain to test in integration (need to log in to the AWS console and terminate instances).

How will developing puppet work in the new world of docker?

rfc-106-docker-for-local-development.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Proposal
+
+Adopt a Docker based local development environment.
+
+## Impact
+
+- How does local puppet development work?
```
barrucadu: I try not to use the dev VM for running apps any more because it's a pain, so the main thing I use the dev VM for is for testing puppet changes.  It can be awkward to test those elsewhere:

- You can't test puppet locally outside the VM
- If you mess up a puppet change the easiest fix is often to rebuild the VM, which makes it a bit of a pain to test in integration (need to log in to the AWS console and terminate instances).

How will developing puppet work in the new world of docker?

tijmenb: This is a really good question. What are some examples of puppet things you've tested in the VM?

rfc-106-docker-for-local-development.md
```diff
@@ -0,0 +1,41 @@
...
+
+Adopt a Docker-based approach for local development instead of the Vagrant VM.
+
+## Problem
+
+We currently rely on the Vagrant development VM for the day to day dev cycle.
+
+- It has [many documented issues](https://docs.publishing.service.gov.uk/manual.html#development-vm)
+- Changes to puppet often result in a broken VM, because the infrastructure team don't use it.
+- We are adopting more cloud-based services into our stack, like S3 and Amazon's Elasticsearch. This means that our development environment doesn't look like production anymore, but we still need code in govuk-puppet.
```
barrucadu: We also have that problem in CI.  Will we be dockerising CI too?

rfc-106-docker-for-local-development.md
theseanything: Another issue for manual install is we'd have to consider supporting multiple platforms - developers are starting use other platforms such as Ubuntu, Guix etc... but agree for running tests its been the easiest.

rfc-106-docker-for-local-development.md
barrucadu: Here are a few:

- Getting both ES5 and ES6 running in CI so we can test search-api against both at the same time: https://github.com/alphagov/govuk-puppet/pull/9216

- Removing cronjobs: https://github.com/alphagov/govuk-puppet/pull/9015

- Tearing down rummager (particularly hard to test in integration!): https://github.com/alphagov/govuk-puppet/pull/9000

I think these will all be minor issues when production and CI are containerised and managed by something more reproducible than puppet: adding another container is straightforward; tearing stuff down should be no more complicated than just removing configuration; etc.  But until we're at that point, we need to do tedious things like manually tearing down state and managing cronjobs.

However, having said that: most of my puppet changes aren't really suitable for testing in the dev VM, stuff like modifying hostnames or environment-specific hieradata.  And I've made a fair few puppet changes which were *only* needed to keep the dev VM working.  So maybe this isn't a show-stopper, but for the changes which are easy to test locally and hard to test remotely the dev VM is nice.

theseanything: Will we also have two sets of configuration to manage? (i.e. Puppet vs docker world)

rfc-106-docker-for-local-development.md
barrucadu: I'm not sure whether we'd *need* to support multiple platforms.  There would be the blessed "this is how you run GOV.UK locally on your GDS-issued Mac" approach, with some documented changes for Linux users, but anything more exotic (Guix, Nix, etc) would be up to the individual to figure out.

rfc-106-docker-for-local-development.md
```diff
@@ -0,0 +1,41 @@
...
+- An approach based on **GNU Guix** package manager, which Chris Baines uses via govuk-guix.
+- Because they are read-only, frontend applications on GOV.UK can be **run against production** using the startup.sh --live flag.
+- Developers on the GOV.UK PaaS work by developing against a **remote set of services** - we’d run many versions of GOV.UK in the cloud, which developers can use in development.
+- Applications can be run locally by **installing dependencies manually**. This is what some of GOV.UK’s Linux users do.
+
+
+Approach | Pro | Cons
+-- | -- | --
+Vagrant VM | Can look like production | Resource intensive, brittle
+Docker | Aligns with future hosting. Highly reproducible. | Slow because on a Mac it needs a VM. Development won’t match production exactly until we migrate Production onto a container-based platform.
```
theseanything: How do we deal with the databases? Would we dockerize the DB engines? Would we have separate DB containers for each app?

rfc-106-docker-for-local-development.md
```diff
@@ -0,0 +1,41 @@
...
+
+- It has [many documented issues](https://docs.publishing.service.gov.uk/manual.html#development-vm)
+- Changes to puppet often result in a broken VM, because the infrastructure team don't use it.
+- We are adopting more cloud-based services into our stack, like S3 and Amazon's Elasticsearch. This means that our development environment doesn't look like production anymore, but we still need code in govuk-puppet.
+- It's hard to keep updated. Running the puppet command will often error and take a long time.
+- GOV.UK is moving to a containerised infrastructure
+
+## Possible approaches to local development
+
+- An approach based on **Docker and docker-compose**. Ben Thorner has a working prototype called [govuk-docker](https://github.com/benthorner/govuk-docker) and we have the [end-to-end tests](https://github.com/alphagov/publishing-e2e-tests) running in the same. Other GDS programmes like GOV.UK Pay use this approach.
```
theseanything: If we are eventually using k8s in production... does using docker-compose introduce another set of config to manage?

rfc-106-docker-for-local-development.md
benthorner: It would be possible, but more resource-intensive to have separate DBs for each app. The normal approach is to re-use dependencies, much like you would natively/locally - something like this: https://github.com/benthorner/govuk-docker/blob/master/docker-compose.yml#L12

rfc-106-docker-for-local-development.md
benthorner: I've been using the local approach too, for a long time. The main drawback was getting apps to talk to each other locally - often the DB is the easiest thing to get working. Eventually I got to the point where I just wanted one dev environment that could do everything well.

rfc-106-docker-for-local-development.md
```diff
@@ -0,0 +1,37 @@
...
+- An approach based on **GNU Guix** package manager, which Chris Baines uses via govuk-guix.
+- Because they are read-only, frontend applications on GOV.UK can be **run against production** using the startup.sh --live flag.
+- Developers on the GOV.UK PaaS work by developing against a **remote set of services** - we’d run many versions of GOV.UK in the cloud, which developers can use in development.
+- Applications can be run locally by **installing dependencies manually**. This is what some of GOV.UK’s Linux users do.
+
+
+Approach | Pro | Cons
+-- | -- | --
+Vagrant VM | Can look like production | Resource intensive, brittle
+Docker | Aligns with future hosting. Highly reproducible. | Slow because on a Mac it needs a VM. Development won’t match production exactly until we migrate Production onto a container-based platform.
```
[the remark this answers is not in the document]
benthorner: This has been a pain for me to. We're pretty greedy with our dependencies, but often we can afford to be more conservative. I've tried to capture this in a bunch of [needs and decisions](https://github.com/benthorner/govuk-docker/blob/master/NEEDS.md#running-services) when I started rolling my own.

bevanloon:
👍 for starting to work on this - the dev vm is almost the first impression developers get of GOV.UK's technology and the current state means that first impression is often not a good one.

Do we have any thoughts on how we plan to approach supporting the eventual solution?  Some suggestions off the top of my head:
* Make a specific team be responsible (naming no names 🙈but we should consider whether we have capacity from a delivery perspective).
* Have some rotational responsibility - teams own the "dev experience" for a period of time and a portion of delivery time can be planned for improving/troubleshooting.
* Free for all - which seems to be how we currently operate with the dev vm. Sometimes if nobody is specifically responsible, nobody takes responsibility. Perhaps it works well enough though?

The other thought I have is that we should be mindful that while we will undoubtedly get a lot done in firebreak, we might not have a completed working solution so it would be good to plan for some time in the beginning of next quarter to carry on with the work so we don't loose momentum. Coupled with a consideration that if we do have something good enough for the majority of GOV.UK to use, we might bump into bugs as we all start using it in anger so we should plan for troubleshooting time as we would with any new release.

rfc-106-docker-for-local-development.md
bevanloon: I think there's a lot to be said for a manual install. I've seen a few people struggle with the dev vm because they mistyped a command and ran in on their host rather than their vm, or were just confused about when they should be using the vm and when not..... A manual install removes this class of problem. The advantage of a VM though is (ideally) it's cheap and easy to throw away if you want to start fresh. That can be harder to do with a manual local install. Probably more so with things like postgres/redis/elasticsearch/rabbit than with the actual code and ruby versions to be fair.

bevanloon: I also strongly agree with @barrucadu - we should support the default case of people developing on a GDS-issued Mac. That should be the golden path. The GOV.UK way if you like. If you want to veer from it then that's fine, but it's unofficial. Part of the expectation then, is you accept you'll need to be somewhat self sufficient and not expect the community to support your setup.

tijmenb: I like developing locally too. It's super nice to work on frontend apps because they can run locally, get data from production and don't need any backing services.

One thing I was thinking about is whether we could run the app you're working on locally, but run the dependencies in Docker. Theoretically you'd only need rbenv on your machine, which limits the number of things that can go wrong. Do we think this could be a secondary supported way of running?

barrucadu: That's what I've been doing for running search-api tests outside the VM: dockerised elasticsearch + rbenv.

theseanything: Yup agree - we wouldn't need to support _everything_, but as we're starting to see GDS issued Linux machines, we _might_ need to consider more than just MacOS as an "official" platform.

rfc-106-docker-for-local-development.md
tijmenb: @barrucadu thanks! Apart from the development VM, we used to be able to use `vagrant` to create specific VMs of a machine:

```sh
~/govuk/govuk-puppet $ vagrant up backend-1.backend
Bringing machine 'backend-1.backend' up with 'virtualbox' provider...
==> backend-1.backend: Box 'govuk_dev_trusty64_20160323' could not be found. Attempting to find and install...
    backend-1.backend: Box Provider: virtualbox
    backend-1.backend: Box Version: >= 0
==> backend-1.backend: Box file was not detected as metadata. Adding it directly...
==> backend-1.backend: Adding box 'govuk_dev_trusty64_20160323' (v0) for provider: virtualbox
```

Would that allow you to test those changes?

tijmenb: @theseanything Yes, presumably. I'll add something to the impact section.

rfc-106-docker-for-local-development.md
tijmenb: Very good point. I guess the answer is yes!

rfc-106-docker-for-local-development.md
barrucadu: @tijmenb That would be great

brucebolt:
One of the bugbears with the current VM setup is that the GDS VPN modifies our Mac's routes table to send requests to `*.dev.gov.uk` through the VPN, where the request eventually times out.  For example, I'm at home at the moment and cannot access any applications in the dev VM from the host machine without disconnecting from the VPN and removing the route introduced by AnyConnect.

I'd assume a dockerised solution would see multiple containers, maybe one per app.  This could impede communication between apps when the VPN is in use if we'll still be using `*.dev.gov.uk`.

Maybe we should look into alternative ways of: a) accessing applications within the containers, and b) applications communicating with each other (assuming the applications will be spread over multiple containers)?

benthorner:
@brucebolt that's interesting - I've heard about this from others, but haven't experienced it myself (using Docker). How exactly does the VPN adapter modify the routes to screw up dev.gov.uk?

brucebolt:
@benthorner From my conversations with IT, any request to anywhere other than `localhost` or `127.0.0.1` gets sent through the VPN.  I'm not sure exactly how it works, but we can "undo" the changes by running the following after disconnecting from the VPN:

```
mac$ sudo ifconfig vboxnet0 delete 10.1.1.1
mac$ sudo ipconfig set vboxnet0 manual 10.1.1.1
```

rfc-106-docker-for-local-development.md
erino: Containerising will happen as part of moving to Concourse CI

barrucadu: It will, but until that happens we'll still have this problem of needing the puppet code to manage services in a different way to production.

rfc-106-docker-for-local-development.md
elliotcm: Not sure if people are aware or if it's still true but the way we were told to work on puppet in the past was to write tests.  Our puppet repo has rspec in it and some prior art for testing it.

You can also spin up VMs for different classes of machines and then run puppet on them, which is how the old sysadmins used to do it.

How that translates to docker I don't know.

rfc-106-docker-for-local-development.md
erino: GSP have been looking at how to run [k8s locally](https://github.com/alphagov/gsp/blob/master/docs/gds-supported-platform/getting-started-gsp-local.md), and there's also a tool to [migrate from docker compose to k8s](https://github.com/kubernetes/kompose)

rfc-106-docker-for-local-development.md
rubenarakelyan: Just to add that the GDS-issues Linux laptops are not "official" in the same sense that the Macs are - the Macs are still the default machine, and there is work ongoing to work out how various GDS internal services and software runs on these machines. So we should try but not kill ourselves attempting to support every flavour of Linux that may run on these machines.

tijmenb:
🙇 Thanks everyone for the comments! I'll leave this open until Monday for some more thoughts. Next week I'll update the RFC with comments and next steps.

rfc-106-docker-for-local-development.md
pauldougan: Good point, however the firebreak work that Will P did some time back with licensify provided a useful benchmark for docker performance on mac compared to linux. From what I remember compiling and running end to end tests took about 40 minutes on mac compared to about 4 . minutes on ubuntu linux, its was about an order of magnitude faster. @stephenharker has some context on this sine he uses linux.

rfc-106-docker-for-local-development.md
pauldougan: On the subject of concourse I'd like to point you to the work that has been going on with the [GDS Supported Platform](https://github.com/alphagov/gsp). 

The platform provides CICD, Docker Registry and Notary tools in the cluster. 
https://github.com/alphagov/gsp/blob/master/docs/architecture/gsp-architecture-overview.md

Its also worth making you aware of the [gsp-local](https://github.com/alphagov/gsp/blob/0b30ca7d05637f4660b67b9df8a0c3581caa37ec/docs/gds-supported-platform/getting-started-gsp-local.md) work which allows you to provision a local instance of the gsp on docker complete with CICD/Registry.  This has recently undergone a round of user research and we are in the process of tweaking the documentation that describes how to use it.

rfc-106-docker-for-local-development.md
pauldougan: The GSP uses [Helm Charts](https://helm.sh/docs/developing_charts/) this is a collection of yaml that describes the application so that it can be deployed to a kubernetes cluster. 

As @erino mentioned kompose may be useful

rfc-106-docker-for-local-development.md
benthorner: I think we should be careful about using Licensify as an example, since it's such an outlier on GOV.UK anyway. While it's true that Docker is slower on Mac than on Linux, where it's as fast as running stuff natively, the impact depends very much on [how we choose to configure it](https://github.com/benthorner/govuk-docker/blob/master/DECISIONS.md#docker-bindmount), which includes utilising recent performance improvements from the Docker developers.

As someone who now uses Docker for everything except editing, it might be reassuring to know it performs well enough to be a contender for the Dev VM. But it is true that running stuff natively on Mac will be faster.

rfc-106-docker-for-local-development.md
benthorner: I've been wondering about this myself. One thing I found valuable in my old job was a learning piece we did on 'how to write Puppet(/Ansible/Chef/Salt...) and provision a machine from scratch' - I have a feeling I'm not alone in my limited knowledge of Puppet. Somewhat outside the scope of this PR, I know, but I wonder if it would help support/clarify our testing approach we want to have for it?

rfc-106-docker-for-local-development.md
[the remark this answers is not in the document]
pauldougan: I'd like to suggest `gsp-local ` as an additional row for your table of approaches for completeness

[gsp-local](https://github.com/alphagov/gsp/blob/master/docs/gds-supported-platform/getting-started-gsp-local.md)  is also docker based and includes a local kubernetes node, Concourse CICD tooling, Docker Registy and Notary, Istio for networking, Prometheus for observability.  

It aligns with the [RE Strategy](https://reliability-engineering.cloudapps.digital/documentation/strategy-and-principles/re-strategy.html#2-build-a-new-common-hosting-platform-for-gds-services) however its still quite new.  We have several teams using it already and has recently undergone a round of user research.

benthorner:
I agree with @elliotcm: we should definitely have someone who has a long term vision for our dev experience, whatever it may be. However, I think the main reason the Dev VM is prone to failure could be because most of the changes to it happen due to the needs of 'prod', which are different to the needs of 'dev'. This could be the main source of the 'forced to do it' problem that you mentioned.

Both the Dev VM and a native solution have a similar problem: it's expensive to throw away either of them, and otherwise difficult to work out how to update them, since they're both very stateful. In contrast, an approach like Docker or Guix is built on the assumption that almost nothing is persisted, which is kind of how we do breaking updates on 'prod' machines in AWS: destroy and rebuild.

One of the big things we lose with a native solution will be the 'dev.gov.uk' capability. I've been using a native solution myself for developing against Publishing API and Asset Manager, and the manual port numbers quickly become fiddly. Of course we could solve this with something like Nginx, but how far do we go before we end up having a 'Puppet for Mac' script/doc in disguise?

I'm also wary about the perceived simplicity of recommending native Mac, even for simple things like running tests. Not so long ago, I remember just installing the 'pg' gem was a trial, as [this article illustrates](https://stackoverflow.com/questions/6040583/cant-find-the-libpq-fe-h-header-when-trying-to-install-pg-gem). The more native Mac setup we recommend, the more I reckon we'll encounter very specific issues for individual machines, much like individual Dev VMs.

That may have turned into a 'personal dev experience rant' 😆 - there's still some useful info in there!

rfc-106-docker-for-local-development.md
```diff
@@ -0,0 +1,37 @@
...
+Guix | Huge amount of work put in | Technology unknown in GOV.UK
+Against production | Low setup, fast | Only works for frontend applications. Still needs local install
+Remote dependencies | Low setup. Can match Production exactly in terms of technologies used | Expensive, no offline development
+Manual install | Fast in development | Lots of setup
+
+## Proposal
+
+Adopt a Docker based local development environment.
```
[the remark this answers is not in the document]
stephenharker: This will become a nice incentive to update the integration/staging/production to match in due course.

benthorner:
Ah, missed something. On the negative side for a Guix/Docker kind of approach, we should bear in mind that choosing any kind of framework could mean that, for every dev task we have, we may need to change the way we think about it - "how do I do X in e.g. Docker?". There's definitely a compromise here, and perhaps it's just a matter of doing the right learning/training to overcome it.

36degrees:
I rarely do anything nowadays with the GOV.UK stack myself nowadays, but worth also considering how a native install impacts rolling out major macOS updates. In my experience, they tend to break a lot of stuff and you end up having to invest days reinstalling and reconfiguring the environment.

rfc-106-docker-for-local-development.md
pauldougan: @benthorner I mentioned licensify more of an example of something where there was some 'prior art' that might be work taking a look at.

rfc-106-docker-for-local-development.md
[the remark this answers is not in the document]
tijmenb: Thanks @pauldougan! I've added gsp-local to the list, and added an intention to move towards that in the future. I think it'd be a good approach to regard docker/docker-compose as the stepping stone towards GSP, since going from the current setup to Kubernetes would introduce many new technologies to GOV.UK at the same time.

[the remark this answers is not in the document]
tijmenb: @thomasleese @benthorner I've added a note about dependencies to the RFC.

tijmenb:
Ok, thanks everyone! This discussion has been super valuable. I've updated the RFC with some notes on GSP, things to do and things to look out for.

In summary, I think we have a some consensus that moving to Docker would be a good thing - provided we tackle the issues raised.

As mentioned, we're going to have a firebreak focused on getting this off the ground. I'll share the plan for that soon, and I'll make sure to integrate the comments from this RFC into the plan.

cbaines:
I'm a bit late in commenting, but here goes...

> This RFC proposes to use Docker. But we know Docker isn't without problems, so this RFC also serves to collect feedback on the types of problems we're likely to see.

To me, the things in the forefront of my mind are the problems we've already encountered when using Docker. We've been using it for a while to provide an environment within which to run Publishing E2E tests, but we haven't got to the point yet where this doesn't sometimes break. I think the most impactful example of this was the incident recently, where it took around a week to get back to a state where the tests could pass.

Relating this back to the problems in the RFC, currently:

> Changes to puppet often result in a broken VM, because the infrastructure team don't use it.

The reality we've seen with the Publishing E2E tests is that changes outside of our control result in a unworkable environment in which to run the tests. To me, this is worse than the problem we've got with the Development VM, as you've got more opportunities to avoid this issue if you're in control of the breaking change. If we want a development environment, using Docker that doesn't encounter the same problems as the Publishing E2E tests environment, then I think something needs doing differently.

> Approach | Pro | Cons
> -- | -- | --
> Docker | Aligns with future hosting. Highly reproducible. | Slow because on a Mac it needs a VM. Development won’t match production/CI exactly until we migrate onto a container-based platform.

I think my comments above also relate to this point on Docker being "Highly reproducible". I'm going to assume that "reproducible" here relates to this [1] concept of software reproducibly, and that the Docker images are the binaries.

1: https://en.wikipedia.org/wiki/Reproducible_builds

I think the problems we've encountered with the Publishing E2E tests could be seen as a build reproducibility problem. No change was made to the Dockerfile which describes the image within which to run the Publishing E2E tests, but, the image still changed, and in such a way that prevented the tests from running.

I guess another way some people talk about reproducibility with Docker is that because you can copy one image from one machine to another, then the behaviour on both machines should be the same (reproducible).

To me, this doesn't put my mind at ease. Unless you're very rigorous about rarely building images, and always using some fixed images, you're going to end up with the same Dockerfile producing different images. Also, when you want to make a specific change to a Dockerfile, there's going to be little you can do to control changes that occur in subsequent steps.

> Guix | Huge amount of work put in | Technology unknown in GOV.UK

To varying extents, it's possible to run the Publishing E2E tests with the environment created through govuk-guix, and also use govuk-guix to manage local system containers which can be used for local development.

I don't think a "Huge amount of work put in" is a advantage of using Guix for local development, but I do believe there are actual advantages that would be of benefit to GOV.UK, and the developers working on it.

There are some things that govuk-guix supports that haven't really come up, things like:
 - Managing Signon configuration
 - Mixing in local code for some apps with the production versions of other apps
 - Being able to run using TLS, and with Rails apps in the production environment

But, I'll comment specifically on the things I've brought up relating to Docker above.

Using Guix enables the configuration within govuk-guix to be exceptionally reliable compared to Docker. Guix provides a rock solid base, and anything that is being pulled in externally (like code from Github, or gem packages from Rubygems) is fixed by a explicit hash, so that you know you're always getting either what you should be getting, or an error.

Of course, just starting from a rock solid base isn't the only problem, reproducibility issues can still crop up when building things, however Guix offers tooling to track these problems down.

While the RFC puts Guix in among the "Possible approaches to local development", at least for me, it's been a actual and real approach for local development, for the couple of years.

Even if the popularity of Docker, or alignment with future hosting options trump any and all of the issues mentioned in this Pull Request, I think it's still important thinking about the problems, as I think there may be ways of taking advantage by the approach that govuk-guix uses, but still using Docker, or even just mitigating some of the issues with using Docker.

On the more rigorous side, the only experience I have using Docker relates to using Guix to produce Docker images directly, skipping out the need for a Dockerfile, and thus gaining the benefits of using Guix, along with those of having a Docker image.

On the less rigorous side, I think it should be possible to at least mitigate some of the risk of using Dockerfiles by being very careful about their construction. Don't put things in the same Dockerfile if you aren't always happy with the behaviour of both of them changing when the image is rebuilt. Or to put it differently, split Dockerfiles up so that you have a fixed image as a base (or maybe a hierarchy of base images) as this will allow more control over at least when bits of the resulting image change (even if you don't have control over the how).

tijmenb:
@cbaines Thanks for the comments! I've updated the RFC a bit to reflect some of your points. The reproducibility point is important, and we should take care this week to look at making the Dockerfiles resilient to external change!

PR #109, opened 2019-07-30
RFC 109 Use AWS CloudFront as primary CDN

chrisfarms:
## What

As a follow up from a first attempt at improving GOV.UK DOS/DDOS protections we believe an architectural move to a CloudFront based CDN/ingress point will meet the desired needs with the least complexity.

## Sign off, Challenges & Open Questions

* [ ] Has had eyes from RE:GOV.UK
* [ ] Has had eyes GOV.UK programme
* [ ] Has had eyes from Cyber 
* [ ] All functionality currently provided by Fastly at the edge have been considers and mentioned in proposal or highlighted as risks (ie vcl used? split brain routing?)
* [ ] We have considered the impact to the failover to non-aws mirrors and added it to proposal/impact
* [ ] We have considered the potential impact on cost
* [ ] We have considered the potential impact on changes to cache invalidation
* [ ] We have considered how A/B testing (split testing) will be implemented
* [ ] We have identified key SLIs and SLOs for the performance of the CDN based on existing data (ie SLI:average-TTBF, SLO:<Xms)
* [ ] We have documented the TLS policies that we want enabled (we still want TLSv1) and that TLS1.3 support we not be available until later in the year and what risk this poses


rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Proposal
+
+_High level proposal for how to meet the needs defined in `Problem` definition_
+
+### Implementation
+
+We should complete the migration of the publishing backends into AWS. This is
+important because attempting to protect the backends behind CloudFront _before_
+completing the migration will lead to significant added copmlexity for
```
erino: There's a typo, copmlexity should be complexity

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+configuration and invalidations, additional traffic costs and potentially hard
+to reason about traffic routes.
+
+We will add a CloudFront distribution configured to replicate the
+routing/origins, cache-control and stale reads currently provided by fastly.
+This should operate in paralell to the exisiting Fastly CDN so as not to impact
```
erino: parallel not paralell

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+  unfortunate expense for a similar product.
+* Fastly does not provide a way of managing rate-limiting, you could
+  potentially approximate rate-limiting using vcl but it would be fragile and a
+  burden to maintain
+
+#### CloudFront -> Fastly
+
+* Run CloudFront in a non-caching configuration IN FRONT of Fastly, to gain the rate-limiting and acls at the edge
+* We would be double spending on traffic, storage and latency
+* We would be increasing the support/maintainability burdon
```
seb-szy: small typo: burdon replace with burden

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
+# Use CloudFront as primary ingress gateway
+
+## Summary
+
+Move the publishing backends into AWS and use CloudFront as the primary CDN
+(Content Distribution Network) for ingress to GOV.UK to take full advantage of
+the protection and support aforded to us from our AWS Advanced support contract
```
seb-szy: typo: aforded replace with afforded

seb-szy:
```suggestion
the protection and support afforded to us from our AWS Advanced support contract
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+and simplify our architecture.
+
+## Problem
+
+### Motivation
+
+GOV.UK should harden their public facing ingress gateways in order to respond
+to (if not mitigate) the growing threats from Denial Of Service, Distributed
+Denial of Service, and common automated vunerability scanning tools
```
seb-szy:
```suggestion
Denial of Service, and common automated vulnerability scanning tools
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+The current architecture of GOV.UK has Fastly as a CDN in front of an
+Application Load Balancer which is protected by a WAF (Web Application
+Firewall) and AWS Shield Advanced's DDoS detection/support.
+
+[TODO: pic]
+
+This architecture has the following problems:
+
+* The ALB/WAF is unable to perform rate limiting functions as traffic will be from Fastly
+* DDoS detection and protection will be inaffective as traffic will only be from Fastly
```
seb-szy:
```suggestion
* DDoS detection and protection will be ineffective as traffic will only be from Fastly
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+* The publishing backends are left unprotected by the WAF/Shield Advance as they are yet to be migrated to AWS
+
+### Goals
+
+* Provide equivilent cache functionality as currently (pass-through cache / stale reads)
```
seb-szy:
```suggestion
* Provide equivalent cache functionality as currently (pass-through cache / stale reads)
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+* Provide a configuration point at the edge for applying ip-based restrictions
+* Provide a configuration point at the edge for applying rate limiting restrictions
+* Provide a configuration point at the edge for applying basic sqli/xss detection
+* Provide detection and options for dealing with large scale distributed attacks
+
+### Non-Goals
+
+* Replicating every Fastly feature with CloudFront (some may be better implemnted elsewhere)
```
seb-szy:
```suggestion
* Replicating every Fastly feature with CloudFront (some may be better implemented elsewhere)
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+
+As a Developer/ReliabilityEngineer I want visibility of potential DDoS attacks
+that could be impacting my service so that I can narrow the scope of
+troubleshooting under these conditions and begin communication with my cloud
+provider for mitigation options as soon as possible.
+
+##### SQL Injection
+
+As a Developer maintaining a database connected application I want requests from
+automated tooling that scans for common SQL Injection vunerabilities to be
```
seb-szy:
```suggestion
automated tooling that scans for common SQL Injection vulnerabilities to be
```

rfc-107-cloudfront-at-edge.md
seb-szy:
```suggestion
completing the migration will lead to significant added complexity for
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+### Implementation
+
+We should complete the migration of the publishing backends into AWS. This is
+important because attempting to protect the backends behind CloudFront _before_
+completing the migration will lead to significant added copmlexity for
+configuration and invalidations, additional traffic costs and potentially hard
+to reason about traffic routes.
+
+We will add a CloudFront distribution configured to replicate the
+routing/origins, cache-control and stale reads currently provided by fastly.
```
seb-szy:
```suggestion
routing/origins, cache-control and stale reads currently provided by Fastly.
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+### Cost comparrison
+
+TODO: compare relative costs between Fastly/CloudFront (traffic, storage,
+cache-invalidation estimates etc) - exclude WAF costs as these would be same
+
+### Risks & Mitigations
+
+Fastly takes the brunt of the load of GOV.UK, any changes to the caching
+behaviour risk altering the load that the applications are currently tuned to
+handle. This can be mitigated by building the new CDN route in paralell,
```
seb-szy:
```suggestion
handle. This can be mitigated by building the new CDN route in parallel,
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+testing througherly and distributing only a fraction of the traffic (DNS/fastly
```
seb-szy:
```suggestion
testing thoroughly and distributing only a fraction of the traffic (DNS/Fastly
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+load balancing/ to the new system and ramping up to build confidence.
+
+Adding IP based rate-limiting has the potential to block organisations who have
+large numbers of users behind NAT, for example, univerisities. This can be mitigated by
+tuning the rate-limiting rules based on historical request data as well as scoping the
+rules to only the request paths/methods that have the most potential for lowering reliability
+(if the frontends can actually handle GET requsts at 100req/s probably don't
+let rate limiting kick in until it's nearer that number - DDOS is a seperate
```
seb-szy:
```suggestion
let rate limiting kick in until it's nearer that number - DDoS is a separate
```

rfc-107-cloudfront-at-edge.md
```diff
@@ -0,0 +1,197 @@
...
+issue that cannot be resolved by rate-limiting alone)
+
+SQL injection and XSS blocking carry the risk of false posatives, where
```
seb-szy:
```suggestion
SQL injection and XSS blocking carry the risk of false positives, where
```

thomasleese:
Fastly [currently enables our AB testing](https://docs.publishing.service.gov.uk/manual/ab-testing.html). Is there an equivalent for CloudFront, and should we mention something about that in the RFC?

---

We currently use [soft purging](https://docs.fastly.com/guides/purging/soft-purges) in the [Cache Clearing Service](https://github.com/alphagov/cache-clearing-service/commit/5ce3e3118b6b16f4b2594fa0318704cb85b66fb1) but [I couldn't find an immediate answer in the docs as to whether CloudFront supports this](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html#invalidating-objects-api), should we add that as a risk?

chrisfarms:
@thomasalanlee I suspect A/B testing type scenarios would be best implemented using a combination of a header as part of the cache key + an ALB's advanced routing stuff ... or even something with lambda at edge to replicate the logic ... but yes we should def make sure this is addressed in this doc, I have added a checkbox to the PR checklist

erino:
One other point to consider, is how moving to CloudFront will affect end user performance over a variety of connection types. I'm not exactly sure how we'd test this.

chrisfarms:
@erino yes I suspect this should be one of the considerations best evaluated when running both system in parallel ... but I would be surprized to see any difference in network/resource performance, in fact CloudFront has a larger network which will be closer to the origins, so it may even be an improvement :)

I think the biggest differences will be in how cache invalidations will work... CloudFront is extremely good at placing cached items very close to people over their large network, the flip side of this is invalidating that cache on demand is likely slower than Fastly

chrisfarms:
Thanks for fixing my terrible spelling @erino and @seb-szy  .... @seb-szy I have added the image you sent me :+1:

Nooshu:
Worth noting that CloudFront currently has a broken implementation of [HTTP/2 prioritisation](https://github.com/andydavies/http2-prioritization-issues). This is something we currently don't use due to the use of sharding with the assets domain, but is on the list web performance optimisations to be enabled when possible.

I assume CF also supports TLSv1.3? This has just been enabled with Fastly (although all nodes don't support it yet, so users may not see it). This has some security implications and also some small web performance implications in some instances (0-RTT and 1-RTT negotiation rather than 2-RTT).

The best way to test the performance impact of this change on users would be to collect Real User Monitoring (RUM) data using a service like Lux with Speedcurve or New Relic. Unfortunately these services can be expensive. We could use synthetic testing tools like Web Page Test, Sitespeed.io, Google Lighthouse, but these only give a very small insight into the change under very specific conditions.

nickcolley:
> The best way to test the performance impact of this change on users would be to collect Real User Monitoring (RUM) data using a service like Lux with Speedcurve or New Relic

It feels too late to test the impact of this change on users after it has shipped, could we do some preliminary testing before we've committed to much to this to ensure we don't make a negative change for our users?

Nooshu:
> I suspect this should be one of the considerations best evaluated when running both system in parallel
I got the impression that they were both going to be run at the same time? Is this to facilitate making the final decision or just part of the migration process over to the new CDN?

kevindew:
Another aspect of Fastly that we use and should be considered is the logging.

We currently have logs streamed from fastly to monitoring boxes and also to S3 (as JSON) which are then used for Athena queries - particularly useful in assessing impact of incidents - this could be tricky during a phased migration

Probably a dumb question too: when I started reading this and saw "publishing backends" I interpreted this to mean moving the stuff that we don't currently have behind a CDN, which are normally publishing applications e.g. [whitehall](https://whitehall-admin.publishing.service.gov.uk/), behind CloudFront - but I now interpret this to be replacing Fastly with CloudFront. Is that correct? and if so would we be putting applications not currently behind a CDN behind CloudFront?

chrisfarms:
> I assume CF also supports TLSv1.3? 

@Nooshu CloudFront does not yet have TLSv1.3 support released. As I understand it, CloudFront (and most other AWS products), make use of the [s2n](https://github.com/awslabs/s2n) library which is developed in the open and the feature can be tracked [here](https://github.com/awslabs/s2n/milestone/1). We can probably get a better estimate of timescale from our AWS TAM of when we would be able to enable it to get the 0RTT benefits for those clients that support it, but I would hope that this would not be a blocker?

> I got the impression that they were both going to be run at the same time? Is this to facilitate making the final decision or just part of the migration process over to the new CDN?

@nickcolley @Nooshu Yes running in parallel gives us the ability monitor the impact and back-out or re-evaluate if there are any undesirable consequences.

Ideally we could identify key metrics/indicators (ie latency, response times, error rate) and agree [SLOs](https://en.wikipedia.org/wiki/Service-level_objective) for the current expectations of system so we can use data to ensure we are meeting our needs. I have added a checkbox to this doc to expand on this point of testing/monitoring performance.

> We currently have logs streamed from fastly to monitoring boxes and also to S3 (as JSON) which are then used for Athena queries

@kevindew CloudFront supports shipping logs in S3, which could be similarly queried via Athena or shipped onwards. Running the two systems in tandem with multiple ways to query logs would not be ideal and if having a single "pane of glass" for all logs from the edge was important enough, then this could be investigated, but given we would not expect to be running both systems for long, we should probably not attempt to preempt the need until it arises.

> [does this mean] moving the stuff that we don't currently have behind a CDN, (e.g. whitehall), behind CloudFront [or] replacing Fastly with CloudFront

@kevindew Both.

Applications not currently behind a CDN (eg whitehall) will be moved behind CloudFront ... however the move (for those applications) is driven by a need for Web Application Firewall and DDoS protection at the edge network,  NOT by a need for a caching layer. The CloudFront distribution for these applications would likely not be configured to cache anything at all.

Elsewhere (cache frontends, etc) CloudFront will replace the existing function of Fastly as a caching layer in addition to providing Web Application Firewall and DDoS protection.

erino:
We also need to consider TLS 1.0/1.1 support. We have an extension from Fastly for continued TLS 1.0 support (I'm not sure for how long). We may also need to support older versions of TLS for publishers. It looks like [CloudFront supports TLS 1.0](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/secure-connections-supported-viewer-protocols-ciphers.html#secure-connections-supported-ciphers), but it would be good to know for how long.

Nooshu:
@chrisfarms No TLSv1.3 wouldn't be considered a blocker. We're moving towards HTTP/2 in the near future, and considerations need to be made for HTTP/3 + (i)QUIC in the somewhat more distant future (which will most likely use TLSv1.3). I'd hope this move is one that we can stick with for a number of years rather than having to change again because they don't iterate. So good to see it on their roadmap.

In terms of parallel running I'd like to start collecting RUM data just before the transition so we can back-up the performance considerations with data. Once you have more info on timings let me know.

@erino good call on TLS 1.0/1.1 support. We currently have an agreement with Fastly for perpetual support of these old protocols. In terms of deprecation, the date for this is for us to dictate and communicate to users. We should make *100% sure* that we can get the same agreement with CloudFront. We don't want to be in a situation where a date is dictated to us by a third-party and we don't have a communication plan in place to prepare effected users.

Nooshu:
Just raising another issue with Cloudfront that impacts frontend performance. @aliuk2012 ran a couple of tests using Web Page Test ([seen here](https://www.webpagetest.org/result/190906_G0_c280273ff8ec0027aa4da9e066838a26/1/details/#waterfall_view_step1)) on Digital Marketplace. There's some very curious results with requests 1 and 2 in the waterfall.

Screen Shot 2019-09-06 at 08 22 52

After some investigation into what exactly they are, it turns out they are requests to the Online Certificate Status Protocol (OCSP) responder from the client to check see if the certificates of the site have been revoked. This is a known issue with OCSP, but to stop clients having to make this third-party request it is often recommended to enable OCSP stapling. This is where the server will periodically check the certificates and then send this information onto the client in the TLS handshake phase. This eliminates individual clients from having to make the request themselves.

Now according to various documentation across the web, Cloudfront does or doesn't support this feature:
* According to is TLS fast yet, [CloudFront does](https://istlsfastyet.com/#cdn-paas)
* According to SSLLabs, [CloudFront doesn't](https://www.ssllabs.com/ssltest/analyze.html?d=www.digitalmarketplace.service.gov.uk&s=99.84.224.38&hideResults=on&latest)
* From our own tests using openssl, [it depends](https://gist.github.com/Nooshu/836949de9d7c6279e3b688f432ed5099)

According to [this documentation](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#request-custom-ocsp-stapling) from AWS:

> When CloudFront receives a lot of HTTPS requests for the same domain, every server in the edge location soon has a response from the CA that it can “staple” to a packet in the SSL handshake; when the viewer is satisfied that the certificate is valid, CloudFront can serve the requested object. If your distribution doesn’t get much traffic in a CloudFront edge location, new requests are more likely to be directed to a server that hasn’t validated the certificate with the CA yet. In that case, the viewer separately performs the validation step and the CloudFront server serves the object. That CloudFront server also submits a validation request to the CA, so the next time it receives a request that includes the same domain name, it has a validation response from the CA.

Since DMP is quite a popular service and the test was run in London, I would imagine there'd be a large amount of traffic to this CloudFront edge location. So I'm not sure why a client would actually have to be making this request, as stapling at the edge should be accomedated for.

@tlwr created a script to test the use of stapling over a large number of connections (which can be [seen here](https://gist.github.com/Nooshu/836949de9d7c6279e3b688f432ed5099/edit)). The results of this test are very strange:

```
# Is OCSP stapling used by Cloudfront?
# code        _count
# --------------------------
# 1 (no)           151
# 0 (yes)          49
```

Only 1/3 of connections to Cloudfront actually use stapling. Meaning 2/3 of clients may be forced to do their own OCSP checking via the OCSP responder. Now if we compare this to Fastly, the CDN we currently use on GOV.UK:

```
# Is OCSP stapling used by Fastly?
# code        _count
# --------------------------
# 0 (yes)          201
```

100% of connections are using stapling. 

I have a concern that moving to Cloudfront, where 2/3 of clients may need to OCSP check themselves, will have a large impact on web performance. As you can see from the waterfall image above, these 2 requests are blocking. The SSL negotiation cannot complete until a responce from this third-party is recieved, leading to a SSL negotiation taking 1 second, and the time to first byte at 2.4 seconds.

If we compare this against a waterfall where stapling was used in the image below (one of the 1/3 of connections), the pink line (SSL negotiation) takes 348ms compared to 1058ms for unstapled. That's a 204% increase in SSL negotiation time.

Screen Shot 2019-09-09 at 13 52 45

I have raised this with AWS for comment, but there looks to be known issues with Cloudfront and OCSP stapling. So it may require major changes on their end to fix it.

kevindew:
@chrisfarms this RFC seems to have stalled. Should we close this or are you likely to move this forward?

chrisfarms:
Yes if it's no longer relevant please close.

This RFC was created in response to a desire to make use of the AWS Shield Advance and WAF offerings for DoS protection, which only really make sense at the network edge (not behind fastly) ... if nobody is pushing that specific need any longer then I do not think this RFC makes sense to pursue.

kevindew:
@chrisfarms Thanks for confirming. 

Shame we didn't get this further as it sounded pretty cool, thanks for digging into it 👍

PR #110, opened 2019-08-23
RFC 108 Include frontend gem component assets individually

andysellick:
Alternative for the current model of including GOV.UK component assets in frontend applications.

👉 [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/bc1ad0677abcbfdf48ba6bc0b2e0f24557e48e9a/rfc-108-including-gem-component-assets.md)

I've also started a [spike for the individual Sass inclusion](https://github.com/alphagov/govuk_publishing_components/pull/1159).

### DIFF
--- a/rfc-108-including-gem-component-assets.md
+++ b/rfc-108-including-gem-component-assets.md
@@ -9,7 +9,7 @@ The [current method for including the components](https://github.com/alphagov/go
 
 This also has some detriments:
 
-- all components are included regardless of whether they are used, so unused CSS/JS may be included
+- all components are included regardless of whether they are used, so unused CSS/JS will be included
 - only publically facing components can be added to the gem, or unused CSS/JS would be included (we now have components that need to be shared across only private applications that use the gem)
 
 ## Proposed solution (Sass)
@@ -33,11 +33,13 @@ The component_support sass file would include any needed `govuk-frontend` import
 
 This solution would allow us to add any components we like to the gem without worrying that they are needlessly adding page weight to the publically facing GOV.UK.
 
+It should be possible to configure the gem so that both this new approach to consuming Sass and the old approach co-exist, so we don't have to upgrade all the apps at once.
+
 ## Proposed solution (Javascript)
 
 We can do something similar with Javascript, however the initialisation of specific modules may be more complicated and require more thought. Additionally knowing whether or not to include jQuery could be difficult.
 
-As a MVP Javascript could be left as it is for now.
+It should be possible to configure the gem so that both this new approach to consuming JS and the old approach co-exist, so we don't have to upgrade all the apps at once.
 
 ## Finding which components are in use
 
--- a/rfc-108-including-gem-component-assets.md
+++ b/rfc-108-including-gem-component-assets.md
@@ -1,0 +1,89 @@
+# Include specific component assets in applications
+
+## Problem and background
+
+The [current method for including the components](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#4-include-the-assets) in a frontend application (such as `government-frontend`) is to include `_all_components.scss` and `_all_components_print.scss` in the Sass of that application. This has some benefits:
+
+- once the application is configured to use the gem, no further configuration is needed
+- all components are automatically included in the application
+
+This also has some detriments:
+
+- all components are included regardless of whether they are used, so unused CSS/JS may be included
+- only publically facing components can be added to the gem, or unused CSS/JS would be included (we now have components that need to be shared across only private applications that use the gem)
+
+## Proposed solution (Sass)
+
+Instead of including all of the component Sass, each application should be configured to only include the components that it needs. For example, in an applications `_application.scss`:
+
+```
+// currently
+@import 'govuk_publishing_components/all_components';
+```
+
+```
+// proposed
+@import "govuk_publishing_components/component_support";
+@import 'govuk_publishing_components/components/accordion';
+@import 'govuk_publishing_components/components/button';
+// (etc.)
+```
+
+The component_support sass file would include any needed `govuk-frontend` imports plus any mixins or variables from sass in the gem. A similar solution could be applied to print styles for components.
+
+This solution would allow us to add any components we like to the gem without worrying that they are needlessly adding page weight to the publically facing GOV.UK.
+
+## Proposed solution (Javascript)
+
+We can do something similar with Javascript, however the initialisation of specific modules may be more complicated and require more thought. Additionally knowing whether or not to include jQuery could be difficult.
+
+As a MVP Javascript could be left as it is for now.
+
+## Finding which components are in use
+
+Adding a not-in-use component into an application with this new model would be a relatively safe procedure, as the developer would notice immediately if the styles and Javascript for that component had not been included in the application. However for the future it would be useful if we had some kind of tool that helped us to avoid leaving components unstyled or non-functional.
+
+We could search through the code base and write a test for that application to ensure the right assets are being included. This test would need to be kept up to date as components are added and removed.
+
+We could also look at altering the component guide to either mark components in use or put them in a separate list. That way we could look at the component guide for an application and immediately know which components are in use and which are not.
+
+## Certainty that a component renders correctly in isolation
+
+If the component model is changed so that only some of the components could be included by an application we would need certainty that they would render correctly in isolation. At the moment the CSS for all components is included as a single file, so it all works.
+
+Separating out these styles is problematic due to each component's dependence on external mixins and variables (from govuk-frontend and the gem itself). An application can `@import button` only so long as it also imports other stuff beforehand.
+
+If we wanted to test a component truly in isolation (in the component guide page for that component) we'd need a CSS file that only included the styles for that component. The output for this would be inappropriate for use in applications so we might end up having two stylesheets for each component, which wouldn't be ideal.
+
+(The Design System site shows each component inside an iframe, but this seems to include all component styles in the iframe, not just the styles for that specific component. Iframes also pose problems with regard to setting an appropriate height)
+
+## Pros and cons
+
+Pros:
+
+- an application only has to include the components it needs
+- the size of the CSS file for each application is reduced (assuming the application uses some but not all of the components. If it uses all of the current components then there's no change)
+
+Cons:
+
+- we have to be vigilant to ensure components in applications are included properly and not accidentally removed
+
+Unanswered questions:
+
+- how do we include only the required Javascript?
+- how does the component guide look in an application that is only using some of the components? Do those unused components appear unstyled? Or not appear?
+
+## Implementation
+
+### Testing
+
+Testing should remain broadly the same. Tests will exist for each component and when the tests are run in the gem all tests should be run. We might need to rewrite the initialisation for some of the Javascript tests depending on decisions made regarding JS inclusion.
+
+### Backwards compatibility
+
+For a new release we could keep the existing `_all-components.scss` file so we don't break any applications that still rely on it. This would be removed at a later date once all applications have been updated to use the agreed new system.
+
+## Next steps
+
+Spike it locally.
+
--- a/rfc-108-including-gem-component-assets.md
+++ b/rfc-108-including-gem-component-assets.md
@@ -1,6 +1,12 @@
 # Include specific component assets in applications
 
-## Problem and background
+## Summary
+
+The govuk_publishing_components gem provides shared components to frontend applications on GOV.UK - chunks of consistent and reusable frontend such as buttons and form elements.
+
+The gem only allows all components to be included in an application, even if they're not used by that application. This RFC proposes changing the gem to allow only required components to be included, to reduce unnecessary page weight.
+
+## Problem
 
 The [current method for including the components](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#4-include-the-assets) in a frontend application (such as `government-frontend`) is to include `_all_components.scss` and `_all_components_print.scss` in the Sass of that application. This has some benefits:
 
@@ -12,7 +18,7 @@ This also has some detriments:
 - all components are included regardless of whether they are used, so unused CSS/JS will be included
 - only publically facing components can be added to the gem, or unused CSS/JS would be included (we now have components that need to be shared across only private applications that use the gem)
 
-## Proposed solution (Sass)
+## Proposal
 
 Instead of including all of the component Sass, each application should be configured to only include the components that it needs. For example, in an applications `_application.scss`:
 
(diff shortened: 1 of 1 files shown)


rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,66 @@
...
+```
+// currently
+@import 'govuk_publishing_components/all_components';
+```
+
+```
+// proposed
+@import 'govuk_publishing_components/components/accordion';
+@import 'govuk_publishing_components/components/button';
+// (etc.)
```
injms: I _think_ the components might need initialising as well?

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,66 @@
...
+```
+// currently
+//= require govuk_publishing_components/all_components
+```
+
+```
+// proposed
+//= require govuk_publishing_components/components/accordion
+//= require govuk_publishing_components/components/button
+// (etc.)
```
injms: I _think_ the components might need initialising as well?

andysellick:
I've updated this following a conversation with @Nooshu to include an alternative option.

rfc-108-including-gem-component-assets.md
injms: In `all_components.js`, there's a function called that initialises all of the components (`window.GOVUKFrontend.initAll()`). I think that there's an initialisation function for each component that needs to be used to make sure that the component's JS is started.

36degrees: For reference, that function is here:

https://github.com/alphagov/govuk-frontend/blob/5a514ce69db68ec17f681cdec8d82769c9c241b9/src/govuk/all.js#L12-L62

andysellick: Yeah, the JS is more complicated than I've made out here. As a first pass, we could still include all of the JS. It shouldn't cause any problems and won't be worse (in terms of filesizes) than what we already have, and in the meantime work on a better solution.

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,111 @@
...
+
+```
+<link rel="stylesheet" media="screen" href="https://assets.publishing.service.gov.uk/frontend/application-a6d734239cc6bfc2534ca11a838575ef7b.css" crossorigin="anonymous" integrity="sha256-73t28PTeOrcJZQJGEPkS9ww=">
+
+<link rel="stylesheet" media="screen" href="https://assets.publishing.service.gov.uk/frontend/components-76f0f4de3ab70965024610f912f70c.css" crossorigin="anonymous" integrity="sha256-ptc0I5zGv8JTTKEag4V1=">
+```
+
+The [rails fingerprinting](https://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark) would ensure that the file is cached between applications, based on its content and size (not date).
+
+This would only work if the components stylesheet in each application was the same, so the solution described above of only including the components needed would not work. Instead, we would need to separate out the public facing component styles from the private application styles. For example (in the gem):
```
vanitabarrett: Not sure if I'm reading this right - would this mean we wouldn't be choosing which components to include in each app, but just choosing whether to include public or private components?

andysellick: Yes.

kevindew:
Yay. I’m really keen on us to work towards doing this, we must be slowing building up quite a big pile of unused JS and CSS in apps. This was quite a clear problem when the accessible autocomplete component was added that the JS was inflicted on everything. So it’d be nice to have an idea how to keep adding components without incurring costs everywhere. 

What I’m not too sure about is the definition of public and private apps? Is that publisher and frontend GOV.UK apps? Could that be defined in the RFC? And if I’m correct in my understanding I don’t think we should distinguish them as two different things, rather than any GOV.UK application can just use any combination of components. 

There’s been talk of doing this before and it used to be partially implemented to do this (at least in CSS) but there were a number of stumbling points. It’d be good if the RFC or discussion here could identify how to resolve them. 

First, would we be defining what parts of GOV.UK frontend is assumed to be available (if any) for all components. At one extreme we could just be importing the GOV.UK frontend bits that are needed into each component and the opposite extreme we could assume that the entirety of GOV.UK frontend is available from a separate style sheet so only variables are imported. Similarly would all JS be isolated? For example each component import govuk modules etc. I’m not sure if asset pipeline does a form of tree shaking but hopefully it does otherwise jquery could be included once for every component 😬. We also rely on the GOV.UK frontend initAll but presumably that can be moved to doing an init on each component. 

Second, how would we be changing GOV.UK Publishing Components to have the preview system only use isolated resources to ensure that components are not relying on any styles provided by other components. This seems a really important step for the preview experience to be an accurate experience. I’d guess we’d maybe want to use iframes on previews to separate them from the presentation code, is that what the design system does? 

Thirdly, would this have any impact on testing strategies, particularly from a JS perspective? 

Presumably we’d still keep the all files approach for backwards compatibility and just recommend importing individual components? If so would importing all become deprecated or would that just be an alternative way of using the components?

I’m not very keen on the caching alternative as it doesn’t feel particularly in the spirit of components.

ChrisBAshton:
> I don’t think we should distinguish [public and private apps] as two different things, rather than any GOV.UK application can just use any combination of components.

+1

> Would we be defining what parts of GOV.UK frontend is assumed to be available (if any)

I think it's fair to assume the entirety of GOV.UK frontend is available. It would then be up to the component to use only the appropriate mixins/variables it needs, and not to pull in more code than necessary.

> how would we be changing GOV.UK Publishing Components to have the preview system only use isolated resources to ensure that components are not relying on any styles provided by other components. This seems a really important step for the preview experience to be an accurate experience. I’d guess we’d maybe want to use iframes on previews to separate them from the presentation code, is that what the design system does?

The BBC defines some [minimal global CSS](https://github.com/bbc/psammead/blob/latest/.storybook/config.js#L41-L56) (normalize, fonts, box-sizing) - all other styling should be included in the component via mixins etc. It uses [Storybook](https://bbc.github.io/psammead/) as a component previewer, which uses iframes under the hood, but had some difficulty accessibility-testing components as a result.

We could investigate [using shadow DOM to encapsulate styles](https://medium.com/bbc-design-engineering/goodbye-iframes-6c84a651e137), which doesn't have the accessibility and performance issues associated with iframes. We'd have to weigh up shadow DOM browser support against our intended component previewing audience.

injms:
I really like the standard being that nothing is automatically included, and that each component has to be added to the app. Whilst this is more work, it should make for a lighter and more considered frontend app. (On previous projects this was how I added GOV.UK Frontend piece by piece to avoid a negative effect for users.)

I'm confused about why public and private apps need to be defined, since the proposal would lead to adding each component's dependencies as needed - regardless of whether the app is public or private.

Regarding the specific example of jQuery as a dependency (before it's hopeful removal soon) - that could be included manually in `application.js`, since not all components need jQuery and it may well be not required in some apps at all.

I would caution against adding *any* global styles, since some strangely powerful unspecific selectors (for example, `a:link:focus`) are causing problems with the upgrade to version 3 of the of GOV.UK Frontend. I think that @ChrisBAshton's suggestion of using mixins would solve this problem for future-us and help make the code more maintainable.

The simplest solution for isolating each component seems to be`iframe`s - especially considering non-JavaScript testing and older browsers that we need to test components in. My understanding of web components is that they need JavaScript, so we couldn't then preview a component in a browser with JavaScript turned off. Off the top of my head, the accessibility testing could be done within the component iframe - this could also fix the recurring `Banner landmark must not be contained in another landmark` error.

I think this could be a change that's introduced in stages to avoid any breaking changes for now - maybe first added to the documentation as the preferred way to do things without actually turning the `_all_components` functionality off.

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Problem and background
+
+The [current method for including the components](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#4-include-the-assets) in a frontend application (such as `government-frontend`) is to include `_all_components.scss` and `_all_components_print.scss` in the Sass of that application. This has some benefits:
+
+- all components are automatically included in the application
+
+This also has some detriments:
+
+- all components are included regardless of whether they are used, so unused CSS/JS may be included
```
alex-ju: I think 'unused CSS/JS _will be_/_is_ included' - since no applications actually uses all of them

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,189 @@
...
+- only publically facing components can be added to the gem in order to prevent the previous problem
```
alex-ju: As discussed in the last `#govuk-frontenders` meeting we should avoid differentiating components based on where they are used. We can potentially rephrase that as: we are currently reserved in adding new components to the gem as we are concern about the performance of the public facing/frontend applications.

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,189 @@
...
+@import 'govuk_publishing_components/components/accordion';
+@import 'govuk_publishing_components/components/button';
+// (etc.)
+```
+
+This would mean we can add any components we like to the gem without worrying that they are needlessly adding page weight to the publically facing GOV.UK.
+
+## Javascript
+
+We can do something similar with Javascript, however the initialisation of specific modules may be more complicated and require more thought. Additionally knowing whether or not to include jQuery could be difficult.
```
alex-ju: In the [absence of a bundler](https://github.com/alphagov/govuk_publishing_components/issues/505) I would suggest that the JavaScript initialisation should be done in each application, after importing the required files.

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,189 @@
...
+
+As a MVP Javascript could be left as it is for now.
```
alex-ju: I think this would make things confusing and will not help with reducing the JavaScript load on public facing/frontend application, which for me is more worrying than the CSS.

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Finding which components are in use
+
+Adding a not-in-use component into an application with this new model would be a relatively safe procedure, as the developer would notice immediately if the styles and Javascript for that component had not been included in the application. However for the future it would be useful if we had some kind of tool that helped us to avoid leaving components unstyled or non-functional.
+
+In the first instance we could search through the code base and write a test for that application to ensure the right assets are being included. This test would need to be kept up to date as components are added and removed.
```
alex-ju: If we're concerned about imports not being aligned with the usage we could potentially have a coverage test for components, in a similar fashion with the work @benthorner did with locales in https://github.com/alphagov/content-publisher/pull/1018.

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,189 @@
...
+
+We could also look at altering the component guide to either mark components in use or put them in a separate list. That way we could look at the component guide for an application and immediately know which components are in use and which are not.
+
+### Certainty that a component renders correctly in isolation
```
alex-ju: This is already done in `govuk-frontend` we could use a similar model.

chao-xian:
@andysellick I think this is a good as it reduces weight for our users. I would like to see this happen _after_ we have got all of the frontend apps on GOV.UK using the components from the gem, just in terms of that being a higher priority in front of us.

rfc-108-including-gem-component-assets.md
andysellick: `govuk-frontend` uses an iframe to render components but appears to include all of the components CSS in that iframe. That does give some isolation but I think we're looking for the kind of isolation where a single component works by itself.

I've attempted this and there's some problems, better to discuss offline.

andysellick:
Hi everyone. I've updated and streamlined this RFC following feedback. There are still some unanswered questions but I think the thing to do now is agree whether this is the way forward and then try to solve the problems.

I've started a spike on some of the stuff [in this PR](https://github.com/alphagov/govuk_publishing_components/pull/1159).

kevindew:
@andysellick this RFC seems to have stalled. What do we need to bring this to be either something that can be accepted or rejected in a reasonable timeframe? Alternatively should it just be closed?

andysellick:
@kevindew thanks for raising. This was discussed at a meeting of the GOV.UK FE team last week. The RFC has been rewritten following feedback ☝️ and @alex-ju was going to re-review it to see if we can get it approved.

kevindew:
> @kevindew thanks for raising. This was discussed at a meeting of the GOV.UK FE team last week. The RFC has been rewritten following feedback ☝️ and @alex-ju was going to re-review it to see if we can get it approved.

Great news. After @alex-ju has reviewed it shall we set a closing date on it? We may want to try re kick off communication about it since it's been rewritten. Do we (as a programme) need/want to try resolve any of the unanswered questions to complete this?

rfc-108-including-gem-component-assets.md
alex-ju: I think it's worth mentioning that JavaScript could also be consumed individually by importing/requiring only the needed JavaScript files from govuk_publishing_components then initialise components individually or by using the `modules.js` initialiser.

andysellick: I've updated this line, I've left it deliberately vague as it seems like the specifics of this can be worked out during spikes - the overall intent is that JS files be consumed individually.

andysellick:
I think this proposal has now been rewritten to address the issues raised. It might lack full detail but we'll work out how to actually do this once we've decided this is what we want to do.

I'm going to set a deadline on this PR for comments of the 9th of December (a week on Monday).

rfc-108-including-gem-component-assets.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Proposed solution (Javascript)
+
+We can do something similar with Javascript, however the initialisation of specific modules may be more complicated and require more thought. Additionally knowing whether or not to include jQuery could be difficult.
+
+It should be possible to configure the gem so that both this new approach to consuming JS and the old approach co-exist, so we don't have to upgrade all the apps at once.
+
+## Finding which components are in use
+
+Adding a not-in-use component into an application with this new model would be a relatively safe procedure, as the developer would notice immediately if the styles and Javascript for that component had not been included in the application. However for the future it would be useful if we had some kind of tool that helped us to avoid leaving components unstyled or non-functional.
```
vanitabarrett: I think this might be useful if we have to pull in print styles separately as these could more easily be missed.

alex-ju:
> This will be a great way forward. Looking forward to seeing the spike.

The spike is already up since October – waiting for this RFC to get merged https://github.com/alphagov/govuk_publishing_components/pull/1159

kevindew:
Hey @andysellick I've caught up on this now and thought it might be useful to add my comments here in advance of the frontenders meeting. Thanks again for your continued work on this - it'll be a great problem to solve 👍 

What I'd most like to resolve in this RFC are a number of areas of ambiguity where we indicate the changes are going to cause problems but don't indicate if we all going to solve them or if we plan to just live with them. For example: rendering in isolation, component guides continuing to work for apps and approach to JavaScript. These make me wonder if the RFC is proposing a solution to the problem or just a small step towards helping an aspect of the problem that instead creates new fresh problems.

So I find my interpretation of the RFC varies between two very different outcomes:

### Pessimistic - where we assume any ambiguous areas remain unresolved in the mid term

Benefits:

- Only selected CSS components are used in apps, making CSS downloads smaller.

Drawbacks:

- All JavaScript components are included in all apps, continuing to prevent ability to add components that aren't used by all apps;
- We will now have 2 different routes to include CSS components in apps that'll need maintaining going forward;
- Apps that use the approach of selecting which CSS components will have component-guides that look broken for any components they haven't included the CSS for;
- The only means to test if a components rendering isn't impacted by another component is to use the component in a different application and see if it's broken.

### Optimistic - where RFC commits to resolving the problems before roll out

Benefits:

- Users of govuk_publishing_components have the means to select which components they need for JS and CSS, allowing reduced download sizes of both and the ability to add components that not all apps have to use.

Drawbacks:

- More complex process to add components to apps;
- Period of transition where GOV.UK applications need to switch from the deprecated all_components approach to the new pick and choose approach.

What I'd like to see is for this RFC to remove the ambiguities to indicate that we will either solve the problems or not. It's ok if we don't know the exact answer to the solution but it'd be good to say whether the roll out of this would be blocked or not by solving problems. This would help indicate whether accepting this RFC also means accepting new problems it introduces or not.

Hope that makes sense and is helpful. I see other people approving this so I'm not sure if I'm misinterpreting things or worrying about issues others don't see as problems?

On the more minor side this RFC seems to have gone for a different structure to other RFCS and should be adjusted to follow the [template](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-000-template.md) I figure you can lay it out like:

<details>
<summary>Layout</summary>

**Summary**

What is govuk_publishing_components and how did applications currently make use of it

**Problem**

1) Can't add components that we're not happy with every app having
2) file download sizes are increased with library code not all apps need

**Proposal**

We change the recommendation of how SCSS is included into apps, while deprecating and continuing to support the all_components approach.
JS TBC <- Can we resolve this being a TBC somehow? it feels inconsistent to have different approaches, presumably we can take a similar approach to SCSS

Maybe breakdown benefits and risks of the proposal

Benefits:
- Reduced SCSS file sizes
- Switching to global include will offer compilation performance benefits

Drawbacks:
- Inconsistent approaches to JS and SCSS
- No means to test component is styled correctly in isolation
- Component guide is going to look broken in apps

</details>

andysellick:
@kevindew thanks for your input. I've rewritten the RFC to follow the layout template and also to solidify some of the language following your comments. 

In a nutshell, I agree that it should be optimistic, and that the only things we currently don't know how to do are the specifics of the JS (although this isn't blocking, as we can at worst keep the current system) and testing components in isolation (which I've spiked a little bit into, and haven't found a solution to yet). Reassuringly, due to the dependent nature of most of the component Sass, compilation errors in apps may at least alert us to some of the isolation problems.

andysellick:
Hi all, I've made changes to this RFC following the discussion we had yesterday. If you haven't already, can you either approve it or get back to me with comments by the end of the day, thanks!

[Latest version here](https://github.com/alphagov/govuk-rfcs/blob/bc1ad0677abcbfdf48ba6bc0b2e0f24557e48e9a/rfc-108-including-gem-component-assets.md).

andysellick:
Thanks everyone for contributing. I'm going to merge this now and start work on it when I can find time - any help anyone wants to offer will be much appreciated!

alex-ju:
Not that it matters too much at this point, but judging by the way this repo is setup (RFC numbers in sync with the PR number) this should actually be referred to as RFC 110.

PR #112, opened 2019-09-17
RFC-112: Remove SRI for 1st-party assets on GOV.UK

Nooshu:
TL;DR;
Due to the security requirements of SRI, `crossorigin="anonymous"` is required for these assets. This forces browsers to open a separate TCP connection to download this resource. Since TCP connections are expensive to open and maintain, this practice has a negative effect on frontend performance.


kevindew:
Thanks for raising this @Nooshu 👍 

As this represents a change to our security policy do we need to get some eyes on this from the cyber security team? I think scenarios where this can be exploited would be some sort of compromised DNS myself which fits your assessment. I think we already apply SRI inconsistently with it being done in most frontend applications (most frontend apps) but not all, and not religiously across our apps. So may not be the biggest alarm bells to remove it.

Are the options that were considered/rejected/not possible that allow SRI and re-use of TCP connections. I wondered if we could use use-credentials for crossorigin and also wondered whether we need to specify crossorigin if we were to use the www host name for the assets (since it's a bit of a hack how we use assets hostname for application assets and for uploaded assets from asset-manager that has caused surprises).

If this is to proceed I'm wondering about some of the consequences:

- Do we know how noticeable removing this would be in either a measure of user experience or statistics? 
- Do we know who would be undertaking the work to change GOV.UK apps for it?

DilwoarH:
This is everywhere `crossorigin="anonymous"` is being used:

https://github.com/search?q=org%3Aalphagov+%27crossorigin%3D%22anonymous%22%27&type=Code

kevindew:
@Nooshu this RFC seems to have stalled. Should we close this or are you likely to move this forward?

Nooshu:
After a number of conversations I've had, I've compiled the outcome of them for future reference.

**Pros**:
* It's always better to have an attacker have to touch more "things" to do damage, which is what SRI does. You are more likely to notice it.
* SRI could be useful for pages with a file upload / attachment functionality in case of a security issue that would allow for overwritten assets via this process.
* Good for "security posturing".

**Cons**:
* In terms of web performance not self-hosting assets is now considered an [anti-pattern](https://csswizardry.com/2019/05/self-host-your-static-assets/)
    * In reference to assets sitting on a 3rd-party CDN, not under our control like our use case: If the asset doesn't change, self-host. If it does change you can't use SRI anyway (manual update of the integrity is required)
* The GOV.UK assets domain is a blocker for the adoption of HTTP/2 and HTTP/3+QUIC in the future.
* For best performance all CSS, JS, Fonts should be served from the same domain as the HTML (so the correct h2 prioritisation can take place). If actioned we would be using SRI for assets on the same domain.
* SRI makes sense when you have to load an asset from a third-party domain which cannot or should not be self-hosted e.g. ad networks, font scripts. It doesn't make much sense for assets fully in our control.
* There will be a small size impact on HTML size due to hashes in the source
* SRI is also enabled on CSS, could be considered overkill for just styling
* 'crossorigin="anonymous"' is a requirement for SRI to work. Under HTTP/1.1 multiple "anonymous mode" TCP connections will need to be created to download all assets. Under HTTP/2 this may have a negative impact on performance since the protocol aims to stream and multiplex all assets over a single TCP connection. 

**Considerations**:
* How likely is it that the assets on the asset server under our control can be tampered with? 
* sha256 (used currently) is a [slow algorithm](https://medium.com/@davidtstrauss/stop-using-sha-256-6adbb55c608), so we should consider moving to sha512 if available.
* What is the impact on the build process? 
    * I believe it may just be a flag that is set in rails `integrity: true`, so easy to add / remove.
* Performance impact in terms of CPU time is probably quite small, but worth testing on a [low spec device](https://www.alcatelmobile.com/product/smartphone/alcatel1/alcatel-1x/) to be sure as can't find any data on it.
* What impact does this have on H2 and the number of connections created vs the correct H2 prioritisation from Fastly.

**Recommendation**:
Close this RFC for the moment. If the removal of the assets domain (https://assets.publishing.service.gov.uk) is ever moved forward to allow us to enable HTTP/2, then we should consider removing SRI.

kevindew:
Thanks @Nooshu 👍

PR #113, opened 2019-10-04
RFC 113: Expanding draft access for unauthenticated users

kevindew:
View rendered: https://github.com/alphagov/govuk-rfcs/blob/rfc-113/rfc-113-expanding-draft-access-for-unauthenticated-users.md

Hello :wave: this is an RFC about the expansion in access limiting that will be needed to support fact checking step-by-steps. Unfortunately the team is on a tight schedule so the work to implement this proposal will be ongoing during the time open for this RFC, so I'm kind of hoping that it won't be too controversial and be a useful source of documentation. This RFC will close on the 11th October.

TL;DR: We want to expand auth-bypass access to draft GOV.UK to allow multiple pages and their assets. As a user may not own these resources this process cannot be used to bypass access limiting on them. We will use link expansion to share an auth-bypass id to GOV.UK content. For assets we will allow a token to grant access to all draft assets (that don't have access limiting).

### DIFF
--- a/rfc-113-expanding-draft-access-for-unauthenticated-users.md
+++ b/rfc-113-expanding-draft-access-for-unauthenticated-users.md
@@ -6,24 +6,28 @@ To view content on the [GOV.UK draft stack][govuk-draft] users have to either
 be authenticated with [GOV.UK signon][signon] or have a token that grants
 them access to that piece of content. Sensitive content is flagged as
 access limited, this limits drafts to authenticated users that meet a criteria
-(user id or organisation id) or users with the token.
+(user id or organisation id) or users that have a valid token.
 
 The reason for granting unauthenticated access is to make it easier for
 non-GOV.UK editors to fact check content, avoiding the overhead of needing a
-GOV.UK signon account creation and access to all of draft GOV.UK. Evidence
-suggests that without a simple process to fact check, editors will instead
-turn to a poorer fact check experience (for example, screenshots or pasting
-content into an email).
+GOV.UK signon account and avoiding them having access to all of draft GOV.UK.
+Evidence suggests that without a simple process to fact check, editors will
+instead turn to a poorer fact check experience (for example, screenshots or
+pasting content into an email).
 
 This RFC makes a proposal to expand the unauthenticated access from a single
 piece of content to scenarios where content can span multiple pages from
 differing publishing applications. This is needed to support the fact checking
-of a step-by-step and involves using a session cookie to store an auth-bypass
+of a step-by-step where a step-by-step is both a page and embedded content
+that appears on other pages. To review a step by step thoroughly the fact
+checker should have access to all these pages and their associated assets.
+
+This proposal suggests the use of a session cookie to store an auth-bypass
 token so multiple pages and their assets can be accessed. Access to content
 will be granted by using Publishing API [link expansion][] to share an
-auth-bypass id, whereas asset access will be provided via a flag within token
-that grants access to any draft assets. These means will not allow bypassing
-access limiting protections.
+auth-bypass id, whereas asset access will be provided via a flag within a token
+that grants access to any draft assets. This token will not allow bypassing
+any access limits set up on the related content or assets.
 
 Key terminology that is used in this RFC:
 
@@ -52,27 +56,34 @@ Key terminology that is used in this RFC:
 ## Problem
 
 GOV.UK editors produce step-by-steps that need fact checking by people who
-may not have a GOV.UK signon account. As step-by-steps can span multiple
-pages on GOV.UK the process to fact check them involves not only
-reviewing the step-by-step navigation page but also reviewing the pages that
-are part of the step-by-step.
+may not have a GOV.UK signon account. A step-by-step is represented on
+GOV.UK as a [step-by-step navigation page](https://www.gov.uk/learn-to-drive-a-car)
+with pages that [represent guidance for individual steps][part-of-sbs] and
+show the progress within the step-by step. These guidance pages may be
+published by any GOV.UK application and may be drafts at the time of
+the step-by-step being fact checked. The process of fact checking a
+step-by-step involves accessing all of these pages.
 
 The current process to provide unauthenticated access to a GOV.UK draft is to
-use an auth-bypass token which allows access to a single content item. This is
-not sufficient to allow the preview of all the pages that are part of a
-step-by-step.
+use an auth-bypass token which allows access to a single content item. In
+the context of a step-by-step this can be used to allow access to the
+step-by-step navigation page but none of the related pages. This
+is not considered adequate by existing editors so is not used for the
+fact checking process.
 
-Preparing a step-by-step for fact check is currently done by someone manually
-saving each draft page to a HTML files and then wiring them up into
-a small Heroku application. This application is then shared with the
-fact checker. This is a cumbersome and time consuming process, that we believe
-can be improved.
+Instead, step-by-steps are fact checked by the use of creating and sharing a
+dedicated Heroku application for each step-by-step that needs fact checking.
+This is done by someone manually saving each draft page to a HTML file and
+then wiring all of these up into an application. This is a cumbersome and time
+consuming process, that we believe can be improved.
+
+[part-of-sbs]: https://www.gov.uk/legal-obligations-drivers-riders?step-by-step-nav=e01e924b-9c7c-4c71-8241-66a575c2f61f
 
 ## Proposal
 
 In considering how this problem can be resolved questions were raised about
 ownership of resources (content and assets) and the rights editors have to
-grant, or restrict, access on the draft stack of GOV.UK. Three publication
+grant - or restrict - access on the draft stack of GOV.UK. Three publication
 states were considered - live, draft and access limited -  and their privacy
 expectations were defined:
 
@@ -81,9 +92,9 @@ expectations were defined:
 - **Draft**: This resource is available to users with GOV.UK signon accounts,
   access is restricted from the wider internet however it is not considered
   private.
-- **Access limited**: this resource is available to users as the result
-  of implicit permissions granted at the editors discretion (for example,
-  limiting to an organisation in Whitehall) and a limited number of GOV.UK
+- **Access limited**: this resource is available to a limited set of users
+  by an option in the publishing application (for example, Whitehall allows
+  limiting content to a users organisation) and a limited number of GOV.UK
   admins. There is an expectation that this resource will be kept private
   from users who do not meet the criteria.
 
@@ -98,24 +109,28 @@ for it.
 Content that is linked as part of a step-by-step-navigation (and assets that
 are associated with that content) may be published by any publishing application
 and by any organisation. A step-by-step editor can grant a fact checker access
-to live and draft resources, however this cannot be used to provide access to
-access limited content and assets as this would break the expectations of
-privacy.
+to view live and draft resources on the draft stack, however this cannot be
+used to provide access to access limited content and assets as this would
+break the expectations of privacy.
 
 To provide access across a selection of pages an auth-bypass token would be
 stored in a session cookie this would allow maintaining this token across
 multiple requests without appending a query string (this cookie
 technique is already used for providing an [auth-bypass token to Asset
 Manager][asset-manager-auth-bypass]). To provide the auth-bypass id to
-supporting pages the Publishing API's link expansion system would be used
-for particular document and link types. As there isn't a linking system between
-assets and content it is impossible to maintain accurate links, thus an
-auth-bypass token can grant access to all draft assets for the duration of a
-session. Access limited assets must not be accessible in this circumstance
+supporting pages the Publishing API link expansion system would be used
+for particular document and link types.
+
+With access granted to pages users may experience problems viewing them
+due to lacking permission to access assets owned by those pages - this would
+be seen most visually with broken images. As there isn't a linking system
+between assets and content it is impossible to maintain accurate links, thus
+an auth-bypass token can grant access to all draft assets for the duration of
+a session. Access limited assets must not be accessible in this circumstance
 unless the asset contains an auth-bypass id that matches the token.
 
 If there is a need to preview access limited content as part of a fact check
-then individual links will have to be provided for that content.
+then individual links will have to be provided for those pages.
 
 ### Consequences
 
@@ -138,20 +153,21 @@ support similar functionality for future GOV.UK content.
   authentication and auth-bypass token when considering access to resources.
 - Content Store will be updated to allow an auth-bypass id to be present in the
   links of a content item, this will be checked against a user provided
-  auth-bypass_id for content items that are not access limited.
-- Publishing API will be changed to support the sharing of auth-bypass ids for an
-  edition through link expansion rules.
+  auth-bypass id for content items that are not access limited.
+- Publishing API will be changed to support the sharing of auth-bypass ids for
+  an edition through link expansion rules.
 - Publishing API and Content Store will have their access limit code decoupled
   from auth-bypass ids. This will be done to remove a frequent source of
   confusion that conflates the separate processes of access limiting and
   auth-bypass.
 - Collections Publisher will be changed to include a flag in auth-bypass tokens
-  that provide draft asset access. To reflect the increased access of a token
-  they will be given an expiry time (anticipated to be 1 month). Collections
-  Publisher will also be updated to include origin information in the token
-  (such as creator id and content id) to allow a token to be audited. The
-  [auth-bypass token documentation][auth-bypass-token-docs] will be updated to
-  reflect this as the best practice for future auth-bypass tokens
+  that provide access to all draft assets that are not access limited. To
+  reflect the increased access of a token they will be given an expiry time
+  (anticipated to be 1 month). Collections Publisher will also be updated to
+  include origin information in the token (such as creator id and content id)
+  to allow a token to be audited. The [auth-bypass token
+  documentation][auth-bypass-token-docs] will be updated to
+  reflect this as the best practice for future auth-bypass tokens.
 
 ### Rejected alternative approaches
 
@@ -189,7 +205,7 @@ whenever it occurs.
 
 #### Providing a fact checker access to all resources on the draft stack that are not access limited
 
-An approach that was considered was to allow the fact checker to have access
+An approach that was considered was to allow a fact checker to have access
 to all pages on the draft stack that are not access limited. This would grant
 them similar access to an authenticated user. This offered the advantage that
 it allows simplifying the logic involved in checking tokens and is consistent
@@ -199,7 +215,26 @@ This was rejected since it was a significant departure from the existing access
 allowed to a fact checker through an auth-bypass token and may require
 considerable communication and consultation time. It remains a viable option
 but should be considered as part of a consideration of whether there should be
-more accountability in draft access.
+more accountability and/or simplicity in draft access.
+
+#### Determining which assets a page has to set auth-bypass ids on them
+
+An ideal scenario for asset access would be for it to mirror the system used
+for content access as discrepancies between them are a source of complication
+and future confusion. To achieve this it was pondered whether there were ways
+to share auth-bypass ids with assets in a similar way to how link expansion can
+share them between linked content.
+
+An approach to achieve this would be to have Publishing API check content
+for asset presence at the point of sending it to the draft content store and
+then updating the assets with the relevant auth-bypass ids. Doing this would
+require Publishing API to have a dependency on Asset Manager and to have the
+means to identify and manage assets.
+
+This approach was rejected due to it risking adding substantial complexity
+to Publishing API in order for it to have the means to identify and manage
+assets. It was also considered undesirable to couple Asset Manager and
+Publishing API since they are currently able to operate distinctly.
 
 #### Automating the generation of a Heroku preview with a GOV.UK signon account
 
--- a/rfc-113-expanding-draft-access-for-unauthenticated-users.md
+++ b/rfc-113-expanding-draft-access-for-unauthenticated-users.md
@@ -1,0 +1,212 @@
+# Expanding draft access for unauthenticated users to allow multi-page fact checks
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+
+To support this functionality a number of changes are required in the GOV.UK
+stack, these are intended to be non-specific to the step-by-step problem and to
+support similar functionality for future GOV.UK content.
+
+- Authenticating Proxy will be changed to recognise auth-bypass tokens that are
+  present in cookies.
+- Asset Manager will be changed to recognise the presence of a flag in an
+  auth-bypass token that will grant an unauthenticated user access to a draft
+  asset that is not access limited. Fact checkers are not expected to be able
```
ChrisBAshton: Is this a flag in the payload part of the JWT token? Just checking my understanding - hadn't heard of JWT before today (NB, have raised https://github.com/alphagov/authenticating-proxy/pull/158 as a result!).

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+  to exploit this access given sensitive assets should be access limited and
+  the URLs of assets are relatively unpredictable.
+- Currently Authenticating Proxy and Asset Manager do not check whether an
+  authenticated user has access to a resource if an auth-bypass token is
+  present. This introduces potential confusion when this token is provided by
+  a session cookie and could prevent an authenticated user viewing other parts
+  of draft GOV.UK. To counter this both apps will be changed to consider both
+  authentication and auth-bypass token when considering access to resources.
```
ChrisBAshton: In cases where both are supplied, which will take precedent?

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+  auth-bypass_id for content items that are not access limited.
+- Publishing API will be changed to support the sharing of auth-bypass ids for an
+  edition through link expansion rules.
+- Publishing API and Content Store will have their access limit code decoupled
+  from auth-bypass ids. This will be done to remove a frequent source of
+  confusion that conflates the separate processes of access limiting and
+  auth-bypass.
+- Collections Publisher will be changed to include a flag in auth-bypass tokens
+  that provide draft asset access. To reflect the increased access of a token
+  they will be given an expiry time (anticipated to be 1 month). Collections
```
ChrisBAshton: Where will that expiry happen? Is it when [creating the token](https://github.com/alphagov/authenticating-proxy/blob/56939ebb83481076811edcadff60f3a086d2961e/README.md#generating-a-auth-bypass-token)?

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+  admins. There is an expectation that this resource will be kept private
+  from users who do not meet the criteria.
+
+These were then used to form conclusions on the level of access a
+step-by-step editor should be able to grant to a fact checker.
+
+The editor should be able to grant access to a fact checker to view the
+draft step-by-step navigation page via an auth-bypass token, this should allow
+the fact checker to bypass access limits on the content and any assets created
+for it.
```
barrucadu: Should this be "this **must not** allow the fact checker to bypass access limits on the content"?  Later you say

> If there is a need to preview access limited content as part of a fact check
> then individual links will have to be provided for that content.

Or is "access limits" here being used to mean "things which are draft", and not "things which are access limited"?

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+considerable communication and consultation time. It remains a viable option
+but should be considered as part of a consideration of whether there should be
+more accountability in draft access.
+
+#### Automating the generation of a Heroku preview with a GOV.UK signon account
+
+This would involve using a special GOV.UK signon account to download the pages
+of the step-by-step and then create a small app that can be hosted on Heroku.
+This was rejected due to having a high level of complexity and being
+unconventional compared to other GOV.UK apps.
```
brucebolt: There's also a cost involved in hosting Heroku apps, and we sometimes reach the limit of our current subscription (which may be the free tier).

rfc-113-expanding-draft-access-for-unauthenticated-users.md
kevindew: Yes it'd be a flag in the token. My spike into it looked like this which probably explains a bit: 

https://github.com/alphagov/collections-publisher/blob/ebf9c7c4d4a9fe1378e9af61cb589c3f3cf6a281/app/lib/jwt_helper.rb#L8-L12

Nice one - opening that PR

rfc-113-expanding-draft-access-for-unauthenticated-users.md
kevindew: I don't believe there's a case for precedence so both can apply independently.

This is how I anticipate the logic to be in content store which takes both into account:

```
if we have a token and it's contains the root auth_bypass_id of the content
  show content
else if we're authenticated and one of the users / organisations in the access limit
  show content
else if the content doesn't have access limit and has links which include the auth_bypass_id that matches the users auth_bypass_id
  show content
else if user doesn't have a auth_bypass_id
  show content (we're assuming this is the live stack)
else
  forbidden
end
```

rfc-113-expanding-draft-access-for-unauthenticated-users.md
kevindew: Yup - my spike demonstrates that somewhat

https://github.com/alphagov/collections-publisher/blob/ebf9c7c4d4a9fe1378e9af61cb589c3f3cf6a281/app/lib/jwt_helper.rb#L8-L12

rfc-113-expanding-draft-access-for-unauthenticated-users.md
kevindew: As the step-by-step navigation page is the document that the editor owns (in constrast to the linked pages which the editor does not own) an auth-bypass token provides access to that page whether or not there are any access limits for the content. What the editor cannot do is provide a token that can bypass access limits on any content they do not own.

kevindew:
Amazing - thanks for commenting so quickly @bilbof, @ChrisBAshton, @barrucadu and @brucebolt 

> This sounds good to me. Thank you for a detailed write up!
> 
> * Will the linking behaviour (granting access to all linked content in one go) be restricted to step-by-steps?

Yes - it'll be configured based on document type and link type. 

This is how I spiked it

https://github.com/alphagov/publishing-api/blob/30f01b956d51cdaf347302540337e83161079402/lib/expansion_rules.rb#L69

and 

https://github.com/alphagov/publishing-api/blob/30f01b956d51cdaf347302540337e83161079402/lib/expansion_rules.rb#L116-L121

> * Is there a way for users to manually invalidate auth-bypass tokens, since they'll be giving access to more content?

Assuming you mean editor when referring to user here. They would likely have not any ability to revoke a token without developer intervention.

With developer intervention access to content could be revoked by changing the auth_bypass_id for a piece of content. Ideas to use a [random string of bytes](https://github.com/alphagov/collections-publisher/blob/d95a9def0735d1885f93fbf67527af1cacb55c06/app/models/step_by_step_page.rb#L32) rather than a [deterministic auth_bypass_id](https://github.com/alphagov/collections-publisher/blob/1cb963d2b38aa7ea58e108da77ab33a3956f3e08/app/models/step_by_step_page.rb#L130-L143) help this.

To restrict access to draft assets would be harder as the token is not verified against a content store entry. Thus the only way to revoke one of these would be to go nuclear and change the JWT_SECRET which would invalidate every single token on GOV.UK. An approach to counter this would be to either have asset-manager talk to content-store when validating tokens or for it to store it's own database of ids it accepts/rejects. I consider both of these to be beyond the scope of this work.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
barrucadu: Ok, that makes sense.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+access limited content and assets as this would break the expectations of
+privacy.
+
+To provide access across a selection of pages an auth-bypass token would be
+stored in a session cookie this would allow maintaining this token across
+multiple requests without appending a query string (this cookie
+technique is already used for providing an [auth-bypass token to Asset
+Manager][asset-manager-auth-bypass]). To provide the auth-bypass id to
+supporting pages the Publishing API's link expansion system would be used
+for particular document and link types. As there isn't a linking system between
```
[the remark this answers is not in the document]
kevindew: > This is the first mention of assets. Need to recollect that currently asset auth bypass is done in a 'cascading' manner (at least, in Content Publisher) from the parent document

Yes, in publishing applications you tend to have assets that are created by the same publishing application for just that document so they're symbolically owned by that resource. Content Publisher probably goes the furthest in that regard by the arrays of assets for govspeak compilation. Though no apps outside of Content Publisher know about the relation of those assets to the Content Publisher document.

> Just a thought, and probably out-of-scope here, but are we considering this feature in the context of publications, which also have multiple 'levels' of content?

Yeah pondered that briefly however as you'd expect to own all of that content (as we don't have the concept of a shared HTML attachment) you'd want that all the content to have the same `auth_bypass_id` so all access limiting restrictions can be bypassed.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+step-by-step editor should be able to grant to a fact checker.
+
+The editor should be able to grant access to a fact checker to view the
+draft step-by-step navigation page via an auth-bypass token, this should allow
+the fact checker to bypass access limits on the content and any assets created
+for it.
+
+Content that is linked as part of a step-by-step-navigation (and assets that
+are associated with that content) may be published by any publishing application
+and by any organisation. A step-by-step editor can grant a fact checker access
```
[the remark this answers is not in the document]
kevindew: > Is that necessary, or just implicit?

Yes it's necessary, if some of the documents are live the fact checker would still be browsing them on the draft stack so would need access granting.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+to live and draft resources, however this cannot be used to provide access to
```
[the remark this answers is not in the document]
kevindew: Resources is used refer to both content and assets.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+- **auth-bypass id:** A string of data that is used to match whether a
+  user is granted access to a particular resource without authentication,
+  this is not shown to end users.
+- **auth-bypass token:** A [JSON Web Token](https://jwt.io/) that is used to
+  encrypt an auth-bypass id and is available to end users.
+- **access limited:** A term used to denote draft GOV.UK resources that are
+  only accessible by particular users, users in a particular organisation or
+  by users that have a token to access them.
+- **step-by-step:** A particular document type on GOV.UK which is
+  both a page in it's own right and a collection of pages that comprise
```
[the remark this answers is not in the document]
kevindew: The navigation is the [step by step page](https://www.gov.uk/learn-to-drive-a-car) itself. Whereas the other pages are defined with the somewhat loose terms of "part of" and "related to"

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+#### Providing a means to embed the step-by-step component into live content
+
+This would involve a means of injecting the step-by-step navigation component
+into live GOV.UK pages via JavaScript. This would prevent a fact checker of a
+step-by-step from viewing any draft content except the step-by-step
+navigation itself. This was quickly rejected as it was considered to be a
+common scenario for the pages that are part of a step-by-step
+to also be drafts.
+
+#### Exclude asset access from fact-checking
```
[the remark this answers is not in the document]
kevindew: Yeah - this is a realistic proposal. This is how GOV.UK access limiting used to work before Content Publisher where assets just weren't part of the solution (I don't know if that was through choice or because they were overlooked), and this was very much how this was nearly done until there was a "oh crap, what about assets?".

For the most part this would work all until someone adds a draft with an asset - which could be a long time away.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+
+#### Providing a fact checker access to all resources on the draft stack that are not access limited
+
+An approach that was considered was to allow the fact checker to have access
+to all pages on the draft stack that are not access limited. This would grant
+them similar access to an authenticated user. This offered the advantage that
+it allows simplifying the logic involved in checking tokens and is consistent
+with privacy expectations.
+
+This was rejected since it was a significant departure from the existing access
```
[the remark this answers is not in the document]
kevindew: > This is interesting, since we seem to be taking this approach with assets. Perhaps that's less of a concern, since the primary route to an asset would be through a document i.e. unrelated assets are effectively hidden if we scope draft access to only the content of the step-by-step. 

Yes, the route to view an asset would be through having to know what both the id and the filename were for that asset. Whereas for content the route would be to browse the site and/or guess slugs (but still be restricted by search not existing for draft content). 

In practice the proposed change of this RFC would be very hard for any fact-checkers to be aware that there is a difference based on this flag as viewing content would be no different and for applications that don't set the flag there'd be no change. Whereas for this rejected change it'd be very easy for fact checkers to experience a difference in behaviour by just clicking a link. 

I actually think that it should be ok for fact-checkers to view the whole draft stack and I imagine if GOV.UK increases it's journey into embedding content from multiple places (as expected with more embracing of structured content) we'll have more questions to resolve on how that is fact checked and what the meaning of access limiting is. However that is quite a big change and could be contentious/surprising so, given the step-by-step team is to exist for only another couple of weeks, it is not a conversation I want to start unnecessarily.

> Can we address/discuss this same issue for assets as part of the proposal?

It's referenced in the consequences section.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+suggests that without a simple process to fact check, editors will instead
+turn to a poorer fact check experience (for example, screenshots or pasting
+content into an email).
+
+This RFC makes a proposal to expand the unauthenticated access from a single
+piece of content to scenarios where content can span multiple pages from
+differing publishing applications. This is needed to support the fact checking
+of a step-by-step and involves using a session cookie to store an auth-bypass
+token so multiple pages and their assets can be accessed. Access to content
+will be granted by using Publishing API [link expansion][] to share an
```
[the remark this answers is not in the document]
kevindew: It ends up being both dependency resolution and link expansion. Dependency resolution determines that those pages need to be updated and link expansion is used to work out the data in those links.

The `auth_bypass_id` would be in the links of the content like so:

```
{
  "content_id": "12345-1234-1234-1231123",
  "auth_bypass_ids": ["dfghjklkjhgf4567897654"],
  "access_limit": { ... },
  "expanded_links": {
    "pages_part_of_step_nav": [
      {
         "api_path": "/api/content/apply-first-provisional-driving-licence",
         ...
         "auth_bypass_ids": ["1234675sdfgsdfghsdfgh765yrtfg"],
      }
  }
```

Logic in content store would first check the content items own `auth_bypass_ids` as they can bypass access limits. Then it would check the presence of `auth_bypass_ids` in the links.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+anticipated and likely having to fall to a support team to resolve or explain
+whenever it occurs.
+
+#### Providing a fact checker access to all resources on the draft stack that are not access limited
+
+An approach that was considered was to allow the fact checker to have access
+to all pages on the draft stack that are not access limited. This would grant
+them similar access to an authenticated user. This offered the advantage that
+it allows simplifying the logic involved in checking tokens and is consistent
+with privacy expectations.
```
ChrisBAshton: There's another approach similar to this that I can't see referenced in the RFC, so thought I'd raise just for the record. It would be:

* If step by step page has a valid token in its URL...
* ...then append tokens to each of the links to other pages

The advantage would be that we're not granting blanket access to the entire draft stack, only to the pages linked to from the step by step. The disadvantages are twofold: 1) it would complicate the renderer, and 2) we don't know what each linked page's token should be.

The latter would probably require a change to how tokens are generated. My understanding is that it is up to the application to create the token and to tell publishing-api what the token is, but if we made it the responsibility of publishing-api to generate its own tokens, we could retrieve the token for each page by querying its API. Admittedly this would mean any application on GOV.UK could query the token of content published by other applications, which is an area of concern.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -0,0 +1,212 @@
...
+common scenario for the pages that are part of a step-by-step
+to also be drafts.
+
+#### Exclude asset access from fact-checking
+
+Assets are the area of access that provide the most complication since they
+are distinct from the Publishing API link system. An option was to exclude
+them from the access checks and have them return forbidden errors if a
+step-by-step page used them - this would typically be most noticeable on an
+embedded image.
```
ChrisBAshton: Probably another rejection worth documenting - couldn't we write a task which goes through all content items, looks for asset URLs, retrieves the associated UIDs of the content items and saves those as links in publishing API?

That would then free up possibilities for providing tokens for assets via link expansion.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
kevindew: @ChrisBAshton Thanks for the suggestion - I know we chatted IRL so I'll just summarise a bit here.

> There's another approach similar to this that I can't see referenced in the RFC, so thought I'd raise just for the record. It would be:
> 
> * If step by step page has a valid token in its URL...
> * ...then append tokens to each of the links to other pages
> 
> The advantage would be that we're not granting blanket access to the entire draft stack, only to the pages linked to from the step by step. The disadvantages are twofold: 1) it would complicate the renderer, and 2) we don't know what each linked page's token should be.

The token itself doesn't provide blanket access to the draft stack - just specific pages and most draft assets. 

As I understand this idea seems to be more about the means to distribute the token rather than the access level. 

To do this we'd either need to adjust the rendering applications involved to know about the potential for tokens and then be able to append them to URLs, which could involve a lot of draft specific changes. This is problematic since we purposefully keep apps distinct from knowing whether they are not the draft stack or not as much as possible.

A thing we could do is change authenticating proxy to rewrite the HTML of pages that are returned so all links have the token on them, which would effectively blanket share the token to all links that matched a pattern (like an old school session id). Though this would greatly increase the code of authenticating proxy and carry with it a number of challenges resulting from the HTTP rewriting. Although on the plus side every URL would be a sharable.

> The latter would probably require a change to how tokens are generated. My understanding is that it is up to the application to create the token and to tell publishing-api what the token is, but if we made it the responsibility of publishing-api to generate its own tokens, we could retrieve the token for each page by querying its API. Admittedly this would mean any application on GOV.UK could query the token of content published by other applications, which is an area of concern.

Applications still need to know the auth_bypass_id so they can share the token with a user. So if Publishing API did create them Collections Publisher would need to know it to share it. 

Retrieving the token at the point of rendering the page isn't ideal for a few reasons (performance being a key one) but one thing we'd want to avoid is that by sharing different pages tokens that could be actually used to grant users access to content that is access limited. For example say I have document "B" that is access limited to just me and you have a step by step that links to "B". You shouldn't be able to get the auth_bypass_id I created for B since you don't have access.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
kevindew: > Probably another rejection worth documenting - couldn't we write a task which goes through all content items, looks for asset URLs, retrieves the associated UIDs of the content items and saves those as links in publishing API?
> 
> That would then free up possibilities for providing tokens for assets via link expansion.

We don't have a concept of links to assets unfortunately - that'd have resolved the need for the draft_asset_manager_access grant. I like to imagine that at some point in the future we actually merge Asset Manager and Publishing API together since they're very similar concepts for providing draft and live resources on GOV.UK and their subtle differences are points of confusion.

With how things currently stand though app wise, the way we could give each asset a token would be to do the following at the point of sending a draft from Publishing API to Content Store:

- Determine whether any links of the edition provide auth_bypass_ids
- Parse all fields that could contain URLs in the Publishing API (this might be as grim as turning details JSON into a  string and pattern matching for URLs) and pick out any that are to assets
- Query asset manager for each of these to see if they are draft
- For all the draft ones send an update request to the asset to set the auth_bypass_id

This is not an unworkable approach but it is undesirable as it'll add a new Asset Manager dependency on Publishing API and add extra time and complexity to the downstream process.

kevindew:
Thanks for all the comments everyone. I've made some clarification amends following comments.

rfc-113-expanding-draft-access-for-unauthenticated-users.md
```diff
@@ -199,7 +215,26 @@ This was rejected since it was a significant departure from the existing access
 allowed to a fact checker through an auth-bypass token and may require
 considerable communication and consultation time. It remains a viable option
 but should be considered as part of a consideration of whether there should be
-more accountability in draft access.
+more accountability and/or simplicity in draft access.
+
+#### Determining which assets a page has to set auth-bypass ids on them
+
+An ideal scenario for asset access would be for it the mirror the system used
```
erino: I think there's a missing word here "would be for it the mirror the"

kevindew: d'oh - thanks sorted that out.

PR #114, opened 2019-12-16
RFC 114: Adjust SRI implementation to allow for HTTP/2

Nooshu:
TL;DR; the way we have implemented Subresource Integrity (SRI) is having an impact on HTTP/2 performance. Adjusting this should allow us to enable it and show have a positive impact on our users web performance. 

👉 [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/3a682960071e7f946d9ecef843d82db4c938f86a/rfc-114-sri-enable-http2.md)


Nooshu:
A more detailed write up of the issue can be [found here](https://nooshu.github.io/blog/2019/12/17/http2-and-sri-dont-always-get-on/)

kevindew:
This is really interesting @Nooshu - I enjoyed reading your blog on it.

Taking a little bit of a tangent from the RFC proposal my eye was particularly caught by:

>And lastly, the final option could be to remove the assets domain (shard) completely, and serve all assets from the main domain. This would be the ideal solution, as it removes any need for the flakey HTTP/2 connection coalescing. But due to the complexity of the GOV.UK website, I don’t see this happening anytime soon.

How much would we be likely to benefit from doing this switch for the switch to HTTP/2 (and maybe HTTP/3) compared to what is proposed in the RFC?

The reason I ask is that while it wouldn't be a walk in the park to swap asset serving to use the www.gov.uk domain it is, I expect, less painful than you might anticipate. If doing this is a significantly better option that what is proposed in the RFC I'd be happy to spend an hour with you to doing a quick spike into if that is realistic. I don't personally anticipate it to involve a significantly greater number of changes than would be needed from this RFC (though they could be famous last words).

Nooshu:
Thanks @kevindew! I think that really comes down to why the assets domain exists in the first place? I'm unsure on the decision behind it originally. If it was purely for the HTTP/1.1 optimization of parallel download of assets rather than for other reasons (like security), then I believe it would well be worth the time deprecating it and serving all assets from a single domain. 

As mentioned in the blog post HTTP/2 coalescing is [notoriously flakey](https://daniel.haxx.se/blog/2016/08/18/http2-connection-coalescing/) and often depends on an individual browsers implementation. Certain browsers don't even support it e.g. Safari versions earlier than 12, older versions of Edge. So that's another consideration. Removal of the assets domain will remove the dependency on coalescing.

At the moment we don't have stats on how many of our users use a browser that support HTTP/2 (since it isn't enabled), but I'm sure it will be very high, since global support is [pushing 95%+](https://caniuse.com/#feat=http2). The reason this is important is the removal of the assets domain may have a slight performance impact on users who are still on HTTP/1.1. The most likely reason for this will be either an [older browser](https://caniuse.com/#feat=http2), or they are connecting through a proxy that doesn't support HTTP/2 and forces the connection to be downgraded. I feel this performance impact will minimal in any case, due to the type of content we serve on GOV.UK, so I don't see it as a blocker.

Personally I think we could have a quick win with the SRI change from `anonymous` to `use-credentials` to allow browsers that support coalescing to do it properly (and also support HTTP/1.1 users at the same time). Then in the background have a bigger piece of work to remove the assets domain completely (assuming it's only use was for HTTP/1.1 optimisation). But I'm happy to follow your lead on that as you have much more context on the architecture.

Domain sharding isn't coming back in HTTP/3, as HTTP/3 fixes known issues with HTTP/2, and introduces QUIC (Quick UDP Internet Connection), TLS 1.3, and 0-RTT (Fastly already offer the latter two). [HTTP/3](https://caniuse.com/#feat=http3) has no support at the moment as it currently sits behind browser flags. At some point in the future it will be supported and the sharding removal will need to be made anyway to keep up with the latest protocol developments.

Nooshu:
Looking into some of my previous research related to users using browsers that support HTTP/2 from October 2018:

> Examining 1 months worth of user data from Google Analytics for GOV.UK, we can see that the current browser support is 98.8%. Users with an incompatible browser: Chrome: 236,729 | IE: 66,872 | Firefox: 8,959 | iOS Safari: 97,636 | Safari: 29,137 | Android Webview: 102,751 | Total users in October 2018: 45,949,539.

It has been 14 months since I last checked and I can only see this increasing over time as users update their device / OS / Browser. So with 99%+ support for HTTP/2, the removal of the sharding domain will have a minimal impact on users

kevindew:
> Thanks @kevindew! I think that really comes down to why the assets domain exists in the first place? I'm unsure on the decision behind it originally. If it was purely for the HTTP/1.1 optimization of parallel download of assets rather than for other reasons (like security), then I believe it would well be worth the time deprecating it and serving all assets from a single domain.

That's my expectation too. At origin both www. and assets. are served from the same machine and just on a different vhost. The host names are different entries in Fastly CDN but I don't think that is much of a concern. 

> As mentioned in the blog post HTTP/2 coalescing is [notoriously flakey](https://daniel.haxx.se/blog/2016/08/18/http2-connection-coalescing/) and often depends on an individual browsers implementation. Certain browsers don't even support it e.g. Safari versions earlier than 12, older versions of Edge. So that's another consideration. Removal of the assets domain will remove the dependency on coalescing and allow all browser that support HTTP/2 to take advantage of the performance optimisations.
> 
> At the moment we don't have stats on how many of our users use a browser that support HTTP/2 (as it isn't enabled), but I'm sure it will be very high, since global support is [pushing 95%+](https://caniuse.com/#feat=http2). The reason this is important is the removal of the assets domain may have a slight performance impact on users who are still on HTTP/1.1. The most likely reason for this will be either an [older browser](https://caniuse.com/#feat=http2), or they are connecting through a proxy that doesn't support HTTP/2 and forces the connection to be downgraded. I feel this performance impact will minimal in any case, due to the type of content we serve on GOV.UK, so I don't see it as a blocker.
> 
> Personally I think we could have a quick win with the SRI change from `anonymous` to `use-credentials` to allow browsers that support coalescing to do it properly (and also support HTTP/1.1 users at the same time). Then in the background have a bigger piece of work to remove the assets domain completely (assuming its only use was for HTTP/1.1 optimisation). But I'm happy to follow your lead on that as you have much more context on the architecture.

Sure, sounds a good idea if it's a relatively quick win. I anticipated it being lots of small changes to a lot of apps so wanted to offer an alternative if that costs us a fair amount of time and doesn't get us to the right destination.

> Domain sharding isn't coming back in HTTP/3, as HTTP/3 fixes known issues with HTTP/2, and introduces QUIC (Quick UDP Internet Connection), TLS 1.3, and 0-RTT (Fastly already offer the latter two). [HTTP/3](https://caniuse.com/#feat=http3) has no support at the moment as it currently sits behind browser flags. At some point in the future it will be supported and the sharding removal will need to be made anyway to keep up with the latest protocol developments.

Thanks - It's good to learn more about this stuff 👍 

---

The sort of things I think we'd need for switching the domains the CSS and JS are served from would be:

Add proxy passes to the www nginx file, similar to https://github.com/alphagov/govuk-puppet/blob/226ed02b70932650b4fc478d2971e33e77b9fd45/hieradata/common.yaml#L1388-L1402 and probably prefix them all with `/assets`.

Then in a frontend app we'd need to tell them to not use a special asset host and change the asset path to `/assets/:app-name`.

Then the slightly tricky point would be the static app that'd need to explicitly be told to use the web root as the asset host.

Those steps would probably get you the whole way to the goal but would leave behind some cleaning up tasks for static hostname and asset-managers dual meaning of using asset-root. There's probably a couple of other snags.

Nooshu:
Opened a separate [RFC (115)](https://github.com/alphagov/govuk-rfcs/pull/115) for changing serving of static assets from the `assets` domain to `www.gov.uk`.

Nooshu:
Closing this RFC, agreed with @kevindew that this is fine to proceed with. FYI this is a change that needs to be made for the work on [RFC 115](https://github.com/alphagov/govuk-rfcs/pull/115) to proceed. Separate PR's for the apps can now be raised.

PR #116, opened 2020-01-03
RFC 116: store attachment data in content items

barrucadu:
**Deadline for discussion: 2020-01-31**
~**Deadline for discussion: 2020-01-24**~

[Rendered](https://github.com/alphagov/govuk-rfcs/blob/msw/rfc-116/rfc-116-store-attachment-data-in-content-items.md)

### DIFF
--- a/rfc-116-store-attachment-data-in-content-items.md
+++ b/rfc-116-store-attachment-data-in-content-items.md
@@ -1,0 +1,393 @@
+# Store attachment data in content items
+
+## Summary
+
+Add a new field to the details hash of content items, `attachments`,
+which has metadata about the document's attachments (if any).
+
+
+## Problem
+
+Documents on GOV.UK can have attachments, which come in a few
+different types.  Whitehall, which has the richest model of
+attachments, has:
+
+- External attachments (links to other websites)
+- HTML attachments
+- File attachments (which can have previews)
+
+When thinking about how attachments are displayed to the user, there
+are two sorts:
+
+- *Document-level attachments*, such on Whitehall publication and
+  consultation documents, which are displayed separately to the main
+  content.
+
+- *Inline attachments*, such as in other Whitehall documents, manual
+  sections, specialist documents, and travel advice documents, which
+  are referenced in the page content and rendered as links.
+
+Publishing applications vary in how easily accessible they make
+attachment metadata in their content items.  Whitehall only sends
+rendered HTML to the Publishing API.  Other publishing apps send a
+combination of rendered HTML and metadata.
+
+This inconistency restricts what we can do with attachments,
+particularly for Whitehall content.  For example, we cannot generate
+comprehensive [schema.org][] metadata for attachments: [see this
+comment][].
+
+Additionally, users of the content API cannot make use of attachments
+without parsing the document body.  This inhibits creative use of our
+content, and makes life more difficult for users such as the National
+Archives, who need to be able to record metadata like the HoC paper
+number of official publications.
+
+We should make metadata for Whitehall attachments available in the
+same way as other publishing applications, but that requires some
+changes into how we represent attachments across the stack.
+
+
+## Proposal
+
+In our content schemas we already have an [`asset_link` type][], which
+is used for attachments in specialist documents and manual sections.
+It's also used in travel advice pages for the page image and a single
+attachment.  This type appears to solve all of our problems:
+[Publishing API can render such attachments with govspeak][], and
+content items can expose this data to enable programmatic use.
+
+However, the metadata Govspeak expects and the `asset_link` type are
+only *informally* the same.  And furthermore, Whitehall has much richer
+attachment metadata than either `asset_link` or Govspeak allow.
+
+Given that `asset_link` is currently used for both images and
+attachments, adding all of the missing metadata to `asset_link` feels
+a mistake: images and attachments have different metadata, so we
+shouldn't conflate the two.
+
+I propose we do this:
+
+1. Split `asset_link` into three new types: `file_attachment_asset`,
+   `specialist_publisher_attachment_asset` and `image_asset`, updating
+   the schemas which use `asset_link` accordingly, with these fields:
+
+   - `file_attachment_asset`:
+     - `url` (required)
+     - `content_type` (required)
+     - `title`
+   - `specialist_publisher_attachment_asset` (extends `file_attachment_asset`):
+     - `content_id` (required)
+     - `created_at`
+     - `updated_at`
+   - `image_asset`:
+     - `url` (required)
+     - `content_type` (required)
+
+   **Migration concerns:** This would need to be done in three steps:
+   1. Update manuals-publisher to not set the `created_at` and `updated_at` for attachments.
+   2. Republish all manual sections.
+   3. Update the schemas.
+
+1. Add new optional fields for extra metadata:
+
+   - `file_attachment_asset`:
+     - `accessible`
+     - `alternative_format_contact_email`
+     - `attachment_type` (enum with permitted values `"file"`)
+     - `file_size`
+     - `filename`
+     - `id`
+     - `locale`
+     - `number_of_pages`
+     - `preview_url`
+   - `image_asset`
+     - `alt_text`
+     - `caption`
+     - `credit`
+
+   **Migration concerns:** This can be done without breaking
+   compatibility, as all new fields are optional.
+
+1. Add a new attachment type for publication attachments:
+
+    `publication_attachment_asset`:
+    - `command_paper_number`
+    - `hoc_paper_number`
+    - `isbn`
+    - `parliamentary_session`
+    - `unique_reference`
+    - `unnumbered_command_paper`
+    - `unnumbered_hoc_paper`
+    - plus a one-of combination of:
+      - the properties of `file_attachment_asset`
+      - or,
+        - `attachment_type` (required; enum with permitted values `"html"`)
+        - `id`
+        - `locale`
+        - `title`
+        - `url` (required)
+      - or,
+        - `attachment_type` (required; enum with permitted values `"external"`)
+        - `id`
+        - `locale`
+        - `title`
+        - `url` (required)
+
+   I've left out the following fields because they can be deduced from
+   others: `csv?`, `external?`, `external_url`, `file_extension`, `html?`,
+   `opendocument?`, `pdf?`
+
+   And these fields because they don't seem to be used:
+   `print_meta_data_contact_address`, `web_isbn`.
+
+   And these fields because the Publishing Workflow team are planning
+   to remove them from Whitehall: `order_url`, `price_in_pence`.
+
+   And these fields because they don't need to be exposed outside the
+   publishing app: `attachable_id`, `attachment_data_id`, `deleted`,
+   `ordering`, `slug`.
+
+   **Migration concerns:** This can be done without breaking
+   compatibility, as the new type will be unused for now.
+
+1. Make `attachment_type` and `id` in `file_attachment_asset` mandatory.
+
+    **Migration concerns:** this would need to be done in three steps:
+    1. Update specialist publisher, travel advice publisher, and
+       manuals publisher to set the fields.
+    2. Republish all affected documents.
+    3. Make the field mandatory.
+
+1. Add a *mandatory* `details.attachments` field to all formats which
+   can have attachments.
+
+    **Migration concerns:** this would need to be done in four steps:
+    1. Add the field as optional.
+    2. Update publishing apps to write to it.
+    3. Republish all affected documents.
+    4. Make the field mandatory.
+
+1. Add a *mandatory* `details.featured_attachments` list to
+   publication formats, which is an ordered list of govspeak or HTML
+   strings (using the `#/definitions/multiple_content_types` schema)
+   rendering attachments from `details.attachments`.
+
+   **Migration concerns:** this would need to be done in four steps:
+   1. Add the field as optional.
+   2. Update publishing apps (just Whitehall and Content Publisher) to write to it.
+   3. Republish all affected documents.
+   4. Make the field mandatory.
+
+1. Remove the `details.documents` field.
+
+   **Migration concerns:** this would need to be done in four steps:
+   1. Update frontend apps to render from
+      `details.featured_attachments`.
+   2. Update publishing apps to not set `details.documents`.
+   3. Republish all affected documents.
+   4. Remove the field.
+
+The final schemas will be:
+
+```
+local FileAttachmentAssetProperties = {
+  accessible: { type: "boolean", },
+  alternative_format_contact_email: { type: "string", },
+  attachment_type: { type: "string", enum: ["file"], },
+  content_type: { type: "string", },
+  file_size: { type: "integer", },
+  filename: { type: "string", },
+  id: { type: "string" },
+  locale: { "$ref": "#/definitions/locale", },
+  number_of_pages: { type: "integer", },
+  preview_url: { type: "string", format: "uri", },
+  title: { type: "string", },
+  url: { type: "string", format: "uri", },
+};
+
+local HtmlAttachmentAssetProperties = {
+  attachment_type: { type: "string", enum: ["html"], },
+  id: { type: "string" },
+  locale: { "$ref": "#/definitions/locale", },
+  title: { type: "string", },
+  url: { type: "string", format: "uri", },
+};
+
+local ExternalAttachmentAssetProperties = {
+  attachment_type: { type: "string", enum: ["external"], },
+  id: { type: "string" },
+  locale: { "$ref": "#/definitions/locale", },
+  title: { type: "string", },
+  url: { type: "string", format: "uri", },
+};
+
+local PublicationAttachmentAssetProperties = {
+  command_paper_number: { type: "string", },
+  hoc_paper_number: { type: "string", },
+  isbn: { type: "string", },
+  parliamentary_session: { type: "string", },
+  unique_reference: { type: "string", },
+  unnumbered_command_paper: { type: "boolean", },
+  unnumbered_hoc_paper: { type: "boolean", },
+};
+
+{
+  image_asset: {
+    type: "object",
+    additionalProperties: false,
+    required: [
+      "content_type",
+      "url",
+    ],
+    properties: {
+      alt_text: { type: "string", },
+      caption: { type: "string", },
+      content_type: { type: "string", },
+      credit: { type: "string", },
+      url: { type: "string", format: "uri", },
+    },
+  },
+
+  file_attachment_asset: {
+    type: "object",
+    additionalProperties: false,
+    required: [
+      "attachment_type",
+      "content_type",
+      "url",
+    ],
+    properties: FileAttachmentAssetProperties,
+  },
+
+  specialist_publisher_attachment_asset: {
+    type: "object",
+    additionalProperties: false,
+    required: [
+      "attachment_type",
+      "content_id",
+      "content_type",
+      "url",
+    ],
+    properties: FileAttachmentAssetProperties + {
+      content_id: { "$ref": "#/definitions/guid", },
+      created_at: { format: "date-time", },
+      updated_at: { format: "date-time", },
+    },
+  },
+
+  publication_attachment_asset: {
+    oneOf: [
+      {
+        type: "object",
+        additionalProperties: false,
+        required: [
+          "attachment_type",
+          "content_type",
+          "url",
+        ],
+        properties: FileAttachmentAssetProperties + PublicationAttachmentAssetProperties,
+      },
+      {
+        type: "object",
+        additionalProperties: false,
+        required: [
+          "attachment_type",
+          "url",
+        ],
+        properties: HtmlAttachmentAssetProperties + PublicationAttachmentAssetProperties,
+      },
+      {
+        type: "object",
+        additionalProperties: false,
+        required: [
+          "attachment_type",
+          "url",
+        ],
+        properties: ExternalAttachmentAssetProperties + PublicationAttachmentAssetProperties,
+      }
+    ],
+  },
+}
+```
+
+I've validated this proposed schema by regenerating the full schemas
+and checking the examples after making the following changes:
+
+- Change `travel_advice` to use `file_attachment_asset` and `image_asset`
+- Change `asset_link_list` to use `specialist_publisher_asset`
+
+The `what-is-content-design.json` example failed as it is missing
+content IDs for its attachments and it was using the specialist
+publisher attachment format.  For the actual change we'd either
+introduce new attachment list types, or do away with `asset_link_list`
+entirely and just define the lists in the relevant schemas.
+
+### Some design considerations
+
+**Why two lists?**
+
+As mentioned earlier, there are really two different types of
+attachments in documents.  There are attachments which appear in the
+body, and attachments which appear separately to the normal document
+text, like in publications.
+
+The `details.attachments` list would contain both types of
+attachments, so there is one list with everything in.  The
+`details.featured_attachments` list fulfils its current purpose:
+providing the ordered list of document-level attachments for
+publication-like document types.
+
+**Why make the lists mandatory?**
+
+Is there a difference between a missing list and a present, but empty,
+list?  Maybe not, but I think it's better to be explicit that there
+are no attachments for a document.
+
+**Why split `asset_link`?**
+
+`asset_link` is currently used for two different things: images and
+attachments.  However images have strictly less permissable metadata
+than attachments, it doesn't make sense to give a page image an ISBN
+for example.  These should be separate types.
+
+**Why have multiple attachment types?**
+
+Different publishing apps have different allowable metadata, so it
+makes sense to have different types constraining what a publishing app
+is able to provide.
+
+We have prior art in using a single-valued enum to make a tagged
+union: it's how the top-level schemas do it, eg:
+
+```json
+"schema_name": {
+  "type": "string",
+  "enum": [
+    "detailed_guide"
+  ]
+},
+```
+
+**Why `publication_attachment_asset`?**
+
+Whitehall calls the Publication and Consultation types, which are the
+types with featured attachments which present all the extra metadata,
+"publicationesque" types.
+
+**Why do we need a type for specialist publisher?**
+
+The `specialist_publisher_attachment_asset` is needed because
+Specialist Publisher presents the `created_at` and `updated_at` fields
+to publishers.  Specialist Publisher uses the Publishing API as its
+database, so all of that information needs to live in the content
+item.
+
+Manuals Publisher and Travel Advice Publisher have their own
+databases, rather than using Publishing API directly as a backing
+store, so they dont have this same problem.
+
+[schema.org]: http://schema.org/
+[see this comment]: https://github.com/alphagov/govuk_publishing_components/pull/1247#pullrequestreview-338008254
+[`asset_link` type]: https://github.com/alphagov/govuk-content-schemas/blob/47a751e7eb193738c2ec43be03b149527a2b8e15/formats/shared/definitions/asset_links.jsonnet
+[Publishing API can render such attachments with govspeak]: https://github.com/alphagov/publishing-api/blob/1b8e328540ab96759807787ab87dcb33bbcc72e3/app/presenters/details_presenter.rb#L71-L73


rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,147 @@
...
+         url: {
+           type: "string",
+           format: "uri",
+         },
+         preview_url: {
+           type: "string",
+           format: "uri",
+         },
+         content_type: {
+           type: "string",
```
ChrisBAshton: What do you envisage `content_type` to be? Are we talking `content_type: "image/png"` or something domain specific such as `content_type: "HtmlAttachment"? In which case, would each content type conform to its own schema?

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,147 @@
...
+
+Information about how to ask for an accessible format only makes sense
+as free text, so I see that as less useful to add to the content item,
+which I imagine as being consumed by machines.
+
+### Does this solve the problems?
+
+**Can we generate schema.org metadata for attachments?**
+
+Yes, the only thing we need for that is the attachment URL.
```
ChrisBAshton: ...and the content type, presumably?

barrucadu: Yes, that's right

rfc-116-store-attachment-data-in-content-items.md
barrucadu: `image/png`, that sort of thing.  See https://www.gov.uk/api/content/employment-tribunal-decisions/mr-f-clayton-v-royal-mail-group-ltd-1800801-2017 for an example

kevindew:
Thanks for raising this Michael 👍 This RFC is of interest to us on the Publishing Workflow team as we intend to work with the publications document type and want to get past the situation where Whitehall is currently rendering an array of HTML blobs from a content item for attachments (see: the documents section of https://www.gov.uk/api/content/government/publications/help-with-budgeting-your-universal-credit as an example). I know this was also a problem for the Content Data team when trying to extract attachment data.

We also want to have Publishing API be able to not only [render attachments](https://github.com/alphagov/publishing-api/blob/1b8e328540ab96759807787ab87dcb33bbcc72e3/app/presenters/details_presenter.rb#L71-L73) in govspeak but also images. We'd like this as we hope that if Publishing API can render Content Publisher govspeak then we could set it up to eventually automatically update embedded content (e.g. organisational contact details).

One of the things where I think you should consider more is how this data can be consolidated with the data needed to render govspeak assets from govspeak. In Govspeak we've already got [informal definitions](https://github.com/alphagov/govspeak/blame/master/README.md#L399-L409) of expected data structure of attachment files (and there'll be a much richer one in Whitehall with their extra data). If we go with a new data structure that can meet the needs for Govspeak then we've got a bunch of migration headaches to restructure everything for that. If we go with a data structure that doesn't meet govspeak needs then we're going to end up having the same data in a content item twice. My view would be that it'd be better if we could be backwards compatible with existing structures of data and actually formalise that rather than the informal approach we have.

An area not mentioned in the RFC is image files since they're treated slightly differently and are similar data that Publishing API is lacking. 

I also think we should be very cautious about adding Asset Manager communication to Publishing API. I think it could easily become complex and lead to undesirable situations - for example they can be superseded by newer versions, they can be setup with their own redirects, and their removal needs managing.

barrucadu:
Thanks Kevin, I've removed the Asset Manager bits (admittedly those were a bit of a stretch) and written about the Govspeak attachments stuff.  Do you think we should go straight to adding all of Whitehall's attachment model to Govspeak, or add it as we discover things are needed?

Another thing I'm not certain on is what exactly the relationship between the `asset_link` schema and the Govspeak attachment metadata should be.  They need to be compatible: field names matching and all fields required in one required in the other; but I'm not sure they need to be exactly the same. I'm leaning towards `asset_link` being a supertype, so we can have metadata in the content item which isn't used in the HTML output.

kevindew:
> Thanks Kevin, I've removed the Asset Manager bits (admittedly those were a bit of a stretch) and written about the Govspeak attachments stuff. Do you think we should go straight to adding all of Whitehall's attachment model to Govspeak, or add it as we discover things are needed?
> 
> Another thing I'm not certain on is what exactly the relationship between the `asset_link` schema and the Govspeak attachment metadata should be. They need to be compatible: field names matching and all fields required in one required in the other; but I'm not sure they need to be exactly the same. I'm leaning towards `asset_link` being a supertype, so we can have metadata in the content item which isn't used in the HTML output.

Great thanks. Do you want to grab a chat about the asset side of things and then we can summarise it as a comment here? it strikes me as something that might be easier to work through face-to-face.

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,110 @@
...
+   content schemas and the data Govspeak attachments use.  For
+   example, all fields required in one must be required in the other.
+   Any pairs of fields which do the same job in both should have the
+   same name.  Implement any changes arising from this.
+
+3. Add new optional fields to the `asset_link` type:
+
+   - `file_size` (used by govspeak)
+   - `preview_url` (used by whitehall)
+
```
maxgds: A piece of data we're currently missing about attachments is language. It would be very helpful to be able to mark links to foreign language content appropriately for assistive tech, and also to be able to sort attachments so that if there are 2 language versions of a page we show only the correctly associated files alongside it. It looks like this might be a place this could be implemented.

kevindew: This is a good idea, however we unfortunately we don't know whether the language of an attachment has deviated from the content of the document as a user never inputs this information.

I will raise this though in the Publishing Workflow team as we design the process for publication attachment management to see if it's something that can be improved.

kevindew:
Thanks for the chat yesterday Michael. I'll summarise the bits we covered:

Could the microformats schema.org data be exposed through HTML attributes on the attachments rather than needing to be JSON+LD data? You have an expectation that Google (and co) would get confused by a mixture of JSON+LD and embedded microformats and felt we also needed the attachment data to be available to search API which doesn't work easily if it's embedded in HTML.

We discussed what the roll out might be. You indicated that the search team would be responsible for doing this I suggested we have some info of that in the RFC.

I learnt that the [asset_link](https://github.com/alphagov/govuk-content-schemas/blob/47a751e7eb193738c2ec43be03b149527a2b8e15/formats/shared/definitions/asset_links.jsonnet) schema already exists. We both concluded it wouldn't be suitable for this to contain both images and attachment handling given the way the data stored for each of these diverges, however felt changing to a new schema for those could be backwards compatible.

Finally we talked about the intention that the attachment data can eventually replace the need to render the [documents](https://github.com/alphagov/govuk-content-schemas/blob/47a751e7eb193738c2ec43be03b149527a2b8e15/formats/publication.jsonnet#L39) in a publishing app and that we can intend for this new field to eventually lead to the deprecation of documents.

rfc-116-store-attachment-data-in-content-items.md
barrucadu: Whitehall does have a `locale` field for attachments.  I'm not sure how valuable it is, due to Kevin's comment.  I've included it (along with all the other Whitehall metadata) in the updated draft.

barrucadu:
@kevindew I updated the draft based on our chat, let me know if I've missed anything

kevindew:
Thanks Michael - this is looking great 👍 

Few comments from me are:

- I'd have thought we'd still need `type` to know whether something is file/html/external attachment?
- I'd have thought we needed `external_url` to know the URL of an external attachment
- Probably very minor but I'm not sure why `file_size` is marked as from Govspeak
- I'd suggest rather than using link in the naming you maybe use asset so `image_asset` and `attachment_asset`, this feels to have more semantic meaning than link in the context
- I'm not sure we should be setting `created_at` and `updated_at` as they may reveal information that is misleading or inaccurate (particularly if they're straight from the db) - I also don't think they're used in presentation

For compatibility with Content Publisher we'd need more fields in both image and attachment, though I think it'd be fine that we add them later when we get to actually, hopefully using it. But seems worth telling you:

For [image](https://github.com/alphagov/content-publisher/blob/46bb3d4b496b82ee38187b33ff613bec23996eff/lib/govspeak_document/in_app_options.rb#L36-L41) these are: `id`, `alt_text`, `caption` and `credit`
For [attachment](https://github.com/alphagov/content-publisher/blob/46bb3d4b496b82ee38187b33ff613bec23996eff/app/helpers/file_attachment_helper.rb#L10-L20) these are: `id` and `filename`

I also notice that `content_type` is not used for images in Content Publisher and thus not needed for govspeak so I'd suggest we drop that being a required field for an image (It seems mostly unnecessary).

Finally, it'd be good to set a deadline on this RFC if there's not one already?

@benthorner are the `print_meta_data_contact_address` and `web_isbn` fields the same fields you identified as potentially removable from Whitehall?

barrucadu:
> I'd have thought we'd still need `type` to know whether something is file/html/external attachment?

I thought we could get that from the URL: if it's a URL on GOV.UK that ends in `.html` it's an HTML attachment, if it's some other url URL on GOV.UK it's a file attachment, if it' s a non-GOV.UK URL then it's an external attachment.  We can keep that field if you think it's useful though.

> I'd have thought we needed `external_url` to know the URL of an external attachment

`url` is already a required field, and does the same job.

> Probably very minor but I'm not sure why `file_size` is marked as from Govspeak

It's in the metadata here: https://github.com/alphagov/govspeak#attachments

> I'd suggest rather than using link in the naming you maybe use asset so `image_asset` and `attachment_asset`, this feels to have more semantic meaning than link in the context

:+1:

> I'm not sure we should be setting `created_at` and `updated_at` as they may reveal information that is misleading or inaccurate (particularly if they're straight from the db) - I also don't think they're used in presentation

Yeah, we can remove those.

> For compatibility with Content Publisher we'd need more fields in both image and attachment, though I think it'd be fine that we add them later when we get to actually, hopefully using it. But seems worth telling you:
>
> For image these are: `id`, `alt_text`, `caption` and `credit`
> For attachment these are: `id` and `filename`

Would the attachment `filename` be in place of the `url` field (in which case I see your point about adding an `external_url` field), or in addition to?

> I also notice that `content_type` is not used for images in Content Publisher and thus not needed for govspeak so I'd suggest we drop that being a required field for an image (It seems mostly unnecessary).

I'll remove that too.  Content type for an image isn't too useful.

> Finally, it'd be good to set a deadline on this RFC if there's not one already?

How about end of Friday this week?

barrucadu:
I've set Friday next week as the deadline.

kevindew:
> > I'd have thought we'd still need `type` to know whether something is file/html/external attachment?
> 
> I thought we could get that from the URL: if it's a URL on GOV.UK that ends in `.html` it's an HTML attachment, if it's some other url URL on GOV.UK it's a file attachment, if it' s a non-GOV.UK URL then it's an external attachment. We can keep that field if you think it's useful though.

Yeah I think it'll be useful as it'd be really nuanced to work out which ones are which based on the URL. Html Attachments don't have a suffix of .html for instance, for assets we'd probably have to assert on the hostname (which [is under some discussion(https://github.com/alphagov/govuk-rfcs/pull/115)), and although we expect an external attachment to be a different hostname there are likely no guarantees of that.

> > I'd have thought we needed `external_url` to know the URL of an external attachment
> 
> `url` is already a required field, and does the same job.

Super

 
> > Probably very minor but I'm not sure why `file_size` is marked as from Govspeak
> 
> It's in the metadata here: https://github.com/alphagov/govspeak#attachments but it's not in the `asset_link` type, so it's a new field from Govspeak (unlike all the other fields which are from Whitehall).

Ah - so Whitehall does also have that field as well: https://github.com/alphagov/whitehall/blob/6c53d01361b83fa816dd2a05892e6f93f4cf602a/app/models/file_attachment.rb#L11

> > For compatibility with Content Publisher we'd need more fields in both image and attachment, though I think it'd be fine that we add them later when we get to actually, hopefully using it. But seems worth telling you:
> > For image these are: `id`, `alt_text`, `caption` and `credit`
> > For attachment these are: `id` and `filename`
> 
> Would the attachment `filename` be in place of the `url` field (in which case I see your point about adding an `external_url` field), or in addition to?

yes, filename should essentially be the basename of the url (however there are circumstances with govspeak when the URL doesn't have the basename so it's independent).  The filename is used to assist in working out the mime-type to display the correct metadata: https://github.com/alphagov/govuk_publishing_components/blob/ecb19eae8f421ade7b650bd1cc02493768d323eb/lib/govuk_publishing_components/presenters/attachment.rb#L23-L28

> > Finally, it'd be good to set a deadline on this RFC if there's not one already?
> 
> How about end of Friday this week?

Sounds good 👍 

---

I've thought of a rather annoying issue we may well hit with this approach due to the attachment list having potentially two meanings that may collide.

In Specialist Publisher, and future apps that may want to allow attachments to be embedded into govspeak, we have this list of attachments and govspeak can reference them. This list doesn't have a particular ordering.

In Publications and similar documents [example](https://www.gov.uk/government/publications/new-year-honours-list-2020) we have a list of attachments that is ordered and should be rendered specially on the page with extra metadata.

The complication seems to arise when we consider what might happen when we want to have both inline attachments that govspeak can render and these publication attachments on the same document. A key difference between these types of attachments is the amount of metadata, the publication ones have all the official document and similar information, whereas inline ones that are only used by govspeak would lack this.

So I'm just wondering what should be done to make them distinct. Is it some form of two different lists? or flags in a singular list, I'm not sure myself. It's worth tagging @benthorner on this too as he's working in this area.

barrucadu:
> Yeah I think it'll be useful as it'd be really nuanced to work out which ones are which based on the URL. Html Attachments don't have a suffix of .html for instance, for assets we'd probably have to assert on the hostname (which [is under some discussion(https://github.com//pull/115)), and although we expect an external attachment to be a different hostname there are likely no guarantees of that.

Ok, I've added that.

> Sounds good 👍

I actually went for Friday of next week, as this week might be a bit short if we still have issues to resolve.  But next week should be fine.

> The complication seems to arise when we consider what might happen when we want to have both inline attachments that govspeak can render and these publication attachments on the same document. A key difference between these types of attachments is the amount of metadata, the publication ones have all the official document and similar information, whereas inline ones that are only used by govspeak would lack this.

That's a good point, I'd not considered that.  Mentally I was grouping all formats into "has an attachment block but no inline attachments" and "has inline attachments but no attachment block".  But that's not necessarily the case.

I'm leaning towards the "two list" solution, as the two types of attachment are very different, so I'm not sure putting them in the same list makes sense.  There's also an issue with mixing inline and publication-style attachments in the same list with govspeak: does the `n`th attachment mean the `n`th one in the list or the `n`th inline attachment in the list?  But it'll be good to get Ben's thoughts on this.

barrucadu:
@benthorner I've updated based on our chat.  The only thing I wasn't sure about was the `id` field - do you think that should be replaced with `filename`?

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,223 @@
...
+}
+```
+
+### Some design considerations
+
+**Why two lists?**
+
+There are really two different types of attachments in documents.
+There are attachments which may be referenced in Govspeak but don't
+otherwise appear on the page, these attachments are more like metadata
```
benthorner:
```suggestion
otherwise appear on the page; in this case the attachment hashes are more like metadata
```

benthorner:
@barrucadu the approach is looking better to me, and hopefully consistent with @kevindew's ideas. The migration comments were food for thought:

   - Should we also have a step to migrate the (one) use of `asset_link`s for images?
   - Should we have a step (perhaps Step 4) to remove `title` for images?

The migration pathway might be a bit fiddly for Specialist Publisher, as suggested in one of the comments. Not a biggie, but might help to provide some explicit consideration:

   - Renaming `content_id` to `id`. We could do this by changing the app to get data from either field, and only save it in the new field.
   - Moving `details.attachments` to `attachments`. Ditto. 

We seem to have got lost with `content_id` vs. `filename` vs. `id`. I think we only need one of these, but it'll need to be a string, since Content Publisher uses a filename/slug as the `id`, whereas (e.g.) a manual section uses a UUID. It's irrelevant really, as long as the [resolution mechanisms in Govspeak](https://github.com/alphagov/govspeak/blob/master/lib/govspeak.rb#L232) all work. So, in summary, I suggest just having `id` but making it more permissive.

kevindew:
I've been thinking a bit more about this and have some fresh ideas - sorry to be giving lots of staggered inputs, it's an area I haven't thought about in detail before and I keep learning/realising new things from our team chats exploring the area.

On Publishing Workflow we've been talking about attachments a lot. We're currently using terminology of **featured attachments** to reference attachments that are ordered and shown on publications and consultations (currently in the documents array) and **in-body attachments** to reference ones that are rendered through govspeak. This terminology isn't without flaws but it seems to work in the absence of something better. 

So one thing in particular to think about, and this goes right back to the problem you're working on Michael, is whether the Search API is interested on the data for all the attachments (in-body and featured) or just the featured attachments that publications and consultations have? 

I do wonder if we really want to have all the attachment fields available for all document types, since these fields are only used by publication and consultation types. Perhaps we want to only apply this at a schema level for publication and consultation schemas? What do you think? I'm not sure if we should consider this the base set of all attachments or just special extra fields for these types. I also wonder a bit what we'd want to do if a document type wanted extra metadata on an attachment, would that be a new type in schemas or adding to base (the former seems more appealing)

Another thing I wondered is that we probably want to use the Publishing API link system to pull in data for HTML attachments. So rather than those having fields like title and url, we may instead want content_id and have a link type that we can pull these in from the links of the document. This seems a bit more fiddly but more consistent with the Publishing API and reduces duplicated data.

On the existing side of things

I think we do need attachment_type, URL alone isn't going to be enough to differentiate between attachment types, and it'd be good to not have to do any regexing of a URL. I expect we should use a oneOf in schema to decide between them as I think some fields won't be valid for all (for example content type should only matter on file attachment).

With the id field can we do anything at a schema level to enforce uniqueness across a collection? from a quick google it doesn't seem so.

Whats the logic for moving attachments to root level? This strikes me as something that increases the migration costs substantially and I'm not sure what we gain from it? I imagine it opens the door straight away to wanting to move images also out of details otherwise those things become inconsistent.

Changing details.documents seems quite risky for a roll-out because we'd not be able to have the old field and the new field at the same time. I feel like we'd do better seeing if we could go with a new name and then deprecate documents. Can you elaborate at all by what is meant by govspeak_hashes? would this be the same data that is in attachments or would it be references to data that is referenced there and pulled in? In Content Publisher we've been talking about having an array where we reference id and types to flag the featured attachments and their ordering. Something like:

```
[{ id: "file.jpg", attachment_type: "file" }, { id: "my-html-attachment", attachment_type: "html" }]
```

On the id, content_id and filename front. I think we'd want to use id as the means to embed an attachment in govspeak, content_id should become surplus to requirements (aside from a usage linking html_attachments) and I think we'd still want filename even if some apps (Content Publisher in particular) have the same value for id.

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,227 @@
...
+      preview_url: { type: "string", format: "uri", },
+      title: { type: "string", },
+      unique_reference: { type: "string", },
+      unnumbered_command_paper: { type: "boolean", },
+      unnumbered_hoc_paper: { type: "boolean", },
+      url: { type: "string", format: "uri", },
+    },
+  },
+  attachment_asset_list: {
+    description: "An ordered list of attachment links",
```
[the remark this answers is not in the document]
ChrisBAshton: Ah, this is a type which applies to both `attachments` and `details.documents` - not just `details.documents`. (Thanks Ben Thorner for clarifying)
The latter needs to be ordered but the former doesn't necessarily have to be. My question still stands - how would you order the `attachments` field to take into account both types of attachment?

barrucadu:
> So one thing in particular to think about, and this goes right back to the problem you're working on Michael, is whether the Search API is interested on the data for all the attachments (in-body and featured) or just the featured attachments that publications and consultations have?

Statistical data sets, which is one place thing we'd want to use attachments from, don't have featured attachments, [they just link to in-body attachments](https://www.gov.uk/government/statistical-data-sets/unclaimed-estates-list).

> I do wonder if we really want to have all the attachment fields available for all document types, since these fields are only used by publication and consultation types. Perhaps we want to only apply this at a schema level for publication and consultation schemas? What do you think? I'm not sure if we should consider this the base set of all attachments or just special extra fields for these types. I also wonder a bit what we'd want to do if a document type wanted extra metadata on an attachment, would that be a new type in schemas or adding to base (the former seems more appealing)
>
> Another thing I wondered is that we probably want to use the Publishing API link system to pull in data for HTML attachments. So rather than those having fields like title and url, we may instead want content_id and have a link type that we can pull these in from the links of the document. This seems a bit more fiddly but more consistent with the Publishing API and reduces duplicated data.

A way of solving both these issues would be to have multiple attachment schemas, and document schemas could use a union of whatever types makes sense.  eg, a "PublicationesqueAttachment", and "HTMLAttachment".

What do you think should be in the basic attachment type and what should be in the publication attachment type?

> I think we do need attachment_type, URL alone isn't going to be enough to differentiate between attachment types, and it'd be good to not have to do any regexing of a URL. I expect we should use a oneOf in schema to decide between them as I think some fields won't be valid for all (for example content type should only matter on file attachment).

Ok, I can bring that back.

> With the id field can we do anything at a schema level to enforce uniqueness across a collection? from a quick google it doesn't seem so.

We could still do that after validating the schema and reject documents with duplicate IDs.

> Whats the logic for moving attachments to root level? This strikes me as something that increases the migration costs substantially and I'm not sure what we gain from it? I imagine it opens the door straight away to wanting to move images also out of details otherwise those things become inconsistent.

This was Ben's suggestion, the motivation being that attachments are basically metadata.  Here's a slack excerpt:  "top-level `attachments` with ordered (backwards compatible) rendering specified in a `details.documents` section. We can use the former to replace `details.attachments` for special/manual docs, since those are only rendered inline - it’s not a curated list."

I'm kind of on-the-fence about this change, I'd be happy to keep `details.attachments`.

> Changing details.documents seems quite risky for a roll-out because we'd not be able to have the old field and the new field at the same time. I feel like we'd do better seeing if we could go with a new name and then deprecate documents. Can you elaborate at all by what is meant by govspeak_hashes? would this be the same data that is in attachments or would it be references to data that is referenced there and pulled in?

This would be an array of `{ content_type: "text/govspeak or text/html", body: "govspeak or html to render attachment" }` values.  Publishing apps could put either govspeak or HTML in there while migration happens, and publishing-api can render the govspeak.

> On the id, content_id and filename front. I think we'd want to use id as the means to embed an attachment in govspeak, content_id should become surplus to requirements (aside from a usage linking html_attachments) and I think we'd still want filename even if some apps (Content Publisher in particular) have the same value for id.

:+1:

There's a lot of feedback so I'll leave you to respond to this before I update the RFC.

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,227 @@
...
+1. Replace `content_id` in attachments with `id`.
+
+    **Migration concerns:** this would need to be done in five steps:
+    add `id`; update app to read from both fields but write to `id`;
+    republish docs; remove `content_id`; update apps to not use
+    `content_id`.
+
+1. Remove non-useful metadata fields:
+
+   - Attachments: `created_at` and `updated_at`
```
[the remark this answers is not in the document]
barrucadu: I'm ambivalent about this change.  They don't seem to be used consistently across publishing apps: specialist-publisher sets them, travel-advice-publisher doesn't.  We could keep them though.

rfc-116-store-attachment-data-in-content-items.md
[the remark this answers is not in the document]
barrucadu: If the `details.documents` list refers to attachments by `id`, then I suppose this list doesn't actually need to be ordered at all: we have in-body attachments referred to by ID, in the markup written by the publisher; and we have featured attachments referred to by ID, in the markup generated by the publishing app.  It's fine for the publishing app to internally track what order attachments need to be in the generated `details.documents`.

kevindew:
> > So one thing in particular to think about, and this goes right back to the problem you're working on Michael, is whether the Search API is interested on the data for all the attachments (in-body and featured) or just the featured attachments that publications and consultations have?
> 
> Statistical data sets, which is one place thing we'd want to use attachments from, don't have featured attachments, [they just link to in-body attachments](https://www.gov.uk/government/statistical-data-sets/unclaimed-estates-list).

Ah cool - so Search API will have an interest in all of them. Would it be only pulling in ones for particular document_types or will it pull in from any? - I ask based on the impact for Search API if schema rules for attachments differ.

> > I do wonder if we really want to have all the attachment fields available for all document types, since these fields are only used by publication and consultation types. Perhaps we want to only apply this at a schema level for publication and consultation schemas? What do you think? I'm not sure if we should consider this the base set of all attachments or just special extra fields for these types. I also wonder a bit what we'd want to do if a document type wanted extra metadata on an attachment, would that be a new type in schemas or adding to base (the former seems more appealing)
> > Another thing I wondered is that we probably want to use the Publishing API link system to pull in data for HTML attachments. So rather than those having fields like title and url, we may instead want content_id and have a link type that we can pull these in from the links of the document. This seems a bit more fiddly but more consistent with the Publishing API and reduces duplicated data.
> 
> A way of solving both these issues would be to have multiple attachment schemas, and document schemas could use a union of whatever types makes sense. eg, a "PublicationesqueAttachment", and "HTMLAttachment".
> 
> What do you think should be in the basic attachment type and what should be in the publication attachment type?

In a basic attachment we should have:

- accessible
- alternative_format_contact_email
- content_type
- filename
- file_size
- hoc_paper_number
- id
- isbn
- locale (not 100% sure of this as no app can populate it accurately yet but seems good thing to aim for)
- number_of_pages
- preview_url
- title
- url

Then a publication file should have those with:

- command_paper_number
- hoc_paper_number
- isbn
- parliamentary_session
- unique_reference
- unnumbered_command_paper
- unnumbered_hoc_paper

Then we'd have to deal with attachment_type and other subtle differences between HTML and external attachments.

> > I think we do need attachment_type, URL alone isn't going to be enough to differentiate between attachment types, and it'd be good to not have to do any regexing of a URL. I expect we should use a oneOf in schema to decide between them as I think some fields won't be valid for all (for example content type should only matter on file attachment).
> 
> Ok, I can bring that back.

👍 

> > With the id field can we do anything at a schema level to enforce uniqueness across a collection? from a quick google it doesn't seem so.
> 
> We could still do that after validating the schema and reject documents with duplicate IDs.
> 
> > Whats the logic for moving attachments to root level? This strikes me as something that increases the migration costs substantially and I'm not sure what we gain from it? I imagine it opens the door straight away to wanting to move images also out of details otherwise those things become inconsistent.
> 
> This was Ben's suggestion, the motivation being that attachments are basically metadata. Here's a slack excerpt: "top-level `attachments` with ordered (backwards compatible) rendering specified in a `details.documents` section. We can use the former to replace `details.attachments` for special/manual docs, since those are only rendered inline - it’s not a curated list."
>
> I'm kind of on-the-fence about this change, I'd be happy to keep `details.attachments`.

Yes - I ended up having a chat with Ben about it. What seems to have been lost to history is what exactly the intended purpose of details was. My understanding is that it's meant to be used for any data that isn't consistent across all document types. There's already numerous examples of metadata being present in details for example: https://www.gov.uk/api/content/aaib-reports, 

I'd suggest that unless we're really clear that this change is beneficial we should avoid it as I think it opens up a bunch of fresh questions about what else should not be in details. Whereas in details, where it is metadata that fulfils the needs of of some but not all formats, it is consistent with existing usages and avoids the needs for adding additional columns to Pub API and Content Store.

> > Changing details.documents seems quite risky for a roll-out because we'd not be able to have the old field and the new field at the same time. I feel like we'd do better seeing if we could go with a new name and then deprecate documents. Can you elaborate at all by what is meant by govspeak_hashes? would this be the same data that is in attachments or would it be references to data that is referenced there and pulled in?
> 
> This would be an array of `{ content_type: "text/govspeak or text/html", body: "govspeak or html to render attachment" }` values. Publishing apps could put either govspeak or HTML in there while migration happens, and publishing-api can render the govspeak.

I think the risk we have with this is that we need to have updated the frontend apps before any publishing apps can send this data, and that means that frontend apps wouldn't have examples to work off while being developed or a means to rollback if there are problems.

> > On the id, content_id and filename front. I think we'd want to use id as the means to embed an attachment in govspeak, content_id should become surplus to requirements (aside from a usage linking html_attachments) and I think we'd still want filename even if some apps (Content Publisher in particular) have the same value for id.
> 
> 👍
> 
> There's a lot of feedback so I'll leave you to respond to this before I update the RFC.

Thanks Michael 👍

rfc-116-store-attachment-data-in-content-items.md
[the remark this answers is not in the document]
kevindew: I think these risk producing unhelpful or incorrect data from a consumers perspective. These fields tell us when asset was created on asset manager or updated there which is not the same as when a publisher created the asset or when they are willing to acknowledge making an edit to it. 

For example, consider migrating content from Whitehall to Content Publisher. As part of this we need to re-create the assets. This would set the metadata of an attachment as showing the asset as created at and updated at for the time of the migration. For a third party using this metadata it creates the bizarre situation that an asset published maybe 5 years ago was created and updated in 2020.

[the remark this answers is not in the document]
benthorner: +1 here! From my experience on DGU and GOV.UK: avoid exposing datetimes wherever possible 😬

barrucadu:
> I think the risk we have with this is that we need to have updated the frontend apps before any publishing apps can send this data, and that means that frontend apps wouldn't have examples to work off while being developed or a means to rollback if there are problems.

Yeah, that's a good point.  Though the frontend logic would be:

```
get_attachment_html(attachment)
    if attachment.is_a? Hash
        attachment[:body]
    else
        attachment
    end
end
```

Which is fairly straightforward.  I'd feel pretty safe making that change, then we can try the publishing app & schema change in govuk-docker or integration before committing to it.

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,223 @@
...
+
+    **Migration concerns:** this would need to be done in four steps:
+    add `id`; update apps; republish docs; remove `content_id`.
+
+1. Remove non-useful metadata fields:
+
+   - Attachments: `created_at` and `updated_at`
+   - Images: `content_type`
+
+   **Migration concerns:** this would need to be done in two steps:
```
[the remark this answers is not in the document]
barrucadu: And specialist-publisher displays that metadata in a table: https://github.com/alphagov/specialist-publisher/blob/51cbb843af9650e3b4ef3648d232d57872418989/app/views/documents/show.html.erb#L76-L77

I guess it depends on how useful publishers find that information.  We could remove it, but if they use it (eg, to quickly check if someone's updated an attachment which needed changing) then we'd probably want to keep it, even though it may be confusing to end-users.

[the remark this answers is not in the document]
benthorner: Oh dear, that's a bit of a blocker - and reasonable since Specialist Publisher is trying to (ab)use Publishing API as a not-so-private database. Should we consider a(nother) separate schema for Specialist Publisher attachments, so we don't interfere with its domain model?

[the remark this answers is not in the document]
barrucadu: I think this could actually be a good candidate for a "basic attachment" type, giving us this hierarchy of types:

- `base_attachment_asset` - has just the fields used by specialist-publisher
- `attachment_asset` - imports `base_attachment_asset` and adds the other fields in Kevin's post
- `extended_attachment_asset` - imports `attachment_asset` and adds the other Whitehall fields in Kevin's post

This would make the `created_at` and `updated_at` fields present but optional in all other attachment types.  I'd be wary of doing this:

```
      base_attachment
         |       |
attachment        specialist_attachment
```

(a fork in the type tree)

Or this:

```
attachment    specialist_attachment
```

(two totally disconnected types)

As I really don't want us to end up in the situation where our attachment types end up diverging and we get logic in govspeak like "if it's a regular attachment do *this*, otherwise if it's a specialist attachment do *that*."  I feel like, if we do have multiple attachment types, there's a clear hierarchy and each extends the one below it.

If we went for this I'd suggest removing `attachment_asset_list` and having the schemas for doctypes just have an appropriate list field.

What do you think? (cc @kevindew)

stevenjmesser:
This RFC looks really interesting. I had a conversation last year with the National Archives (representing other web archiving organisations) who said it would be useful to have JSON for attachments in the API responses in content items. It would allow them to collect attachments from pages.

I'll send the notes of the conversation to Michael, who can circulate it. (Apologies if this has already been mentioned in the thread above.)

barrucadu:
The relevant bit of the notes is that they mostly use the json content items, but have to scrape the html to get "A more specific title for the specific PDF attachment we found.", "The ISSN, if any.", and "If this appears to be Command or Act paper (Cm/HC)."  And that they would like "The machine readable metadata to include explicit fields for attachments, including title, ISSN, HC/Cm status".

There's some other stuff in the notes, but that's the bit about attachments.  So this RFC solves at least one external problem, which is good!

rfc-116-store-attachment-data-in-content-items.md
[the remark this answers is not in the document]
benthorner: @barrucadu I agree we should have a simple hierarchy. I do think we should consider something custom for Specialist Publisher, because it is custom and prone to diverge, being coupled to the internal needs of the app. I'd call it a smell.

The nice thing about `extended_attachment_asset` is that it implicitly disallows further subclassing. Reading the name, I'm interpreting it as "a dump of extra stuff that we don't care about much". I wonder if we should consider something more specific, that's equally "final". The idea that comes to mind is a `government_attachment`, since (I think) that uniquely groups all the HOC/ref/etc. fields we're discussing - not sure if @kevindew meant to duplicate these fields between the types he was suggesting?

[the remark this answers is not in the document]
kevindew: Oh great catch on the updated_at / created_at for Specialist Publisher, that is a pain.

That does seem to suggest multiple attachment types unless we're just prepared to keep those fields everywhere else.

So yes if we proceed with an approach towards other types what we'd likely do is a have a base_attachment set up in our schemas that acts like an abstract class. I'd guess that'd have:

accessible, alternative_format_contact_email, content_type, filename, file_size, locale, number_of_pages, preview_url, title (required), and url (required)

A `specialist_publisher_attachment_asset` would inherit from `base_attachment_asset` that and add the following fields:

content_id (required), created_at, updated_at

A `attachment_asset` would inherit from `base_attachment_asset` and have the following fields:

id (required)

A `government_attachemnt_asset` (or `publication_attachment_asset`) would inherit from `attachment_asset` and have the following fields:

attachment_type (required), command_paper_number, hoc_paper_number, isbn, parliamentary_session, unique_reference, unnumbered_command_paper, unnumbered_hoc_paper

A couple of things to note are:

- There isn't going to be a signifier as to which of these types you'd be dealing with based on the data that would depend on an applications knowledge that a document type follows a particular schema
- As attachment_type would only be present on `government_attachment_asset` we'd have to assume that attachments without it are file attachments, which seems a pretty reasonable assumption. This would probably only be a factor for govspeak that doesn't know about document types.

Would we be able to switch any apps that use the existing asset_link away from that so they can use the `attachment_asset` or would we find other apps end up needing to use the specialist publisher variant?

This also includes my suggestions on the ever thorny problem of names :-) I think for specialist publisher we might want to just verbosely name it to avoid any confusion - as a specialist attachment sounds like it's a special type rather than for a specialist publisher. I prefer the idea of naming the publication ones government than extended as that feels more direct, though I have concerns that it may cause confusion as all attachments are government ones. I slightly prefer using the name publication myself as these are government publications, even when they're on the consultation document type. Though I can understand resistance based on document type confusion so I'd be willing to go with government if that's what's preferred.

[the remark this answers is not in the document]
barrucadu: Regarding other apps, manuals-publisher [sets `updated_at` and `created_at`](https://github.com/alphagov/manuals-publisher/blob/master/app/adapters/publishing_adapter.rb#L247-L256).  But uses mongo as its database, so those fields could be dropped without impacting the experience for publishers.

```ruby
        attachments: section.attachments.map do |attachment|
          {
            content_id: SecureRandom.uuid,
            title: attachment.title,
            url: attachment.file_url,
            updated_at: attachment.updated_at,
            created_at: attachment.created_at,
            content_type: attachment.content_type,
          }
        end,
```

And travel-advice-publisher [doesn't set them at all](https://github.com/alphagov/travel-advice-publisher/blob/master/app/presenters/asset_presenter.rb#L10-L17).

```ruby
  def present
    if asset
      {
        "url" => asset["file_url"],
        "content_type" => asset["content_type"],
      }
    end
  end
```

So I think we could change those to use `attachment_asset`, and have both of them use `SecureRandom.uuid` for the `id`.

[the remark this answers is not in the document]
barrucadu: Since it seems like I'm the only one who wants specialist publisher attachments to be the base type, I'll go for the separate type approach.  I'll write up that change now.

barrucadu:
I've updated the RFC with the current feedback and set the deadline to **next Friday**.

rfc-116-store-attachment-data-in-content-items.md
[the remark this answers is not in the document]
sihugh: > This RFC looks really interesting. I had a conversation last year with the National Archives (representing other web archiving organisations) who said it would be useful to have JSON for attachments in the API responses in content items. It would allow them to collect attachments from pages.
> 
> I'll send the notes of the conversation to Michael, who can circulate it. (Apologies if this has already been mentioned in the thread above.)

Same for Parliament people who want command papers etc. 👍

barrucadu:
@kevindew @benthorner Just a reminder that the deadline for this RFC is Friday

chao-xian:
Is there room to add a "lang" field? We currently have the problem of translated pages pointing to documents that are not of that language, and that's an accessibility issue. We don't currently have a way to flag this up. This could be a neat solution to that.

barrucadu:
Yes, there's a `locale` field here: https://github.com/alphagov/govuk-rfcs/pull/116/files#diff-91c7217baecea0e153f5e30c9ab2d8fbR210

maxgds:
@chao-xian Language was discussed here: https://github.com/alphagov/govuk-rfcs/pull/116#discussion_r364113675 :)

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,321 @@
...
+     - `alt_text`
+     - `caption`
+     - `credit`
+
+   **Migration concerns:** This can be done without breaking
+   compatibility, as all new fields are optional.
+
+1. Add new attachment type for publication attachments:
+
+    `publication_attachment_asset` (extends `attachment_asset`) with fields:
```
benthorner: Minor: it helped me to check Whitehall and realise `Publication` and `Consultation` are both `Publicationesque`, which is presumably the reason for the naming here?

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,321 @@
...
+   `image_asset`, updating the schemas which use `asset_link`
+   accordingly, with these fields:
+
+   - `base_attachment_asset`:
+     - `url` (required)
+     - `content_type` (required)
+     - `title`
+   - `attachment_asset` (extends `base_attachment_asset`):
+     - no extra fields at this time
+   - `specialist_publisher_attachment_asset` (extends `base_attachment_asset`):
```
benthorner: Can we add some indicator about the need for this type? It's the first time we've mentioned Specialist Publisher in the RFC, and I think future readers would find it helpful to understand the additional persistence problem I believe we're trying to solve here.

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,321 @@
...
+1. Add a *mandatory* `details.attachments` field to all formats which
+   can have attachments.
+
+    **Migration concerns:** this would need to be done in four steps:
+    1. Add the field as optional.
+    2. Update publishing apps to write to it.
+    3. Republish all affected documents.
+    4. Make the field mandatory.
+
+1. Add a *mandatory* `details.document_attachments` list to
```
benthorner: I think there's some potential for confusion here between `document_attachments` and `attachments`, as they superficially imply the same thing. To some extent we could resolve this with a well-written description in the schema. Did we consider `featured_attachments` as a name here?

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,321 @@
...
+      filename: { type: "string", },
+      locale: { "$ref": "#/definitions/locale", },
+      number_of_pages: { type: "integer", },
+      preview_url: { type: "string", format: "uri", },
+      title: { type: "string", },
+      url: { type: "string", format: "uri", },
+    },
+  },
+
+  attachment_asset: {
```
benthorner: It looks like we need this because Specialist Publisher will be the only app not sending IDs with its attachments. Given the relative simplicity of have a single "base" `attachment_asset`, would it be worth including Specialist Publisher in the step where we republish other attachments with arbitrary IDs? Then I think we could drop `base_attachment_asset` and just have `attachment_asset` with a mandatory ID.

There may be other considerations that are motivating this hierarchy. If so, could we articulate them in here somewhere - maybe in the 'questions' section at the bottom?

kevindew:
Thanks for the update @barrucadu - I think this is looking really good and not far off now.

A snag I can see is the required content_type field on base_attachment that wouldn't make sense with html and external attachments, and thinking about there a few more fields there that wouldn't make sense in those contexts (file_size / filename) 

I wonder if we should make base_attachment have less fields so those don't have to cascade down. Also base_attachment won't actually be a real thing just a way we share data in schemas (right?) so it may not actually need to be as highlighted here in the RFC which might make life easier if we don't need to think as much in terms of a hierarchy and the minor pain points that can give. It'd also make it easier to use a oneOf (file, html, external) schema for publication attachments to cover any divergences between these types.

Other thing I thought was about the removal of order_url and price. Depending on how quickly work will begin on implementing this we may get caught where Publishing Workflow haven't finished or come to full conclusions on its removal. This may not really be a problem going forward as it can just be dealt with post the RFC depending on context/status but thought I'd flag.

rfc-116-store-attachment-data-in-content-items.md
barrucadu: Yes, do you think that would be a good name, if it were explained in the RFC?

rfc-116-store-attachment-data-in-content-items.md
barrucadu: :+1:

rfc-116-store-attachment-data-in-content-items.md
barrucadu: I think `featured_attachments` would be a good name for it

rfc-116-store-attachment-data-in-content-items.md
barrucadu: I agree, when writing up the schema I did start to question why `attachment_asset` needed to be a separate thing.  I'll add a step about adding IDs to specialist-publisher attachments.

barrucadu:
> A snag I can see is the required content_type field on base_attachment that wouldn't make sense with html and external attachments, and thinking about there a few more fields there that wouldn't make sense in those contexts (file_size / filename)
> [...]
> It'd also make it easier to use a oneOf (file, html, external) schema for publication attachments to cover any divergences between these types.

Do you think `publication_attachment_asset` should be split into those three types?

> Also base_attachment won't actually be a real thing just a way we share data in schemas (right?)

I was imagining it to be a real thing.

rfc-116-store-attachment-data-in-content-items.md
benthorner: I think `publication_attachment_asset` is fine, and it would be helpful to explain the reasoning briefly.

rfc-116-store-attachment-data-in-content-items.md
benthorner: @kevindew?

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,391 @@
...
+    4. Make the field mandatory.
+
+1. Add a *mandatory* `details.featured_attachments` list to
+   publication formats, which is an ordered list of govspeak or HTML
+   strings (using the `#/definitions/multiple_content_types` schema)
+   rendering attachments from `details.attachments`.
+
+   **Migration concerns:** this would need to be done in four steps:
+   1. Add the field as optional.
+   2. Update publishing apps (just Whitehall and Content Publisher) to write to it.
```
benthorner: Just a minor thing to check: we're intentionally leaving this open as to whether the app, or the Publishing API, does the actual rendering?

barrucadu: I think it would be nice if the Publishing API rendered it, for consistency between apps, but ultimately it doesn't matter too much so long as all the metadata is still in `details.attachments`.

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -0,0 +1,391 @@
...
+          content_id: { "$ref": "#/definitions/guid", },
+          created_at: { format: "date-time", },
+          updated_at: { format: "date-time", },
+        },
+      },
+    ],
+  },
+
+  publication_attachment_asset: {
+    allOf: [
```
kevindew: I don't think this allOf is going to quite work because the various attachment asset schemas are set to have `additionalProperties: false` which would mean they fail the oneOf schema if they have any of the publication properties. 

I am impressed though with this allOf approach, I didn't know about this technique to allow inheritance - very clever.

barrucadu: Hmm, that's annoying.  The json-schema maintainer says that json-schema isn't for inheritance (https://github.com/json-schema-org/json-schema-spec/issues/348#issuecomment-322940347), so there might be no better solution than duplicating all the fields.

kevindew: Sure - field duplication is how we've solved this problem in the past so would be consistent with our prior approaches.

As we use [jsonnet](https://jsonnet.org/) to define the schema definitions we can cut this duplication out of the code that manages schemas. For instance, my initial expectation of this base_attachment_asset was that it'd be something that existed in jsonnet.

barrucadu: :+1:

barrucadu:
I'm leaning towards removing the `html_attachment_asset` and `external_attachment_asset` schemas, as they're no longer used in the `oneOf` in `publication_attachment_asset`.  Other than that, I think I'm now pretty happy with this.

PR #118, opened 2020-03-03
Add proposal to store Global Content in content-store

chao-xian:


kevindew:
Hey, in lieu of a meeting this week and with me about to go on leave for a fortnight I thought I'd leave some notes on the parts of this I've done some work on. 

So, where were we last? We were considering options for the removal of static and in doing so what could pick up static's responsibilities. We felt it'd be good if content items had all the necessary information so then frontend applications wouldn't need an extra dependency. We considered first whether this is data that could be shared via Publishing API link expansion with every content item, this was rejected because a change to the data would take hours to propagate for each item. Instead we thought the content store could combine this data since it would apply the same to every single content item. We then wondered whether we should have a special global content item or use the homepage for any global data. We decided on the homepage since any special content item would need a base path to be present in the content store.

I put together an example of a [schema approach in govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas/compare/homepage-global?diff=unified) where the [homepage can have a global field](https://github.com/alphagov/govuk-content-schemas/commit/9b732abb5a0822231a653a80b3d2c77ce01a401c) which stores all the necessary data to populate GOV.UK header and footer components to render a GOV.UK page. Then I added the ability for all [GOV.UK frontend schemas to have the global field](https://github.com/alphagov/govuk-content-schemas/commit/34e6f32998269fed7e9cb782544e6d133f7ad898), some key thoughts on this work were:

- is global the right terminology?
- is this missing any fields that are needed?
- what is the right level of customisability for footer links? some options would be:
  - all managed in component, nothing in content items at all (downside is a change would require co-ordinated deploys)
  - sections not customisable but links for each are (as per my appraoch, downside is adding a new section like COVID-19 needs to be in all frontend apps before can be enabled)
  - sections can be fully controlled by content item (downside would be that this would be most complex and may still not be enough for the next change to the footer)

Then I put together an example of [what changes would be made in frontend to publish the homepage](https://github.com/alphagov/frontend/commit/36aface888526940b08cc09a5a0c0cdd2dc46ae6). This would have some consequences:

- adding a new alert bar (which I renamed from global bar in govuk publishing components) would be managed in frontend rather than govuk_publishing_components making it easier to test and deploy
- emergency banners would be managed through code rather than the rake task which may offer some risks for sensitive information, but should make them easier to deploy and remove the existing infrastructure for them

Finally I put together an example of [changes to the content store to embed the global data](https://github.com/alphagov/content-store/commit/2ff32d382b2b1859f39eddb5c97ef8a4daa57c91). This was mostly quite painless but did reveal a couple of potential areas of confusion/annoyance. These are:

- to use a local version of the content store you'd want to have the homepage published which adds some extra effort for a dev
- the updated time of content items could be a little confusing, in my spike I decided it was best to use which ever updated at was most recent for homepage or request content item when presenting

Other things pondered about:

- All frontend apps would want access to a [DRAFT_ENVIRONMENT environment variable](https://github.com/alphagov/govuk-puppet/blob/0b768d56e4d0cbe3681ba40735a482a6bc57de9a/modules/govuk/manifests/apps/static.pp#L92-L98) so that govuk_publishing_components could apply the "DRAFT" background text
- We'd need to move the [error page rendering](https://github.com/alphagov/static/tree/e5692c3119c8984f2248b36ff88630408db1f6f6/app/views/root) somewhere, I'd suggest using [frontend](https://github.com/alphagov/frontend)
- It'd be nice if all links weren't hardcoded with titles and some could use Publishing API link expansion, however I think that is a future development which is best considered out of scope (it's not a part of the system being replaced)

kevindew:
This hasn't moved for a few months so may as well close for now. Can always re-open when time and will and re-aligned.

PR #119, opened 2020-03-06
RFC 119: Merge mainstream browse and specialist topics into taxons

theseanything:
🗒 TLDR: We should merge mainstream browse and specialist topics content items into taxons.

🖼 [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/rfc-119/rfc-119-merge-mainstream-browse-topics-and-taxons.md)

📆 **Deadline for comments**: Friday 27th November 2020

### DIFF
--- a/rfc-119-merge-mainstream-browse-topics-and-taxons.md
+++ b/rfc-119-merge-mainstream-browse-topics-and-taxons.md
@@ -1,0 +1,169 @@
+# RFC 119: Merge mainstream browse, specialist topic into to taxon content items
+
+## Summary
+
+Mainstream browse, specialist topic and taxon content items are functionally equivalent, in that they represent a category that groups content items. We would like to maintain future consistency between our taxonomy, mainstream browse and specialist topic titles. We propose merging the attributes of mainstream browse and topic content items into taxons and deprecating use of mainstream browse and specialist topics in the Publishing API and Content Store.
+
+## Problem
+
+We currently maintain three seperate ways to categorise content items - mainstream browse pages, specialist topics and taxons. There is currently ongoing work to make sure mainstream browse and specialist topics have equivalents in the taxonomy i.e. mainstream browse and specialist topics become a subset of taxons. This is being done by Content Designers who are liaising with departments to come to agreement.
+
+Once this work is complete we want to maintain alignment and prevent divergence from taxons. We would have to put constraints in collections publisher and content tagger (or any future apps) to ensure that mainstream browse and specialist topics continue to be subsets of taxons. This would involve adding complex logic to the publishing system to deal with addition, changing or removal of these content items.
+
+## Proposal
+
+We propose merging the attributes of mainstream browse and specialist topic content items into **existing** taxon content items with the same `title`. (For clarity we aren't proposing the creation of new taxons or taxon tree.)
+
+Generally there are two main type of information we want to tranfer from mainstream browse and specialist topic to the equivalent taxon. The first is the links to content items. Currently content items (guidance pages, manuals etc..) do not contain links to mainstream browse or specialist topics. It's the reverse, mainstream browse pages and specialist topics have a list of content that's associated to them. This list we can migrate over to the taxon.
+
+Secondly, we'd carry across the relationships (links) between mainstream browse pages/specialist topics. This manifests as multiple hierarchies within the same taxons. This is because different relationships are useful for different things ie. the currently hierarchy for taxonomy is good for emails subscriptions and organising taxons, but not so great at modelling navigation. So we'd bring over the mainstream browse hierarchy to serve that purpose.
+
+#### Top level attributes
+
+All top level attributes will continue to have the existing values in the taxon:
+
+- `title`
+- `description`
+- `analytics_identifier`
+- `content_id`
+- `document_type`
+- `first_published_at`
+- `locale`
+- `phase`
+- `public_updated_at`
+- `publishing_scheduled_at`
+- `rendering_app`
+- `scheduled_publishing_delay_seconds`
+- `schema_name`
+- `updated_at`
+- `withdrawn_notice`
+- `publishing_request_id`
+
+This is because they don't cause conflict or the mainstream or specialist topic value becomes redundant.
+
+- `publishing_app`
+
+Currently references Collections Publisher for mainstream and specialist topics, however references Content Tagger for taxons. We propose that `publishing-app` remains Content Tagger as it still has the responsibility of creating and removing taxons, even though some sub properties (i.e. links and details) may be set by Collections Publisher. In the future, we could also look at consolidating the publishing functionality of Collections Publisher and Content Tagger.
+
+- `base_path`
+
+Mainstream browse and specialist topic pages have different `base_path` to taxons. We propose that taxons keep their existing `base_path`. We could use new mainstream browse/specialist topic links within taxons to generate old base paths or find the appropriate taxon from an existing mainstream browse or specialist topic base path. Alternatively, we keep the old base paths as renamed attributes (e.g. `mainstream_browse_base_path`).
+
+#### Link attributes
+
+Taxons continue to have the following links:
+- `legacy_taxons` (potentially remove once specialist topics are removed)
+- `parent_taxons`
+- `available_translations`
+
+Add from mainstream browse:
+- `browse_parent` (parent taxons which should be used for mainstream browse)
+- `browse_children` (reverse links to `browse_parent`)
+
+Add from specialist topic:
+- `specialist_topic_parent` (parent taxons which should be used for specialist topics)
+- `specialist_topic_children` (reverse links to `specialist_topic_parent`)
+
+Dropped from mainstream browse:
+- `top_level_browse_pages` (an be derived from `base_path` and following `browse_parent`)
+- `active_top_level_browse_page` (can be derived from `base_path`)
+- `second_level_browse_pages` (an be derived from `base_path` and following `browse_parent`)
+- `related_topics` (is the related specialist topic, and would become a self reference)
+- `primary_publishing_organisation` (redundant)
+- `available_translations` (redundant)
+
+Dropped from specialist topic:
+- `taxons` (is the taxon, would become a self reference)
+- `topic_taxonomy_taxons` (redundant)
+- `primary_publishing_organisation` (redundant)
+- `available_translations` (redundant)
+
+#### Details attributes
+
+Taxons continue to have the following details:
+- `internal_name`
+- `notes_for_editors`
+- `visible_to_departmental_editors`
+
+Added from mainstream browse:
+- `browse_groups` (the curated groupings of browse children)
+
+Added from specialist topic:
+- `specialist_topic_groups` (the curated groupings of specialist_topic_children)
+
+Dropped from mainstream browse:
+- `internal_name` (redundant)
+- `second_level_ordering` (can be deduced by following link to parent)
+- `ordered_second_level_browse_pages` (can be deduced by following link to parent)
+
+Dropped from specialist topics:
+- `internal_name` (redundant)
+
+### Example of proposed taxon content item after merging:
+```json
+{
+  "title": "Universal Credit",
+  "description": "Applying, signing into your account, and help with housing, disability, health conditions and unemployment",
+  "base_path": "/welfare/universal-credit",
+  "content_id": "62fcbba5-3a75-4d15-85a6-d8a80b03d57c",
+  "document_type": "taxon",
+  "schema_name": "taxon",
+  "publishing_app": "content-tagger",
+  "rendering_app": "collections",
+  "locale": "en",
+  "phase": "live",
+  "first_published_at": "2018-03-08T16:38:00.000+00:00",
+  "public_updated_at": "2018-08-22T13:14:40.000+00:00",
+  "updated_at": "2019-08-09T20:07:42.761Z",
+  "publishing_request_id": "12821-1535466331.861-10.3.3.1-394",
+  "publishing_scheduled_at": null,
+  "scheduled_publishing_delay_seconds": null,
+  "analytics_identifier": null,
+  "withdrawn_notice": {},
+  "links": {
+    "parent_taxons": [],
+    "browse_children": [],
+    "browse_parent": [],
+    "specialist_topic_children": [],
+    "specialist_topic_parent": [],
+    "primary_publishing_organisation": [],
+    "available_translations": []
+  },
+  "details": {
+    "internal_name": "Universal Credit [M]",
+    "notes_for_editors": "",
+    "visible_to_departmental_editors": false,
+    "specialist_topic_groups": [],
+    "browse_groups": [],
+    "browse_ordering": "curated",
+    "ordered_second_level_browse_pages": []
+  }
+}
+```
+
+We would probably approach this work in two stages by merging mainstream browse *and then* specialist topics. After both are merged we'd look at deprecating mainstream browse and specialist topics content items in the Publishing API and Content Store.
+
+To do this we'd change application dependencies to read from taxons. Application dependencies we'd have to consider:
+
+- Rendering `/browse/*` and `/topic/*` pages in Collections Frontend
+- Displaying metrics for `/browse/*` and `/topic/*` pages in Content Data
+- Signing up and email subscriptions per specialist topic
+- Finding browse links in the Knowledge Graph
+
+Once we are confident we are solely using the taxons, we can unpublish the existing mainstream browse and specialist topic content items.
+
+## Benefits
+This would allow us to ensure ongoing consistency of our "Information Architecture" and grouping of content, without the complexity of maintaining three document types or seperate sets of content items.
+
+## Appendix
+Examples of equivalent content items:
+
+Mainstream browse
+URL: https://www.gov.uk/api/content/browse/benefits/universal-credit
+
+Specialist topic 
+URL: https://www.gov.uk/api/content/topic/benefits-credits/universal-credit
+
+Taxon
+URL: https://www.gov.uk/api/content/welfare/universal-credit
+
--- a/rfc-119-merge-mainstream-browse-topics-and-taxons.md
+++ b/rfc-119-merge-mainstream-browse-topics-and-taxons.md
@@ -33,7 +33,7 @@ We propose merging the attributes of mainstream browse and specialist topic cont
 
 Generally there are two main type of information we want to tranfer from mainstream browse and specialist topic to the equivalent taxon. The first is the links to content items. Currently content items (guidance pages, manuals etc..) do not contain links to mainstream browse or specialist topics. It's the reverse, mainstream browse pages and specialist topics have a list of content that's associated to them. This list we can migrate over to the taxon.
 
-Secondly, we'd carry across the relationships (links) between mainstream browse pages/specialist topics. This manifests as multiple hierarchies within the same taxons. This is because different relationships are useful for different things ie. the currently hierarchy for taxonomy is good for emails subscriptions and organising taxons, but not so great at modelling navigation. So we'd bring over the mainstream browse hierarchy to serve that purpose.
+Secondly, we'd carry across the relationships (links) between mainstream browse pages/specialist topics. This manifests as multiple hierarchies within the same taxons. This is because different relationships are useful for different things i.e. the current hierarchy for taxonomy is used for modelling emails subscriptions, but cannot be used for modelling navigation. So we'd bring over the mainstream browse hierarchy to serve that purpose.
 
 #### Top level attributes
 
--- a/rfc-119-merge-mainstream-browse-topics-and-taxons.md
+++ b/rfc-119-merge-mainstream-browse-topics-and-taxons.md
@@ -4,9 +4,26 @@
 
 Mainstream browse, specialist topic and taxon content items are functionally equivalent, in that they represent a category that groups content items. We would like to maintain future consistency between our taxonomy, mainstream browse and specialist topic titles. We propose merging the attributes of mainstream browse and topic content items into taxons and deprecating use of mainstream browse and specialist topics in the Publishing API and Content Store.
 
+## Background
+
+There are currently three different and independent ways to categorise content items - [mainstream browse pages][mainstream-browse-page-docs], [specialist topics][specialist-topic-docs] and [taxons][taxon-docs]. A content item can be associated with one or more mainstream browse pages, specialist topics or taxons. Each of which represents a grouping of content items that are related in subject matter. For example content items representing a "Income Tax" guidance page and a "Inheritance Tax" guidance page are assigned to the "Tax" mainstream browse page.
+
+Mainstream browse pages, specialist topics and taxons themselves are represented as content items. Each type can contain links to other items of the same type and usually do so building a hierarchical tree structure. For example, the "Child benefit" and "Universal Credit" specialist topics are linked and represented as children of the "Benefits" specialist topic. For each type there is one "root" item that represents the start of the hierarchy. There is an exception for taxons, where there are two hierarchies with two distinct roots and sets of items - [topic taxonomy][topic-taxonomy-docs] and [world taxonomy][world-taxonomy-docs]. For purposes of this RFC ignore the world taxonomy.
+
+
 ## Problem
 
-We currently maintain three seperate ways to categorise content items - mainstream browse pages, specialist topics and taxons. There is currently ongoing work to make sure mainstream browse and specialist topics have equivalents in the taxonomy i.e. mainstream browse and specialist topics become a subset of taxons. This is being done by Content Designers who are liaising with departments to come to agreement.
+We have to maintain three separate types of categorisation for content items - mainstream browse pages, specialist topics and taxons. This means the categories themselves and the content assigned to them may be different between each type. For example, the "Education and learning" mainstream browse page is the logical equivalent of the "Education, training and skills" taxon and should be the same, however they differ in name and in the set of content items they represent. There are also categories that only exist in some types and not others, for example the "Setting up a business" mainstream browse page has no equivalent taxon.
+
+Examples of equivalent content items:
+
+Mainstream browse page: https://www.gov.uk/api/content/browse/benefits/universal-credit
+
+Specialist topic: https://www.gov.uk/api/content/topic/benefits-credits/universal-credit
+
+Taxon: https://www.gov.uk/api/content/welfare/universal-credit
+
+There is currently ongoing work to ensure mainstream browse and specialist topics have equivalents in the taxonomy, i.e. mainstream browse and specialist topics become a subset of taxons. This is being done by Content Designers who are liaising with departments to come to agreement.
 
 Once this work is complete we want to maintain alignment and prevent divergence from taxons. We would have to put constraints in collections publisher and content tagger (or any future apps) to ensure that mainstream browse and specialist topics continue to be subsets of taxons. This would involve adding complex logic to the publishing system to deal with addition, changing or removal of these content items.
 
@@ -152,18 +169,15 @@ To do this we'd change application dependencies to read from taxons. Application
 
 Once we are confident we are solely using the taxons, we can unpublish the existing mainstream browse and specialist topic content items.
 
+This work should not introduce any user facing changes.
+
 ## Benefits
 This would allow us to ensure ongoing consistency of our "Information Architecture" and grouping of content, without the complexity of maintaining three document types or seperate sets of content items.
 
-## Appendix
-Examples of equivalent content items:
 
-Mainstream browse
-URL: https://www.gov.uk/api/content/browse/benefits/universal-credit
-
-Specialist topic 
-URL: https://www.gov.uk/api/content/topic/benefits-credits/universal-credit
-
-Taxon
-URL: https://www.gov.uk/api/content/welfare/universal-credit
+[mainstream-browse-page-docs]: https://docs.publishing.service.gov.uk/document-types/mainstream_browse_page.html
+[specialist-topic-docs]: https://docs.publishing.service.gov.uk/document-types/topic.html
+[taxon-docs]: https://docs.publishing.service.gov.uk/document-types/taxon.html
+[topic-taxonomy-docs]: https://docs.publishing.service.gov.uk/manual/taxonomy.html
+[world-taxonomy-docs]: https://docs.publishing.service.gov.uk/manual/world-taxonomy.html
 


rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,149 @@
...
+- available_translations
+
+Add from mainstream browse:
+- children (rename to mainstream_browse_children)
+- top_level_browse_pages (potentially could be dropped, as can be derived from path)
+- second_level_browse_pages (superseded by mainstream_browse_parent)
+
+Add from specialist topic:
+- children (rename to specialist_topic_children)
+- parent (rename to specialist_topic_parent)
```
thomasleese: This might be my lack of knowledge of the taxonomy showing here... Taxons already model a parent/child relationship [through the `parent_taxons` link and `child_taxons` reverse link](https://github.com/alphagov/publishing-api/blob/0baad432aa0f573322bc2c28a10afdd968031ab4/lib/expansion_rules.rb#L40), rather than adding `children` and `parent` links, if we're merging them all into taxons, can we not use the existing `parent_taxons` and `child_taxons` links.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Link attributes
+
+Taxons continue to have the following links:
+- legacy_taxons (potentially remove once specialist topics are removed)
+- parent_taxons
+- available_translations
+
+Add from mainstream browse:
+- children (rename to mainstream_browse_children)
```
thomasleese: Would this be [a reverse link](https://github.com/alphagov/publishing-api/blob/master/doc/link-expansion.md#reverse-links)?

theseanything: Yup!

rfc-119-merge-mainstream-browse-topics-and-taxons.md
theseanything: So we'd be looking at having multiple hierarchies within the same taxons as the different sets of relationships are useful for different things ie. the currently hierarchy for taxonomy is good for emails subscriptions and organising taxons, but not so great at modelling navigation. So we'd bring over the mainstream browse hierarchy to serve that purpose.

theseanything:
Thanks @thomasleese for the early review! 

> * Do we have a plan on how we're going to re-link content that's currently linked to a specialist sector or a mainstream browse page?
Currently content items (guidance pages, manuals etc..) contain the links to mainstream browse or specialist topic. It's the reverse, mainstream browse content items have a list of content that's associated to them. This list we can migrate over to the taxon.

> * What's going to happen to the specialist sector and mainstream browse content items once they've merged into a taxon content item?
After making sure all the application dependencies are reading from the taxons, we'd probably just "unpublish" the pecialist sector and mainstream browse content items to keep a history.

> * It's currently possible to subscribe to email alerts for specialist sectors, do we have any thought as to how we would manage this transition over to subscribing to taxons instead?
Yes it is! We'd have to convert existing emails subscription use the relevant taxons, but be concerned only with the specialist sector content tagged.

I'll update the RFC to include answers the these comments! :smile

theseanything:
> I'm personally still not too clear on why we need both specialist sectors and taxons, but I guess that's an issue for content to deal with.

@barrucadu - agree and this might be a stepping stone towards removing specialist topics. But we haven't done the think around that and possibly out of scope for this RFC.

> Can email-alert-api do this already? Or would this be a new thing?

I think email alert api can have SubscriberLists based on taxons or specialist topics. So we'd have to update SubscriberLists to reference the relevant taxon rather than specialist topic. However those SubscriptionLists still need to only send emails for content changes to content tagged via the specialist topics relationship, rather than content tagged via the taxon relationship. Hopefully that makes sense?

kevindew:
Hey @theseanything, just wanted to check in whether this RFC is still going to be going through given all the recent changes?

theseanything:
> Hey @theseanything, just wanted to check in whether this RFC is still going to be going through given all the recent changes?

Sorry @kevindew for not replying sooner! Ah so given the recent changes, I don't know who would actually be implementing this. (I'd hope for this work to be picked up again in the not too distant future). I don't what our process is about pausing RFCs - as I think it's still relevant and don't want to completely bin the proposal. Also aware that it's not ideal to approve and merge an RFC has no dedicated resources to actually making it happen. Shall I just close this PR and reopen when appropriate in the future?

kevindew:
@theseanything Thanks, yeah I think close and re-open later is the best approach to pause an RFC. Can also pursue getting it complete and merged ready for later if you have the time but if not best to pause it.

theseanything:
Closing for now. To be re-opened when we are able to implement.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,169 @@
...
+    "visible_to_departmental_editors": false,
+    "specialist_topic_groups": [],
+    "browse_groups": [],
+    "browse_ordering": "curated",
+    "ordered_second_level_browse_pages": []
+  }
+}
+```
+
+We would probably approach this work in two stages by merging mainstream browse *and then* specialist topics. After both are merged we'd look at deprecating mainstream browse and specialist topics content items in the Publishing API and Content Store.
```
huwd, line 144: I've been wondering about this in relation to [the comments above about self-referential links](https://github.com/alphagov/govuk-rfcs/pull/119/files#diff-b26d5a4ca9078275cfb5727792c332e19f06339d723e85ccec934c15d9e9668eR71) - so seems like if we stage this we might want to modify some of the removals or create some initial duplication that we only clear up in a final pass?

Is that still easier than trying to munge these both together at the same time?

I can imagine both at once creates quite a lot of frontend and publishing work along-side the data reconciliation?

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Proposal
+
+We propose merging the attributes of mainstream browse and specialist topic content items into **existing** taxon content items with the same `title`. (For clarity we aren't proposing the creation of new taxons or taxon tree.)
+
+Generally there are two main type of information we want to tranfer from mainstream browse and specialist topic to the equivalent taxon. The first is the links to content items. Currently content items (guidance pages, manuals etc..) do not contain links to mainstream browse or specialist topics. It's the reverse, mainstream browse pages and specialist topics have a list of content that's associated to them. This list we can migrate over to the taxon.
+
+Secondly, we'd carry across the relationships (links) between mainstream browse pages/specialist topics. This manifests as multiple hierarchies within the same taxons. This is because different relationships are useful for different things ie. the currently hierarchy for taxonomy is good for emails subscriptions and organising taxons, but not so great at modelling navigation. So we'd bring over the mainstream browse hierarchy to serve that purpose.
```
benthorner, line 19: I don't think there's any evidence that taxons are good for email subscriptions. We suspect the opposite: they are far too generic to represent any intent of the user, who will end up receiving all content tagged the taxon and its children, including things like "meeting minutes". The Notifications team are aiming to do a discovery that will include finding out whether we need to support taxon subscriptions at all.

I'm also not sure what we mean by "organising taxons" - in one sentence this reads as "the taxonomy is good for organising taxons".

Perhaps the conclusion is that it's not got for anything? 😆

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,169 @@
...
+
+This is because they don't cause conflict or the mainstream or specialist topic value becomes redundant.
+
+- `publishing_app`
+
+Currently references Collections Publisher for mainstream and specialist topics, however references Content Tagger for taxons. We propose that `publishing-app` remains Content Tagger as it still has the responsibility of creating and removing taxons, even though some sub properties (i.e. links and details) may be set by Collections Publisher. In the future, we could also look at consolidating the publishing functionality of Collections Publisher and Content Tagger.
+
+- `base_path`
+
+Mainstream browse and specialist topic pages have different `base_path` to taxons. We propose that taxons keep their existing `base_path`. We could use new mainstream browse/specialist topic links within taxons to generate old base paths or find the appropriate taxon from an existing mainstream browse or specialist topic base path. Alternatively, we keep the old base paths as renamed attributes (e.g. `mainstream_browse_base_path`).
```
benthorner, line 50: Are we proposing to redirect all the old base paths, or have different "views" of the taxon for different paths? Content items already support [multiple "routes"](https://github.com/alphagov/govuk-content-schemas/blob/b012c9c8c28781393f25f818a30dfd36a2cc4ae3/formats/shared/definitions/routes_redirects.jsonnet#L3) - perhaps we could use that functionality instead.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,169 @@
+# RFC 119: Merge mainstream browse, specialist topic into to taxon content items
+
+## Summary
+
+Mainstream browse, specialist topic and taxon content items are functionally equivalent, in that they represent a category that groups content items. We would like to maintain future consistency between our taxonomy, mainstream browse and specialist topic titles. We propose merging the attributes of mainstream browse and topic content items into taxons and deprecating use of mainstream browse and specialist topics in the Publishing API and Content Store.
+
+## Problem
+
+We currently maintain three seperate ways to categorise content items - mainstream browse pages, specialist topics and taxons. There is currently ongoing work to make sure mainstream browse and specialist topics have equivalents in the taxonomy i.e. mainstream browse and specialist topics become a subset of taxons. This is being done by Content Designers who are liaising with departments to come to agreement.
```
benthorner, line 9: I think it would be helpful to link to the schemas/examples for these formats and/or explain more about them. Many people may not be aware of what they are, and the fact that they're so similar makes them hard to understand at a glance - another argument for getting rid of them, of course!

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,169 @@
...
+- `legacy_taxons` (potentially remove once specialist topics are removed)
+- `parent_taxons`
+- `available_translations`
+
+Add from mainstream browse:
+- `browse_parent` (parent taxons which should be used for mainstream browse)
+- `browse_children` (reverse links to `browse_parent`)
+
+Add from specialist topic:
+- `specialist_topic_parent` (parent taxons which should be used for specialist topics)
```
benthorner, line 64: I can sort of see the need for a taxon to have "specialist children". How come it needs a different parent link than `parent_taxon`, though?

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,169 @@
...
+Taxons continue to have the following details:
+- `internal_name`
+- `notes_for_editors`
+- `visible_to_departmental_editors`
+
+Added from mainstream browse:
+- `browse_groups` (the curated groupings of browse children)
+
+Added from specialist topic:
+- `specialist_topic_groups` (the curated groupings of specialist_topic_children)
```
benthorner, line 92: Since we're merging these in, we have an opportunity to give them a more meaningful name - so we don't have to explain what they are.

Could we rename them to clarify that they're "overrides" of some kind?

rfc-119-merge-mainstream-browse-topics-and-taxons.md
theseanything: Yup, how about this?

---

## Background

There are currently three different and independent ways to categorise content items - [mainstream browse pages](https://docs.publishing.service.gov.uk/document-types/mainstream_browse_page.html), [specialist topics](https://docs.publishing.service.gov.uk/document-types/topic.html) and [taxons](https://docs.publishing.service.gov.uk/document-types/taxon.html). A content item can be associated with one or more mainstream browse pages, specialist topics or taxons. Each of which represents a grouping of content items that are related in subject matter. For example content items representing a "Income Tax" guidance page and a "Inheritance Tax" guidance page are assigned to the "Tax" mainstream browse page. 

Mainstream browse pages, specialist topics and taxons themselves are represented as content items. Each type can contain links to other items of the same type and usually do so building a hierarchical tree structure. For example, the "Child benefit" and "Universal Credit" specialist topics are linked and represented as children of the "Benefits" specialist topic. For each type there is one "root" item that represents the start of the hierarchy. There is an exception for taxons, where there are two hierarchies with two distinct roots and sets of items - [topic taxonomy](https://docs.publishing.service.gov.uk/manual/taxonomy.html) and [world taxonomy](https://docs.publishing.service.gov.uk/manual/world-taxonomy.html). For purposes of this RFC ignore the world taxonomy. 

See the [appendix](#appendix) and developer docs for more information and examples.

## Problem

We have to maintain three separate types of categorisation for content items - mainstream browse pages, specialist topics and taxons. This means the categories themselves and the content assigned to them may be different between each type. For example, the "Education and learning" mainstream browse page is the logical equivalent of the "Education, training and skills" taxon and should be the same, however they differ in name and in the set of content items they represent. There are also categories that only exist in some types and not others, for example the "Setting up a business" mainstream browse page has no equivalent taxon. 

There is currently ongoing work to ensure mainstream browse and specialist topics have equivalents in the taxonomy, i.e. mainstream browse and specialist topics become a subset of taxons. This is being done by Content Designers who are liaising with departments to come to agreement.

Once this work is complete we want to maintain alignment and prevent divergence from taxons. We would have to put constraints in collections publisher and content tagger (or any future apps) to ensure that mainstream browse and specialist topics continue to be subsets of taxons. This would involve adding complex logic to the publishing system to deal with addition, changing or removal of these content items.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
theseanything: Ah yup - so here we are talking specifically about the **hierarchical** property of the taxonomy. I've removed the "organising taxons" as its confusing - was trying to describe that the hierarchy makes it easy to find specific taxons for publishers.

 I think the point to be made is that different hierarchical structures are used for different purposes. Would this be better?
 
```suggestion
Secondly, we'd carry across the relationships (links) between mainstream browse pages/specialist topics. This manifests as multiple hierarchies within the same taxons. This is because different relationships are useful for different things i.e. the current hierarchy for taxonomy is used for modelling emails subscriptions, but is less useful for modelling navigation. So we'd bring over the mainstream browse hierarchy to serve that purpose.
```

rfc-119-merge-mainstream-browse-topics-and-taxons.md
theseanything: So _I think_ this talking about application code that makes use of the `base_path` attribute in a content item. E.g.`content_item["base_path"]`

I think can easily derive the relevant base paths by prepended or dropping the "topic" or "browse" bit.

i.e.
Mainstream browse/topic base path <=> Taxon base path
/browse/foo/bar                 <=> /foo/bar
/topic/foo/bar                     <=> /foo/bar

```suggestion
Mainstream browse and specialist topic pages have a different `base_path` to taxons. We propose that taxons keep their existing `base_path`. 

Applications can easily derive the old base paths for mainstream browse or specialist topic by prepending "browse" or "topic" respectively e.g. `"/browse/#{taxon_base_path}"` or `"/topic/#{taxon_base_path}"`. Alternatively, we keep the old base paths as renamed attributes (e.g. `mainstream_browse_base_path`).
```


But yup, I think we need to expand on what happens to the old base paths when we access the URL e.g GET `www.gov.uk/browse/something` or `www.gov.uk/api/content/browse/something`. 

We might need to add two content items to represent the [/browse](https://www.gov.uk/browse) and [/topic](https://www.gov.uk/topic) pages. They'd need a prefix routing rule for `/browse*` or `/topic*`, so that requests are redirected to the frontend app (which can fetch the correct taxon instead of the mainstream browse or topic content item). This would prevent us adding redirects from mainstream browse/topics pages to taxons (e.g. `/browse/education` => `/education`) but don't think it's an issue.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
benthorner: Really helpful, thanks!

Minor: we could inline the three examples in the appendix here, since this is where we introduce them.

I think it would also be helpful to clarify what these three things are for, as the docs we're linking to don't explain, and the examples are just examples. Why do we have these three navigation systems?

Related to that, I'm confused about whether this would lead to any user-facing changes on GOV.UK. Are we proposing to continue rendering existing mainstream browse and specialist topic pages as we are now, but using a different content item? This is relevant for the problem/background section because I'm not sure if we're trying to work towards only having one navigation system, or just combining the attributes of the existing ones.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
benthorner: I'm still confused, sorry. The point about `/browse` and `/topic` is definitely worth adding, though, as that's an unexpected edge case: `/browse` doesn't have a content item, and `/topic` is the root content item.

I think we generally try and avoid prefix routes where possible, so the app doesn't have to handle 404s itself. We should be able to use sub-routes on the taxon content item to keep all the old paths in place.

But it might be worth pausing this thread, as my confusion seems to stem from not fully understanding the background / problem, which we're discussing in [another thread](https://github.com/alphagov/govuk-rfcs/pull/119#discussion_r524295884).

rfc-119-merge-mainstream-browse-topics-and-taxons.md
benthorner: That's clearer, thanks!

Minor: I'd suggest "used" instead of "useful" everywhere. This will hopefully become clearer with [my other comment](https://github.com/alphagov/govuk-rfcs/pull/119#discussion_r524295884) around the purpose of these three things.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
theseanything: This is because the organisational hierarchy of specialist topics is different than taxons. 

For example, the specialist topic parent for 'motorways-major-roads' is 'transport', but in the taxonomy the parent for 'motorways-major-roads' might be ' driving-and-road-transport' (who's parent is 'transport'). This is needed to render pages like (https://www.gov.uk/topic/transport/motorways-major-roads)

Specialist topic links:
'motorways-major-roads' -> 'transport'

Taxon links:
'motorways-major-roads' -> 'driving-and-road-transport' -> 'transport'

Does that make sense?

benthorner: Oh, I see, yes. I guess it's kind of obvious: if there are different children, there will be different parents 🙄.

Now I appreciate the hierarchical distinction, I'm starting to worry about the end result, which could get quite confusing: we really are modelling three hierarchies in one object. I can see this leading to synchronisation issues where, for example, a taxon is moved to new parent, but it's mainstream browse and specialist topic hierarchies are left as-is. We could end up with quite a complex UI to manage the different levels of parents/children.

Since the aim of this RFC is to align these three things, I'm wondering if we could go further: could we just have one hierarchy (of taxons), and use flags to filter the hierarchy into a mainstream browse hierarchy (for navigation), and a specialist topic hierarchy (still not sure what this one is even for 😆)?

So, in your example, we could have:

'motorways-major-roads' (`browse: true`) -> 'driving-and-road-transport' (`browse: false`) -> 'transport' (`browse: true`)

I guess a disadvantage of this is that the rendering apps would have a harder time using the hierarchy, since they'd have to manually skip between levels. However, that seems reasonable, since the concern seems to be primarily a rendering one, and we need to load [the whole hierarchy](https://www.gov.uk/browse) then anyway.

Even if we don't go down this route, it would be helpful to explain why it's a rejected approach. Sorry, I know that's a fair chunk of effort. Let's chat if you want to go over anything!

rfc-119-merge-mainstream-browse-topics-and-taxons.md
theseanything: This work itself should not introduce any user facing changes - mainstream browse and topic pages will continue to exist. The content design work to ensure mainstream browse and specialist topics have equivalents in the taxonomy may cause user facing changes. But this RFC is contingent on that work already being done.

Cool - I'll add in my suggested changes with the examples + a note that this shouldn't introduce user facing changes. @WorkSutton do mind you addressing @benthorners comments about adding in more information about what these three navigation systems used for?

WorkSutton: @theseanything @benthorner Sure I can add some background on the 3 navigation systems based on my current understanding

rfc-119-merge-mainstream-browse-topics-and-taxons.md
theseanything: Thanks @benthorner! I'm interested in your comment

 > I can see this leading to synchronisation issues where, for example, a taxon is moved to new parent, but it's mainstream browse and specialist topic hierarchies are left as-is. 

So your example where changing a taxon parent would't effect the mainstream browse or specialist topic parents. I think this is desired, the taxon hierarchy isn't used for navigation. Having separate hierarchies prevents side effects e.g. we don't want to change the browse pages because we changed the taxon hierarchy for another reason. Can you think of any specific examples where this an issue?

> We could end up with quite a complex UI to manage the different levels of parents/children.

So currently, we already have multiple UIs and complexity in the publishing apps to deal with the three different categorisation types (mainstream browse, topic, taxons). Do you think the complexity is going to worse?

Thanks also for you suggestion about an alternative approach:

> use flags to filter the hierarchy into a mainstream browse hierarchy (for navigation), and a specialist topic hierarchy (still not sure what this one is even for 😆)?

So I think this assumes that siblings in mainstream browse hierarchy share the same parent in the taxon hierarchy - which isn't true. For example:

image

rfc-119-merge-mainstream-browse-topics-and-taxons.md
theseanything: @huwd have talked outside these comments. Think we agreed it may be helpful to have an examples of what the taxon content item looks after mainstream browse is added, but before specialist topic is. Is this something you could add @WorkSutton ?

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Proposal
+
+We propose merging the attributes of mainstream browse and specialist topic content items into **existing** taxon content items with the same `title`. (For clarity we aren't proposing the creation of new taxons or taxon tree.)
+
+Generally there are two main type of information we want to tranfer from mainstream browse and specialist topic to the equivalent taxon. The first is the links to content items. Currently content items (guidance pages, manuals etc..) do not contain links to mainstream browse or specialist topics. It's the reverse, mainstream browse pages and specialist topics have a list of content that's associated to them. This list we can migrate over to the taxon.
+
+Secondly, we'd carry across the relationships (links) between mainstream browse pages/specialist topics. This manifests as multiple hierarchies within the same taxons. This is because different relationships are useful for different things i.e. the current hierarchy for taxonomy is used for modelling emails subscriptions, but cannot be used for modelling navigation. So we'd bring over the mainstream browse hierarchy to serve that purpose.
```
kevindew, line 36: The email part of the topic taxonomy isn't particularly valued (it causes pain and confusion) - do we know if the current taxonomy shape provides any particular value? sub-hierarchies do sound like something complex and confusing so not sure if there's any effort we can make to at least try avoid some of them.

Is navigation the same as the mainstream browse hierarchy or is this a fourth hierarchy?

Since specialist is rather niche and legacy does that need its own hierarchy? It feels like we should deal with the problems of it being different to the topic taxonomy rather than defer them.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Problem
+
+We have to maintain three separate types of categorisation for content items - mainstream browse pages, specialist topics and taxons. This means the categories themselves and the content assigned to them may be different between each type. For example, the "Education and learning" mainstream browse page is the logical equivalent of the "Education, training and skills" taxon and should be the same, however they differ in name and in the set of content items they represent. There are also categories that only exist in some types and not others, for example the "Setting up a business" mainstream browse page has no equivalent taxon.
+
+Examples of equivalent content items:
+
+Mainstream browse page: https://www.gov.uk/api/content/browse/benefits/universal-credit
+
+Specialist topic: https://www.gov.uk/api/content/topic/benefits-credits/universal-credit
```
kevindew, line 22: I feel uneasy about work going into keeping these specialist topics as a distinct thing. It seems that this was one of the main things the topic taxonomy was meant to replace it feels like we should consolidate them and topic taxonomy pages rather than keep them. As I understand, outside of the hierachy, the main difference to the topic taxonomy is that we let departments curate the links of these. Could we consider how these could become a single page?

It seems a strange to me that we have https://www.gov.uk/topic/personal-tax/self-assessment and https://www.gov.uk/money/self-assessment that are essentially the same thing (and would become the same content item) but they show different things. 

If consolidation is too much work I'd have thought it be better to just have a link between the specialist topic page and the taxon than merge them as at least then they are clearly separate things.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,183 @@
...
+- `schema_name`
+- `updated_at`
+- `withdrawn_notice`
+- `publishing_request_id`
+
+This is because they don't cause conflict or the mainstream or specialist topic value becomes redundant.
+
+- `publishing_app`
+
+Currently references Collections Publisher for mainstream and specialist topics, however references Content Tagger for taxons. We propose that `publishing-app` remains Content Tagger as it still has the responsibility of creating and removing taxons, even though some sub properties (i.e. links and details) may be set by Collections Publisher. In the future, we could also look at consolidating the publishing functionality of Collections Publisher and Content Tagger.
```
kevindew, line 63: Editing the details by a different publishing application wouldn't be allowed at a technical level due to Publishing API path reservation. It also seems quite a smell that something isn't sufficiently isolated. 

It seems that you'd want the specialist topic links to be entirely managed with the taxon, otherwise I think we'd find this would forever have the smell of an odd coupling.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
kevindew: It'd be good to be clearer on the `browse/*` `topic/*` prefix routes as a consequence. I think I had this explained to me last time this RFC went around but I didn't pick up on it reading through this again until I read this thread.

I'd interpreted it as we'd have say `www.gov.uk/{taxon-base-path}` and thought that might render as a specialist topic, a taxon or a mainstream browse page - this made me a bit concerned about overloading of concepts.

I'd also suggest staying away from a second base_path in the details - that seems a smell that this is two distinct things with a relationship (a browse item coupled to a topic taxon) than the same thing in a different scenario.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,183 @@
...
+- Displaying metrics for `/browse/*` and `/topic/*` pages in Content Data
+- Signing up and email subscriptions per specialist topic
+- Finding browse links in the Knowledge Graph
+
+Once we are confident we are solely using the taxons, we can unpublish the existing mainstream browse and specialist topic content items.
+
+This work should not introduce any user facing changes.
+
+## Benefits
+This would allow us to ensure ongoing consistency of our "Information Architecture" and grouping of content, without the complexity of maintaining three document types or seperate sets of content items.
```
kevindew, line 175: Are there further benefits than this? I feel like this is probably solving some problems that someone is experiencing that aren't called out?

I'm wondering that whilst these things still exist as 3 separate items item on the frontend whether we're introducing limited benefits and instead risks and confusion. For example, someone might not understand that editing a taxon name actually changes a mainstream browse page and a specialist topic? Or that removing an item from the topic taxonomy could break a whole branch of a mainstream browse.

There's probably also some risks that should be called out with these

kevindew (commented):
Great, to see this moving again 🙌

rfc-119-merge-mainstream-browse-topics-and-taxons.md
benthorner: Oh wow, nice diagram! The criss-cross hierarchy is definitely a surprise, but if it's a product need, then I agree we need to support it and the additional links make sense. I do think we should explain more about the hierarchies in the background section, though, perhaps with a version of this diagram to support it. I appreciate that's a fair amount of effort. What I'm finding with this RFC is that there's actually quite a lot of hidden knowledge with these hierarchies, which makes it hard to understand the proposal.

The need for separate hierarchies does actually lead me to question the value we get from merging these three things: if the hierarchies are meant to deviate, then would it be clearer to continue representing them as three distinct things, despite some commonalities they may have? What problems are we solving by combining them? I can see there might be some value in not having to synchronise titles, for example. Could we also elaborate the problem section to explain more specifically what we're concerned about with divergence?

rfc-119-merge-mainstream-browse-topics-and-taxons.md
```diff
@@ -0,0 +1,183 @@
...
+
+Taxon: https://www.gov.uk/api/content/welfare/universal-credit
+
+There is currently ongoing work to ensure mainstream browse and specialist topics have equivalents in the taxonomy, i.e. mainstream browse and specialist topics become a subset of taxons. This is being done by Content Designers who are liaising with departments to come to agreement.
+
+Once this work is complete we want to maintain alignment and prevent divergence from taxons. We would have to put constraints in collections publisher and content tagger (or any future apps) to ensure that mainstream browse and specialist topics continue to be subsets of taxons. This would involve adding complex logic to the publishing system to deal with addition, changing or removal of these content items.
+
+## Proposal
+
+We propose merging the attributes of mainstream browse and specialist topic content items into **existing** taxon content items with the same `title`. (For clarity we aren't proposing the creation of new taxons or taxon tree.)
```
benthorner, line 32: I'm wondering how this would work for publishing apps tagging to these things. From other comments, I understand that mainstream browse and specialist topics will be a subset of the existing taxonomy. Once the merge has happened, how will we scope the total set of taxons so that a publisher can't accidentally tag to one that's not in the mainstream browse hierarchy, for example? (To be clear: this is where we're specifically asking them to "tag to mainstream browse categories" in the UI, so there's an expectation that the document will appear in certain places.)

samdub: I can think of two options here:
1. Hide topics from publishing apps that don't correspond to user-facing topic pages
2. Allow tagging that's more granular that our user-facing topics show, but roll up this tagging to first user-facing level.
e.g. If tag I tag a content item to 'Education > Schools > Maths' ...but Maths is not shown to users, the content would appear on the frontend tagged to Schools instead. Taxonomy pages already roll-up tagging in this way.

I'm inclined to option 1 to keep things simple for publishers. There may a case for slimming down the taxonomy considerably if we're not using the lower levels.

rfc-119-merge-mainstream-browse-topics-and-taxons.md
WorkSutton: ### Mainstream browse
- Currently the `primary` browse navigation of GOV.UK
- 2 level hierarchy containing ~ 150 browseable topoics on GOV.UK homepage
- sub-topics users see as curated lists of Mainstream guides
- users cannot subscribe to email updates

### Specialist topics
- Currently the `secondary` browse navigation on GOV.UK
- 2 level hierarchy containing ~ 410 browseable from department and organisation pages
- sub-topics users see as curated lists of Specialist guids
- users can subscribe to email updates 

### Taxonomy topics
- Very few users utilize this to navigate the site
- multi-level hierarchy of variable depth (1-5 levels) comprising ~ 1300 topics
- reachable from breadcrumbs and related links
- content of all types is tagged to every level of the hierarchy
- uncurated
- users can subscribe to email updates

benthorner: Thanks for the summary @WorkSutton. Definitely some new insights in there for me, and I reckon others would find it helpful for this to be incorporated in the RFC.

I'm not sure what `primary` and `secondary` mean, though. I can see that Mainstream Browse Categories are the ones listed on the homepage (so perhaps `primary` in the sense that they're the main way for users to browse GOV.UK content). Where do we surface the Specialist Topic pages that makes them `secondary`?

samdub: `primary` and `secondary` is here is a shorthand for the fact that in terms of navigational prominence, traffic, and effort - mainstream topics are currently prioritised.

Specialist topics are surfaced on organisation pages, related links and breadcrumbs.

bevanloon:
tldr; I'm broadly supportive of the direction this PR is going, but I don't believe we can give it the dedicated attention it deserves right now. We'll leave the PR open until Friday in case people have more comments.

Thanks for all the comments on this PR everyone. Richard and I had a good chat with the Explore team yesterday to discuss the work they are planning and doing around this PR.

It's clear that the technical work here is fairly complex and will touch on large parts of GOV.UK's ecosystem. It's also fair to say, I think, that there are some good discussion points that need to be ironed out before the PR is accepted. I don't think we have bandwidth on GOV.UK to tackle this properly at the moment. I think we need someone with deep experience of GOV.UK dedicated to the Explore team to guide this PR through, which at the moment, given the work on COVID and Brexit, we don't have.

We'll leave the PR open until Friday anyway, just in case people wish to comment more, but at that point we'll close it until we can dedicate the right people to the problem. The Explore team, in the meantime, will carry on with some front end prototyping and user research to further their understanding of the domain.

WorkSutton:
Closing for now. To be re-opened and reviewed when we are able to implement.

3 additional associated PR's for merge of information into this PR were also closed
[RFC-119: Add draft delivery plan](https://github.com/alphagov/govuk-rfcs/pull/130)
[RFC-119: Add some product background in Appendix A](https://github.com/alphagov/govuk-rfcs/pull/131)
[RFC-119: definitions of existing topic systems](https://github.com/alphagov/govuk-rfcs/pull/132)

PR #120, opened 2020-03-09
Update RFC 116 to reflect revised featured_attachments approach

kevindew:
This amends the RFC to resolve some misunderstandings with the approach
we'd be taking for this work. This change reflects a desire to no longer
pass HTML or Govspeak into content items for attachments and instead
pass in an id to reference which attachment should be rendered, the
rendering applications will then take this id and use a component to
render the resulting attachment.

### DIFF
--- a/rfc-116-store-attachment-data-in-content-items.md
+++ b/rfc-116-store-attachment-data-in-content-items.md
@@ -32,7 +32,7 @@ attachment metadata in their content items.  Whitehall only sends
 rendered HTML to the Publishing API.  Other publishing apps send a
 combination of rendered HTML and metadata.
 
-This inconistency restricts what we can do with attachments,
+This inconsistency restricts what we can do with attachments,
 particularly for Whitehall content.  For example, we cannot generate
 comprehensive [schema.org][] metadata for attachments: [see this
 comment][].
@@ -385,7 +385,7 @@ item.
 
 Manuals Publisher and Travel Advice Publisher have their own
 databases, rather than using Publishing API directly as a backing
-store, so they dont have this same problem.
+store, so they don't have this same problem.
 
 [schema.org]: http://schema.org/
 [see this comment]: https://github.com/alphagov/govuk_publishing_components/pull/1247#pullrequestreview-338008254
--- a/rfc-116-store-attachment-data-in-content-items.md
+++ b/rfc-116-store-attachment-data-in-content-items.md
@@ -169,24 +169,35 @@ I propose we do this:
     4. Make the field mandatory.
 
 1. Add a *mandatory* `details.featured_attachments` list to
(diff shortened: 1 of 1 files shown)


thomasleese:
> This change reflects a desire to no longer pass HTML or Govspeak into content items for attachments and instead pass in an id to reference which attachment should be rendered

Why is this the case? (I can imagine some good reasons why we would want it like that, but it might be good to be explicit.)

kevindew:
> > This change reflects a desire to no longer pass HTML or Govspeak into content items for attachments and instead pass in an id to reference which attachment should be rendered
> 
> Why is this the case? (I can imagine some good reasons why we would want it like that, but it might be good to be explicit.)

Sure, good point. I'll see if I can find an appropriate spot to expand on this a little. It did seem tricky to go into detail as it's quite tangential to the RFC, but I can see what that is good to capture.

kevindew:
@thomasleese I've added a list of reasons in.

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -169,24 +169,35 @@ I propose we do this:
...
+   an id of an entry in the details.attachments collection.
+
+   The reasons for preferring an id over the previous embedded HTML approach
+   are:
+
+   - less scope for inconsistency in content items - there isn't a risk
+     the HTML doesn't match the attachment data;
+   - removes a risk of unexpected content, since the HTML is a string
+     it can contain any content and may not necessarily reflect a single
+     attachment;
```
ChrisBAshton: I think this point can be removed, as implementing this RFC won't remove the possibility of unexpected content. (Someone may still try and put markdown referencing attachments in their Govspeak).

rfc-116-store-attachment-data-in-content-items.md
```diff
@@ -169,24 +169,35 @@ I propose we do this:
     4. Make the field mandatory.
 
 1. Add a *mandatory* `details.featured_attachments` list to
-   publication formats, which is an ordered list of govspeak or HTML
-   strings (using the `#/definitions/multiple_content_types` schema)
-   rendering attachments from `details.attachments`.
+   publication formats, which is an ordered list of strings that each reference
+   an id of an entry in the details.attachments collection.
```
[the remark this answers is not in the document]
kevindew: Thanks - have amended.

rfc-116-store-attachment-data-in-content-items.md
kevindew: As discussed IRL this will remove the chance of there being unexpected HTML in featured_attachments and doesn't refer to body.

ChrisBAshton: 👍 early morning review before coffee was consumed :)

PR #121, opened 2020-03-12
Remove content-audit-tool

nicholsj:
This removes mention of content-audit-tool as it's now retired.

### DIFF
--- a/rfc-095-long-term-future-apps.md
+++ b/rfc-095-long-term-future-apps.md
@@ -21,10 +21,7 @@ The purpose of doing this is to make clear:
 | calendars | Serves /bank-holidays and /when-do-the-clocks-change on GOV.UK | Merge into custom frontend |
 | ckanext-datagovuk | Extension for use with datagovuk_publish | Unknown |
 | collections-publisher | Publishes step by steps, /browse pages, and legacy /topic pages on GOV.UK | While topics and browse pages will be removed, this can still be the app for curating collections. We might move Whitehall's document collections into this app |
-<<<<<<< Updated upstream
-=======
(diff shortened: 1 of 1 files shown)


issyl0:
Hmmm, do we usually edit previous RFCs? If we wanted to substantially change the approach we'd make a new one. But IMO RFCs that list apps should be a point-in-time snapshot of what we decided to do then.

nicholsj:
I took from https://github.com/alphagov/govuk-rfcs#managing-action-plans that this was an "action plan" and so should be updated. But I'm not entirely clear on how RFCs work

thomasleese:
> Hmmm, do we usually edit previous RFCs? If we wanted to substantially change the approach we'd make a new one. But IMO RFCs that list apps should be a point-in-time snapshot of what we decided to do then.

Yeah, I had a similar opinion but then Jonathan pointed me https://github.com/alphagov/govuk-rfcs#managing-action-plans and I think this RFC is probably considered an action plan.

kevindew:
> Hmmm, do we usually edit previous RFCs? If we wanted to substantially change the approach we'd make a new one. But IMO RFCs that list apps should be a point-in-time snapshot of what we decided to do then.

Yep, you're right they should be a snapshot in time as much as possible for highest relevance of context.

I think this is fine though but we should probably leave a note after the table to explain the edit.

There's good chances that many ideas in this plan won't be fully realised and will change over time so I don't think we should try too hard to keep this up-to-date beyond minor changes like this.

Rather than remove the line we might want to just add a note after the retire bit that says "(Update 12 March 2020: Content Audit Tool was retired on <insert date>)"?

PR #122, opened 2020-03-31
Add a new RFC for previewing CSV attachments

benthorner:
Trello card: https://trello.com/c/Udwixrwu/1446-write-an-rfc-for-csv-preview

**Deadline for discussion: 2020-05-05**

👉 [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/csv-preview/rfc-122-csv-preview.md)

### DIFF
--- a/rfc-122-csv-preview.md
+++ b/rfc-122-csv-preview.md
@@ -1,0 +1,158 @@
+# CSV Preview
+
+[govuk-content-schemas]: https://github.com/alphagov/govuk-content-schemas
+[government-prefix]: https://trello.com/c/74KTVXBs/52-whitehall-relies-on-prefix-government-and-world-routes
+[router]: https://github.com/alphagov/router#backend-handler
+
+## Summary
+
+Move the CSV Preview feature out of Whitehall and into Government Frontend, so that it can be used for documents published by other apps, like Content Publisher. The new implementation will work by rendering a content item in a new `attachment_preview` format, which represents an attachment to preview.
+
+## Problem
+
+A user viewing a Whitehall document on GOV.UK can click a "View Online" link for any of its CSV file attachments [[example](https://www.gov.uk/government/publications/hmrc-exchange-rates-for-2020-monthly)]. This takes the user to a new page, currently rendered by Whitehall. The page contains:
+
+   - A 'header' section with details of the parent document (and navigation links)
+   - A 'body' section with a preview of the content of the CSV attachment
+
+This CSV Preview feature only works for CSV file attachments uploaded through Whitehall. We need it to work for CSV attachments published through Content Publisher, which is superseding Whitehall.
+
+## Proposal
+
+### Step 1
+
+Create a new `attachment_preview` format in [govuk-content-schemas][] to represent the preview page for an attachment. A publishing app will publish content items in this format for each CSV attachment to preview.
+
+```
+# formats/attachment_preview.jsonnet
+(import "shared/definitions/publishing_api_base.jsonnet") + {
+  document_type: "attachment_preview",
+  attachment: {
+    "$ref": "#/definitions/file_attachment_asset"
+  },
+  edition_links: {
+    parent: {
+      description: "The parent content item.",
+      maxItems: 1,
+    },
+    primary_publishing_organisation: {
+      maxItems: 1,
+    },
+  }
+}
+```
+
+- Using a content item is consistent with the one-to-one mapping of pages and content on GOV.UK. This should make it easier to debug issues with the feature, since it follows an existing pattern.
+
+- The schema includes a link to the parent document. Using an expanded link means we don't need to make an extra API call to the Content Store in order to show information about the parent document.
+
+- The schema includes a link to the primary publishing organisation. Including organisation info is [a longterm endeavour](https://github.com/alphagov/govuk-rfcs/pull/92), and will also make it a bit simpler to render the organisation logo in the header.
+
+- Using the publishing system means we can follow an existing approach for things like automatic redirects (in case the base path changes), unpublishing and anonymous preview ('fact check'). However, this entails an overhead of extra Publishing API operations to manage the separate content items - draft, publish, remove, etc. - with the associated risk they get out-of-sync e.g. due to transient network errors when calling the API.
+
+- Using a separate content item means we're not coupled to a particular rendering app. This provides a foundation for extending the preview feature in the future.
+
+### Step 2
+
+Re-implement the CSV Preview feature in Government Frontend.
+
+- Government Frontend currently renders all documents published by Whitehall, that have CSV file attachments. This makes it a relevant codebase to implement this feature in.
+
+- We expect the implementation to be small, isolated and not to grow/change much. Adding a new responsibility to Government Frontend seems like a reasonable compromise when compared to the [rejected proposals](#rejected-proposals).
+
+
+### Step 3
+
+Change Whitehall to use the new CSV Preview feature for newly published documents. We will need to use a different URL structure, since the current one is coupled to Whitehall.
+
+```
+# current URL example
+https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/856932/Management_information_-_further_education_and_skills_-_as_at_31_December_2019.csv/preview
+
+# proposed URL example
+https://www.gov.uk/government/remaining-parent-document-base-path/preview/asset-id
+```
+
+The `asset-id` should be the ID used by Asset Manager e.g. `597b098a759b743e0b759a96`, since it persists for the lifetime of the asset, even if the original file is replaced.
+
+### Step 4
+
+Change existing attachments (in Whitehall) to use the new CSV Preview. This will involve republishing all Whitehall documents that include a CSV preview link.
+
+Once this is done, we can remove some of the config for [the `assets` domain](https://github.com/alphagov/govuk-puppet/blob/51d7f7b648e8fde5aae3adfb774ec3bca6325bd8/modules/router/templates/assets_origin.conf.erb#L69-L74) that supported the legacy feature ([`gov.uk` routing](https://github.com/alphagov/government-frontend/blob/dcbc563030c470e6242d6432896d84feebb2138b/config/routes.rb#L14) remains). We can also delete the code for the legacy feature.
+
+> We may want to add redirects for the old preview URLs in Whitehall, if they continue to receive lots of traffic after all the 'View Online' links have updated.
+
+
+## Rejected proposals
+
+### Use a top-level prefix for the routing (e.g. `/preview/:base_path/:attachment_id`)
+
+This would involve registering a new `/preview` prefix route in [router][router], which would set the backend (rendering app) to Government Frontend for all paths beginning with that route.
+
+- This involves less effort than individually managing lower-level routes for each attachment preview. However, because we don't have any knowledge of these routes, we can't help the user if the URL they enter is invalid; this could happen when the asset is removed, or the parent document changes e.g. unpublished, redirected.
+
+- Using a top-level prefix for CSV Preview is at-odds with the behaviour of other top-level prefixes (`/info`, `/api/content`), since it only works for specific URLs, and doesn't apply to GOV.UK content in general.
+
+### Improve the feature in Whitehall (frontend)
+
+This would involve rewriting the existing CSV Preview feature so that it can work for documents published elsewhere, instead of being coupled to the attachment model in Whitehall.
+
+> We could make the feature work for attachments published by others apps, using the proposed URL structure, which gives us the parent document and the `asset-id` of attachment being previewed.
+
+- This would mean continued reliance of the Whitehall (frontend) codebase and the `/government` prefix route, which has been identified as [tech debt][government-prefix].
+
+- Not using specific routes for each attachment preview means we won't be able to do things like automatic redirects (if the base path of the parent document changes).
+
+### Using sub-routes of the parent document for the routing
+
+The Publishing API supports any number of `routes` for a content item. Normally, this will just be a single route, equal to the `base_path` of the content item. We could add an extra route or routes to support CSV Preview.
+
+> Any additional routes that exist (and which are not redirects) will use the same rendering app as the 'main' route. This fits with the proposal to re-implement CSV Preview in Government Frontend.
+
+> We could use [a prefix sub-route, or specific sub-routes for each attachment](#appendix-sub-routes). Using a prefix sub-route is less effort, but the behaviour is ambiguous if an attachment is not found or associated with the document.
+
+- This would have a lower implementation effort for individual publishing apps, as opposed to creating a separate content item for each attachment preview, which could get out-of-sync with the parent document.
+
+- Most of the lifecycle management, such as unpublishing and redirects, would be done by the Publishing API. However, we would need to address a number of issues with sub-routes in the Publishing API:
+
+  - Publishing API does not clean up sub-routes when they are no longer present in the content item. It _will_ handle redirects for them if the parent base path changes, but that's it.
+
+  - Publishing API does not propagate changes in document state to its sub-routes. For example, a remove/redirect on the parent document does nothing to its sub-routes.
+
+- It's surprising to see routes that aren't related to the actual content item being defined as part of the content item itself. This would be a new pattern for GOV.UK, which someone working on the feature would need to learn.
+
+### Implement the feature in Asset Manager
+
+- Asset Manager has direct access to the CSV attachment, whereas other apps need to download it. Since Asset Manager also 'owns' the attachment, it seems reasonable for it to support preview for attachments.
+
+- It would be hard to reproduce the 'header' section of the preview page, since Asset Manager does not store data about the parent document, and we have no precedent for an `assets` URL to contain such data, either.
+
+- Rendering a CSV preview page would break the convention of Asset Manager being an API-only app. We expect that getting Asset Manager to render pages would be a relatively large/complex solution to the problem of this RFC.
+
+### Implement the feature in a new app
+
+This is a variant of the proposal to reimplement the preview capability in Government Frontend. Instead, we would create a new app dedicated to previewing attachments.
+
+- Using a separate codebase would allow us to extend the preview feature to support other file types.
+- Creating a new app would also be a much larger change, and increase the complexity and support burden of GOV.UK.
+
+## Appendix: Sub-Routes
+
+- Publishing API payload with a prefix sub-route
+
+  ```
+  routes: [
+    { path: "#{document.base_path}/preview", type: "prefix" }
+  ]
+  ...
+  ```
+- Publishing API payload with specific sub-routes
+
+  ```
+  routes: [
+    "{ path: #{document.base_path}/preview/#{attachment1.asset_id}", type: "exact" },
+    "{ path: #{document.base_path}/preview/#{attachment2.asset_id}", type: "exact" },
+  ]
+  ...
+  ```


rfc-120-csv-preview.md
```diff
@@ -0,0 +1,55 @@
+# CSV Preview
+
+## Summary
+
+rove the 'CSV Preview' feature out of Whitehall and into Government Frontend, changing the structure of the associated URLs so they can also be used with documents published by Content Publisher.
```
bevanloon: sorry - pedantic comment of the day `r` should be `M`

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,55 @@
...
+
+```
+# current URL example
+https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/856932/Management_information_-_further_education_and_skills_-_as_at_31_December_2019.csv/preview
+
+# proposed URL example
+https://www.gov.uk/government/remaining-parent-document-base-path/csv-preview/asset-id
+```
+
+> Note: the `asset-id` should match the `id` of an asset in the `attachments` array, as defined by [RFC 116][rfc-116]. The `attachments` array contains the URL needed to download the asset in order to preview it.
```
barrucadu: Currently the `id` is used (or is planned to be used) for the rendering app to fetch the metadata of the featured attachments, which means changing the `id` in a new edition is fine and doesn't break anything.  This is also kind of the case for content IDs: in principle we could change the content ID of a document and, as long as we updated the `links` of everything which referred to it, everything would just work.

This proposed change means we can no longer change the `id`, as public URLs will stop working.  This is fine, but something to be aware of going forwards, as I could see it potentially catching people out.

rfc-120-csv-preview.md
benthorner: But I wanted to "rove" it 😆

bevanloon: move_it

rfc-120-csv-preview.md
benthorner: Thanks for comments @barrucadu - this is a good point and something I hadn't thought of. After comparing two potential approaches, I can see arguments both ways:

   - **Using the `asset-id`** (as [badly defined in Asset Manager](https://github.com/alphagov/asset-manager/tree/7cd3d543e3b44ba3bcb276d8c9aef11eadd3bb1a#create-an-asset) - it's just the string on the end). This would mean we match the behaviour of the asset (replace/redirect), at the expense of an extra API query to get the asset URL.

   - **Using the `attachment-id`** (as per RFC 116). This saves an API query, since the payload contains the asset URL. Changing the ID (e.g. by replacing the file) would cause any external links to 404, unlike the `asset-id` approach.

Since we're not using an `assets` URL for CSV Preview, it seems OK to have a different behaviour. In both cases, a deleted asset would cause a 404. Perhaps a general approach to address this (hopefully infrequent) concern is to unblock the user by showing a more helpful error page, with a link (or just an automatic redirect) to the parent document.

I'll make a note of this to consider when we re-implement. 

Do you think it should be included in the RFC?

(One more minor advantage of the `attachment-id` approach is that it's more likely to be human-readable, at least for attachments published by Content Publisher.)

barrucadu: :+1:

I don't think this needs adding to the RFC, as I don't think rendering in Government Frontend but still querying Asset Manager is a good idea, it's almost the worst of both worlds!

benthorner:
@thomasleese thanks for the question. I've been a bit terse and used the phrase "deprecate/delete the feature" to mean the following:

- Republish all docs to use the new CSV Preview feature
- Remove all support for the feature in Whitehall (wait for cache clear, reindex)
- Remove all routing/redirects to the legacy feature (wait for cache clear, reindex)

In doing this, we acknowledge that any saved links to CSV previews will break, which seems reasonable as it's a convenience feature, not actual content.

I can merge this into the RFC, to make it more explicit. Before I do that, do you have any comments on this approach i.e. something to do differently?

thomasleese:
> In doing this, we acknowledge that any saved links to CSV previews will break, which seems reasonable as it's a convenience feature, not actual content.

We could probably put in some redirection controller in place in Whitehall to support this. It does feel a bit cleaner to not break existing URLs, but maybe it's not worth it...

> I can merge this into the RFC, to make it more explicit. Before I do that, do you have any comments on this approach i.e. something to do differently?

I think it is worth being a bit more explicit about the migration path in the RFC. What you've outlined above makes sense to me, it's the sort of thing I had in mind as well. I just wanted to be sure we were committing to removing the old functionality from Whitehall as well as implementing the new functionality.

kevindew:
This sounds good - I agree it'd be great to get this out of Whitehall. My only comments are regarding the URL structure.

The proposed structure would require the routes to be registered with Router unlike the current approach which uses the Whitehall `/government` prefix route. I seem to remember early conversations about this feature involving setting-up a `/preview` or `/csv-preview` prefix route so not sure if this something considered or not as part of the RFC. Options I see are:

- Have the content item register additional routes for any previewable CSVs, which would add additional complexity to Content Items but does leave them in control of the routes and probably not too big a burden since they need to set preview_url fields
- Have the content item itself define itself as a prefix route in case it has any CSVs, this doesn't seem ideal as we'd probably have to do this to most of GOV.UK's content
- Set a different prefix route for supporting this feature.

There is a part of me that wonders, based off Michaels comment if you want to have the asset manager id and filename still so that you can maintain parity with that interface, which would allow maintaining redirects.

Other minor thing I don't think you need to refer to this as csv-preview in the URL. I know the scope of this feature is just CSV but I think with or without `csv-` in the path we're going to have perform the same checks that the file is a CSV and having that in the path name will just make it more inconvenient if we decide to add another preview type in future.

benthorner:
Thanks for pointing out the rather large flaw in my plan @kevindew 🙉. For the interested reader: the problem with the suggested URL (`/government/:base_path/preview/:attachment_id`) is that `/government` is actually a prefix route to **whitehall-frontend**, not government-frontend.

In order to proceed, we need to agree on how the routing and the URL will work together. There are a few options, which I've expanded on from the suggestions from @kevindew. Before talking about those, it's worth explaining a few quirks about "sub-routes" in the Publishing API.

## Sub-Routes

The Publishing API supports any number of `routes` for a content item. Normally, this will just be a single route, equal to the `base_path` of the content item. Any additional routes that exist (and which are not redirects) will **initially** have the same routing behaviour as the 'main' route. There are problems with using "sub-routes" (`/base_path/other-path`) for CSV Preview.

- Publishing API does not clean up sub-routes when they are no longer present in the content item. It _will_ handle redirects for them if the parent base path changes, but that's it.

- Publishing API does not propagate changes in document state to its sub-routes. For example, a remove/redirect on the parent document does nothing to its sub-routes.

Depending on the option we choose to implement CSV Preview, we will want to fix these issues with the Publishing API, so that sub-routes are treated similarly to the parent document. One way to fix the absence of a sub-route would be to automatically redirect it to the parent base path.

## Options

### 1. Use a top-level prefix (e.g. `/preview/:base_path/:attachment_id`)

- ➕  Easy to configure once, and setup a wildcard route to handle it in goverment-frontend.
- ➖  At-odds with the behaviour of other top-level prefixes (`/info`, `/api/content`), since it only works for specific URLs, and doesn't apply to GOV.UK content in general.
- ➖ Wouldn't automatically redirect if the base path of the parent document changed.
- ➖ Wouldn't automatically redirect if the individual attachment IDs changed.
- ➖ Could/should need to check that the attachment being previewed is actually associated with the document in the `:base_path`. (Could be misleading otherwise.)

### 2. Improve the feature in whitehall-frontend

This would involve rewriting the existing CSV Preview feature so that it can work for documents published elsewhere, instead of being coupled to the attachment model in Whitehall.

- ➕ No changes needed to support the routing.
- ➖ Continued reliance on the `/government` prefix route. (Talking with @kevindew, the use of this prefix is not ideal on GOVernment.UK, and we should try to move away from it.)
- ➖ Continued reliance on the Whitehall codebase.
- ➖ Could/should need to check that the attachment being previewed is actually associated with the document in the `:base_path`. (Could be misleading otherwise.)
- ➖ Wouldn't automatically redirect if the individual attachment IDs changed. It would be dubious to add this behaviour in general for prefix routes.
- ➖ Wouldn't automatically redirect if the base path of the parent document changed.

### 3. Use a prefix sub-route (`/government/:base_path/preview/*`)

- ➕ Easy to configure once for each document.
- ➕ We get automatic redirects for free, if the parent's base path changes.
- ➕ Wouldn't automatically redirect if the individual attachment IDs changed. We could still support this behaviour in the rendering app (if the attachment isn't found/previewable).
- ➖ Could/should only exist if the document has attachments that are previewable. We could add code to make this happen, or just make sure the N/A behaviour is well-defined.
- ➖ Could/should need to check that the attachment being previewed is actually associated with the document in the `:base_path`. (Could be misleading otherwise.)

### 4. Use specific sub-routes (`/government/:base_path/preview/:attachment_id`)

- ➕ We get automatic redirects for free, if the parent's base path changes.
- ➕ We could setup automatic redirects for the individual preview paths, if the attachment ID changes. This is mentioned in the discussion about sub-routes ☝️ 
- ➖ It's surprising to see routes that aren't related to the actual content item being defined as part of the content item itself. This seems like it's a new pattern for GOV.UK.

### 5. Use a separate content item (e.g. `attachment_preview`)

This would involve defining a new content schema to represent an attachment preview page on GOV.UK. The content item would have a `parent` linkable to the parent document, and some minimal metadata about the attachment, in order to get it from Asset Manager.

- ➕ More consistent with the normal content-to-page relationship on GOV.UK.
- ➕ Using a (expanded) `parent` link means we don't need to make an extra API call to the Content Store in order to show information about the parent document.
- ➕ We get automatic redirects for free, if the parent's base path changes.
- ➕ We get automatic redirects for free, if the full base path changes.
- ➕ The URL isn't coupled to the ID used [in the attachments of the parent document](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-116-store-attachment-data-in-content-items.md).
- ➖ Complexity of managing the separate lifecycles of these additional content items: draft, publish, withdraw, remove. It's possible the content items will get out-of-sync.
- ➖ Overhead of additional API calls to manage the separate content item lifecycles.

## Other thoughts

One other issue with the proposed URL is that, depending on the `:attachment_id`, it may end in an extension like `.csv`, which is at odds with the response type and may cause unexpected issues. We could instead use the [asset_id](https://github.com/alphagov/asset-manager/tree/7cd3d543e3b44ba3bcb276d8c9aef11eadd3bb1a#create-an-asset), and make this part of [the attachment schema](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-116-store-attachment-data-in-content-items.md) (as a new `asset_manager_id` field, which could be specific to file attachments):

- ➕ The `asset_id` is more stable than the `attachment_id`, since the latter is intended for use in markdown snippets and publishing apps may allow the user to change it.

> Even though the `asset_id` is part of the URL for assets in Asset Manager, this isn't true for (non-file) attachments in general; using the URL to find attachments could lead to issues.

## Next steps

We should reach a decision about the above:

- Which option should we go with to make the routing work? (My personal preference is either 4 or 5.)
- Are we comfortable adding `asset_manager_id` to file attachment metadata?

It may be easier if we gather interested parties to talk this through. Once we have a decision, I can write up a migration plan, as @thomasleese suggested 👍.

 (Thanks for reading this far down 🏅.)

kevindew:
Thanks for the write up Ben - and many apologies for pointing out the proverbial spanner in the works. 

As for the questions:

> Which option should we go with to make the routing work? (My personal preference is either 4 or 5.)

I don't think there is a clear slam dunk proposal as all come with pros and cons. My order of preferences would be:

Option 5 - this one seems to have a bunch of consistency with HTML attachments so, despite it being rather complex, it feels like it's not heading down either a surprising or a new path. It does feel like it could have some pain if things slip out of sync.
Option 4 - Sidesteps any sync concerns of option 5 but working closely with routes feels a bit of a minefield.
Option 1 - Most similar to current approach, it feels like a reasonable approach to get things out but leaves this feature as something rather special. It also carries with it [the confusion that a prefix route brings to any content store requests](https://github.com/alphagov/content-store/issues/324).
Option 3 - Having prefixes inside many content items feel like lots of wasted and confusing routes.
Option 2 - The infamous /government prefix will never die.

> Are we comfortable adding asset_manager_id to file attachment metadata?

I am, seems quite reasonable since this is continuous data

benthorner:
Thanks for the handy summary (of my summary) @kevindew! I'm going to try and break it down to some decision points, to see if we can reach a conclusion before I try and write it up.

### If we don't care about redirects

Option 2 is arguably the simplest change we can make.

### If we care about ditching Whitehall (now) and don't care about redirects

Option 1 is arguably the simplest change we can make.

### If we care about ditching Whitehall (now) and care about redirects

Option 4 has the advantage of fixing broken behaviour, but locks the rendering app. It does represent the coupling between the content item and the preview capability, so they can't be out of sync. Most of the implementation would be in Publishing API. What do you anticipate being the difficulties of this approach (aside from the things to fix)?

### If we want to build a foundation for more advanced preview capability

Option 5 is architecturally consistent with the way this feature appears to the user, in that each CSV preview is a distinct page on GOV.UK, represented by a content item. It also allows the rendering app to differ. It would require quite a lot more work to implement twice across Whitehall and Content Publisher, and could indeed lead to sync issues.

## Next Steps

- @kevindew does that way of organising things suggest a clear winner for you? I realise this depends on the team implementing it. Let's assume it's the same team, with the same goal of "just making this feature 'work' for stuff not published in Whitehall".

- @thomasleese @barrucadu it would be really helpful to have your input on the larger question here. I'd like to come to a provisional agreement here before I spend the time lovingly rewriting the RFC. @barrucadu could you also give your input on this change?

> Are we comfortable adding asset_manager_id to file attachment metadata?

barrucadu:
> Which option should we go with to make the routing work? (My personal preference is either 4 or 5.)

I think my order of preferences is:

- **5**: this feels like the best end-state, really, regardless of what approach we take for now.  It's consistent with how HTML attachments work and with the idea of there being a more-or-less 1-1 mapping between pages and content items.  The downside is that it's by far the most work to implement...

- **4**: it's a bit weird to have what are essentially sub-routes in the content item, but not *that* weird.  For example, we don't have separate content items for guide parts, all their content is in the content item for the guide itself, despite being on different URLs: https://www.gov.uk/universal-credit/how-youre-paid

- **3**: this feels like a worse, but easier to implement, version of **4**.

- **1** and **2**: I don't have a strong preference between either of these.  We either rely on the existing top-level prefix route, or make a new one.  Not great.

> `asset_manager_id`

It feels a bit strange to have two IDs for the same thing, but making everything just use the asset-manager ID could be awkward.  For example, if an asset is replaced, the asset manager ID will change, but if you're using some more user-friendly ID for the attachment ID, the markup in the content won't need changing.

I'm fine with adding this to the schema.

thomasleese:
> * @thomasleese @barrucadu it would be really helpful to have your input on the larger question here. I'd like to come to a provisional agreement here before I spend the time lovingly rewriting the RFC.

I don't think I have anything different to add really, I agree with @kevindew and @barrucadu in that my preferred option would be number 5 for the reasons outlined in their responses. From that number 4 would be my second favourite.

kevindew:
> Thanks for the handy summary (of my summary) @kevindew! I'm going to try and break it down to some decision points, to see if we can reach a conclusion before I try and write it up.
> 
> ### If we just don't care
> Option 2 is arguably the simplest change we can make.

Yup yup, if we're happy to build features that couple us deeper into tech debt: https://trello.com/c/74KTVXBs/52-whitehall-relies-on-prefix-government-and-world-routes

> ### If we care about ditching Whitehall (now)
> Option 1 is arguably the simplest change we can make.

The caveat for this is a bit wider than ditching Whitehall. The `/government` prefix route isn't a necessary part of it, so it's more about ditching one of the wilder features of Whitehall frontend. Then next level a good GOV.UK outcome would be no Whitehall frontend, and thus no prefix route, but still have the publishing app.

So to summarise this ramble; this is more focused on not leveraging something bad. 

> 
> ### If we care about redirects
> Option 4 has the advantage of fixing broken behaviour, but locks the rendering app. It does represent the coupling between the content item and the preview capability, so they can't be out of sync. Most of the implementation would be in Publishing API. What do you anticipate being the difficulties of this approach (aside from the things to fix)?

I'd have thought most of the implementation for this would be in publishing apps where they need to send additional information (routes) and manage the life cycle of those routes. It's the lifecycle management that seems a challenge if we want to maintain any redirects or information when items are moved or deleted.

> ### If we want to build a foundation for more advanced preview capability
> Option 5 is architecturally consistent with the way this feature appears to the user, in that each CSV preview is a distinct page on GOV.UK, represented by a content item. It also allows the rendering app to differ. It would require quite a lot more work to implement twice across Whitehall and Content Publisher, and could indeed lead to sync issues.
>
> ## Next Steps
> * @kevindew does that way of organising things suggest a clear winner for you? I realise this depends on the team implementing it. Let's assume it's the same team, with the same goal of "just making this feature 'work' for stuff not published in Whitehall".

That sounds good - I've added some comments. I'm with 5 for the winner.

benthorner:
Thanks for your comments everyone! I've updated the document to incorporate all of them. The whole document has changed quite a bit, due to the comments, so it's worth reading again. 📚

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+### Step 3
+
+Change Whitehall to use the new CSV Preview feture for newly published documents. We will need to use a different URL structure, since the current one is coupled to Whitehall.
+
+```
+# current URL example
+https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/856932/Management_information_-_further_education_and_skills_-_as_at_31_December_2019.csv/preview
+
+# proposed URL example
+https://www.gov.uk/government/remaining-parent-document-base-path/preview/asset-id
```
ChrisBAshton, line 68: Could you explain here why the `remaining-parent-document-base-path` portion of the URL is necessary? I didn't get my answer until the end of the document:

> - It would be hard to reproduce the 'header' section of the preview page, since Asset Manager does not store data about the parent document

Have we run this proposal past publishers? I think we may have made an assumption that using the parent document slug and asset ID is acceptable in lieu of a filename, but publishers may not be happy that the filename is no longer reflected in the URL. I don't see this explained/justified in the RFC at the moment.

I can also imagine cases where the publisher might not want the asset to be coupled to a particular document. We may have two different pages referring to the same asset - under the current proposal, one of their slugs would be in the asset preview URL, and one would not.

The 'header' section certainly complicates things, but we can deduce what should go in there by looking at the parent document content item - I don't think the document necessarily needs to be part of the asset preview URL.

For feature parity with the current implementation, I think I would opt for a structure like this: `https://www.gov.uk/government/uploads/preview/asset-id/asset-filename.csv`. If the publisher renames the file (either explicitly or indirectly through a file replacement), this should delegate to the standard redirect patterns we have in publishing API / content store.

Keeping it under `/government/uploads/preview` would involve [adding another regex entry to govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/137e774faa1eb3ff1122c6e71645bcfcfef7815f/hieradata/common.yaml#L372-L377), but I don't see a particular disadvantage to that.

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+The `asset-id` should be the ID used by Asset Manager e.g. `597b098a759b743e0b759a96`, since it persists for the lifetime of the asset, even if the original file is replaced.
+
+
+### Step 4
+
+Change existing attachments (in Whitehall) to use the new CSV Preview. This will involve republishing all Whitehall documents that include a CSV preview link.
+
+Once this is done, we can remove some of the config for [the `assets` domain](https://github.com/alphagov/govuk-puppet/blob/51d7f7b648e8fde5aae3adfb774ec3bca6325bd8/modules/router/templates/assets_origin.conf.erb#L69-L74) that supported the legacy feature ([`gov.uk` routing](https://github.com/alphagov/government-frontend/blob/dcbc563030c470e6242d6432896d84feebb2138b/config/routes.rb#L14) remains). We can also delete the code for the legacy feature.
+
+> We may want to add redirects for the old preview URLs in Whitehall, if they continue to receive lots of traffic after all the 'View Online' links have updated.
```
ChrisBAshton, line 80: Might be worth adding a line saying we can use Google Analytics to help inform this decision. Or would it be CDN logs? 🤔

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+
+
+## Rejected proposals
+
+### Use a top-level prefix for the routing (e.g. `/preview/:base_path/:attachment_id`)
+
+This would involve registering a new `/preview` prefix route in [router][router], which would set the backend (rendering app) to Government Frontend for all paths beginning with that route.
+
+- This involves less effort than individually managing lower-level routes for each attachment preview. However, because we don't have any knowledge of these routes, we can't help the user if the URL they enter is invalid; this could happen when the asset is removed, or the parent document changes e.g. unpublished, redirected.
```
ChrisBAshton, line 89: Could you elaborate on how/why the current proposal _would_ allow us to help the user if the URL they enter is invalid?

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+
+- Using a top-level prefix for CSV Preview is at-odds with the behaviour of other top-level prefixes (`/info`, `/api/content`), since it only works for specific URLs, and doesn't apply to GOV.UK content in general.
+
+### Improve the feature in Whitehall (frontend)
+
+This would involve rewriting the existing CSV Preview feature so that it can work for documents published elsewhere, instead of being coupled to the attachment model in Whitehall.
+
+> We could make the feature work for attachments published by others apps, using the proposed URL structure, which gives us the parent document and the `asset-id` of attachment being previewed.
+
+- This would mean continued reliance of the Whitehall (frontend) codebase and the `/government` prefix route, which has been identified as [tech debt][government-prefix].
```
ChrisBAshton, line 99: I'm a bit confused by this, as your proposed URL structure is `https://www.gov.uk/government/remaining-parent-document-base-path/preview/asset-id` (i.e. it still has a `/government` prefix).

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+
+- Not using specific routes for each attachment preview means we won't be able to do think like automatic redirects (if the base path of the parent document changes).
```
ChrisBAshton, line 101:
```suggestion
- Not using specific routes for each attachment preview means we won't be able to do things like automatic redirects (if the base path of the parent document changes).
```

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Using sub-routes of the parent document for the routing
+
+The Publishing API supports any number of `routes` for a content item. Normally, this will just be a single route, equal to the `base_path` of the content item. We could add an extra route or routes to support CSV Preview.
+
+> Any additional routes that exist (and which are not redirects) will use the same rendering app as the 'main' route. This fits with the proposal to re-implement CSV Preview in Government Frontend.
+
+> We could use a prefix sub-route (`.../*`), or specific sub-routes for each attachment. Using a prefix sub-route is less effort, but the behaviour is ambiguous if an attachment is not found or associated with the document.
```
ChrisBAshton, line 109: I'm finding this a little difficult to parse - could you provide example routes for lines 105-109?

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+
+- This would have a lower implementation effort for individual publishing apps, as opposed to creating a separate content item for each attachment preview, which could get out-of-sync with the parent document.
+
+- Most of the lifecycle management, such as unpublishing and redirects, would be done by the Publishing API. However, we would need to address a number of issues with sub-routes in the Publishing API.
```
ChrisBAshton, line 113:
```suggestion
- Most of the lifecycle management, such as unpublishing and redirects, would be done by the Publishing API. However, we would need to address a number of issues with sub-routes in the Publishing API:
```

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+
+  - Publishing API does not clean up sub-routes when they are no longer present in the content item. It _will_ handle redirects for them if the parent base path changes, but that's it.
+
+  - Publishing API does not propagate changes in document state to its sub-routes. For example, a remove/redirect on the parent document does nothing to its sub-routes.
+
+- It's surprising to see routes that aren't related to the actual content item being defined as part of the content item itself. This would be a new pattern for GOV.UK, which someone working on the feature needs to learn.
```
ChrisBAshton, line 119:
```suggestion
- It's surprising to see routes that aren't related to the actual content item being defined as part of the content item itself. This would be a new pattern for GOV.UK, which someone working on the feature would need to learn.
```

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Step 1
+
+Create a new `attachment_preview` format in [govuk-content-schemas][] to represent the preview page for an attachment. A publishing app will publish content items in this format for each CSV attachment to preview.
+
+```
+# formats/attachment_preview.jsonnet
+(import "shared/default_format.jsonnet") + {
+  document_type: "attachment_preview",
+  attachment: {
```
kevindew, line 30: Should this belong in the details hash?

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+   - A 'header' section with details of the parent document (and navigation links)
+   - A 'body' section with a preview of the content of the CSV attachment
+
+This CSV Preview feature only works for CSV file attachments uploaded through Whitehall. We need it to work for CSV attachments published through Content Publisher, which is superseding Whitehall.
+
+## Proposal
+
+### Step 1
+
+Create a new `attachment_preview` format in [govuk-content-schemas][] to represent the preview page for an attachment. A publishing app will publish content items in this format for each CSV attachment to preview.
```
kevindew, line 24: Do we want to consider specifying what we recommend using for update_type? It's probably something of an implementation detail just one of the questions that comes to my mind when thinking about it as a content item, similarly for public_updated_at timestamp. Sorry it this is a bit of tangential comment, maybe can just copy HTML attachment approaches.

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+      description: "The parent content item.",
+      maxItems: 1,
+    },
+  }
+}
+```
+
+- Using a content item is consistent with the one-to-one mapping of pages and content on GOV.UK. This should make it easier to debug issues with the feature, since it follows an existing pattern.
+
+- The schema includes a link to the parent document. Using an expanded link means we don't need to make an extra API call to the Content Store in order to show information about the parent document.
```
kevindew, line 44: We will also include a link to the primary organisation? I can't remember if some things fully expect that or not. I suppose we have precedent in html publications to follow.

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+Re-implement the CSV Preview feature in Government Frontend.
+
+- Government Frontend currently renders all documents published by Whitehall, that have CSV file attachments. This makes it a relevant codebase to implement this feature in.
+
+- We expect the implementation to be small, isolated and not to grow/change much. Adding a new responsibility to Government Frontend seems like a reasonable compromise when compared to the [rejected proposals](#rejected-proposals).
+
+
+### Step 3
+
+Change Whitehall to use the new CSV Preview feture for newly published documents. We will need to use a different URL structure, since the current one is coupled to Whitehall.
```
kevindew, line 61: s/feture/feature

kevindew:
Oh minor - the file is numbered as 120 whereas this is [RFC 122](https://github.com/alphagov/govuk-rfcs/blame/45e09c088bfde6c1d4dfb68f9bc1c635f63c9a39/README.md#L10)

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+Move the CSV Preview feature out of Whitehall and into Government Frontend, so that it can be used for documents published by other apps, like Content Publisher. The new implementation will work by rendering a content item in a new `attachment_preview` format, which represents an attachment to preview.
+
+## Problem
+
+A user viewing a Whitehall document on GOV.UK can click a "View Online" link for any of its CSV file attachments. This takes the user to a new page, currently rendered by Whitehall. The page contains:
+
+   - A 'header' section with details of the parent document (and navigation links)
+   - A 'body' section with a preview of the content of the CSV attachment
+
+This CSV Preview feature only works for CSV file attachments uploaded through Whitehall. We need it to work for CSV attachments published through Content Publisher, which is superseding Whitehall.
```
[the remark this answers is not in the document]
benthorner: Sure, I've added one a bit further up, which should be stable.

rfc-120-csv-preview.md
```diff
@@ -0,0 +1,134 @@
...
+    },
+  }
+}
+```
+
+- Using a content item is consistent with the one-to-one mapping of pages and content on GOV.UK. This should make it easier to debug issues with the feature, since it follows an existing pattern.
+
+- The schema includes a link to the parent document. Using an expanded link means we don't need to make an extra API call to the Content Store in order to show information about the parent document.
+
+- Using the publishing system means we can follow an existing approach for things like automatic redirects (in case the base path changes) and anonymous preview ('fact check'). However, this entails an overhead of extra API calls to manage the separate content items, with the risk they get out-of-sync.
```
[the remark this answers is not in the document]
benthorner: I've corrected this to say "Publishing API calls', and reworded this a bit to try and explain my concern, although we can always hope future developers will write flawless code 😉.

rfc-120-csv-preview.md
benthorner: The nice thing about this proposal is that it decouples the URL from the implementation; we can indeed use the expanded `parent` link to help us render the page header.

I think the general choice of URL is subjective, and not something we need agreement from publishers on. Publishers aren't able to modify the URLs currently, and I'm not aware of any scenario where this has been a real issue.

If we were starting from scratch today, I don't think we would have come up with `/government/uploads/preview/asset-id/asset-filename.csv`. There are a few objective reasons for not using this particular structure:

- It's semantically coupled to Whitehall: `/government/uploads` is a Whitehall thing, which we're trying to move away from, with the `assets.` domain.

- It's less stable: the URL contains two IDs, which could both change. Only one ID is required, of which the `asset-id` is more stable.

- It's more surprising: [these kinds of ad-hoc regex overrides](https://github.com/alphagov/govuk-puppet/blob/137e774faa1eb3ff1122c6e71645bcfcfef7815f/hieradata/common.yaml#L372-L377) are part of what makes GOV.UK routing (even) hard(er) to understand.

- It's inconsistent: the proposal we have here is to treat CSV Preview like any other GOV.UK content, so it should have a normal-looking URL structure for content.

The proposed URL structure was self-explanatory to me, with the exception of the choice of `/preview/`. We do have a precedent for it: both HTML attachments and manual sections have are 'child' documents that (redundantly) have the parent in the base path.

Does that help? I can add some or all of this into the RFC if need be.

rfc-120-csv-preview.md
benthorner: I think we can leave that as a detail for the person doing it. Either should work, although CDN logs would be more reliable these days.

rfc-120-csv-preview.md
benthorner: There's some discussion of that with the current proposal [here](https://github.com/alphagov/govuk-rfcs/pull/122/files#diff-4e2dd4712e73fb35ede7ffcaf64a7e70R49):

> Using the publishing system means we can follow an existing approach for things like automatic redirects (in case the base path changes), unpublishing and anonymous preview ('fact check')

rfc-120-csv-preview.md
benthorner: Yes, that's right. The difference is that, with this rejected proposal, we'd be relying on the `/government` prefix to do the routing. With the preferred proposal, we'll be registering a more specific route for each content item, which will override the behaviour of the prefix - this is the trie mechanism of the Router in action.

rfc-120-csv-preview.md
benthorner: Fixed 👍

rfc-120-csv-preview.md
benthorner: Yeah, I see what you mean. The URL structure is unchanged for this proposal; it's the routing mechanism that's in discussion here. I've added a (link to) appendix that shows what doing using each of these mechanisms would look like.

rfc-120-csv-preview.md
benthorner: Fixed 👍

rfc-120-csv-preview.md
benthorner: I was wondering about that. It seemed a bit weird/verbose, since the attachment pretty much _is_ the details. Are you thinking it's more conventional?

rfc-120-csv-preview.md
benthorner: I don't _think_ we _need_ it. However, I found a couple of reasons to include it (updated).

> Including organisation info is [a longterm endeavour](https://github.com/alphagov/govuk-rfcs/pull/92), and will also make it a bit simpler to render the organisation logo in the header.

rfc-120-csv-preview.md
benthorner: Tricky. I just realised I'd inherited from `default_format`, but we actually want `publishing_api_base`, since we'll (at least) want `access_limited` stuff.

So, while I've fixed that, I do think `update_type`, `change_note`, and `public_updated_at` are more of an implementation detail.

rfc-120-csv-preview.md
benthorner: Fixed 👍

benthorner:
@ChrisBAshton @kevindew thanks for your comments. Phew, the scrolling effort alone is tiring on this one! Anyway, I think I've responded/address everything.

Could you take another look?

rfc-120-csv-preview.md
ChrisBAshton: > Publishers aren't able to modify the URLs currently, and I'm not aware of any scenario where this has been a real issue.

We've had tickets on 2nd line asking to rename assets, e.g. `court-hearing-22-april` => `court-hearing-23-april`, which seems a reasonable request if they originally made a mistake when naming the asset. I like that under your proposal this will never be an issue, as the filename is never exposed.

I think it would be worth adding an explanation for the proposed URL structure, so that we have something we can point to in the future, and so that it shows you've thought about it. This should include an explanation for the `remaining-parent-document-base-path` portion of the URL - as we've not coupled files to their parent document in the past - and explain why we're not just using the `parent` link in the content ID to get that information for the header.

rfc-120-csv-preview.md
ChrisBAshton: I still don't really get it, apologies if I'm being slow!

Wouldn't we just have a route in Government Frontend like this:

```ruby
"/preview/:base_path/:attachment_id" => AttachmentHandler
```

...and in the `AttachmentHandler` code we'd make a request to Publishing API `https://publishing-api.dev.gov.uk/content/<attachment_id>` and either use the returned data or follow the redirect or display a "Withdrawn" page?

rfc-120-csv-preview.md
ChrisBAshton: Thanks 👍

rfc-120-csv-preview.md
ChrisBAshton: Thanks - I think this can be disambiguated in the ['URL proposal explanation' I suggest here](https://github.com/alphagov/govuk-rfcs/pull/122#discussion_r417836323).

rfc-120-csv-preview.md
benthorner: I'm happy to write something. Can you help me understand what the explanation should look like? The discussion we're having here is really about why the original URL structure wouldn't be appropriate, but doesn't preclude other structures. Would be it be sufficient to document why we're not using the original URL structure?

rfc-120-csv-preview.md
benthorner: Fixed 👍

rfc-120-csv-preview.md
benthorner: I think we can continue a discussion about this as part of implementation. The intention here is really just to show the data would be there is some format.

PR #123, opened 2020-04-20
RFC-123: GitHub Actions CI

issyl0:
- This RFC was requested by @kevindew [when I switched `alphagov/seal` to not build on Jenkins](https://github.com/alphagov/seal/pull/162#pullrequestreview-396261703). And he's had input on some of the words!

- I've tried to put [the words I wrote in response](https://github.com/alphagov/seal/pull/162#issuecomment-616423677) into better words here, but there's more paragraphs there so that comment is worth a read too.

- This is open for two weeks, until **midday on 2020-05-05**.

👉[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/github-actions-ci/rfc-123-github-actions-ci.md)

### DIFF
--- a/rfc-123-github-actions-ci.md
+++ b/rfc-123-github-actions-ci.md
@@ -1,0 +1,102 @@
+# RFC-123: Use GitHub Actions for CI
+
+## Summary
+
+This proposes using [GitHub Actions](https://github.com/features/actions) as our preferred CI tool where applicable.
+
+CI in this case is scoped as follows:
+
+1. Running branch and PR builds for tests and linting, including the `master` branch.
+2. Pushing `release` tags on successful master builds.
+
+This does not include gem publishing or DockerHub pushes, because we feel we need to talk to IA before this decision can be made.
+
+TechOps [started a discovery into CI tooling](https://github.com/alphagov/reliability-engineering/pull/106) a while ago, but it's not reached a conclusion yet. Now we've procured GitHub Enterprise Cloud, making use of the tools available there reduces platform complexity.
+
+## Problem
+
+We currently have many solutions for running branch builds in GOV.UK, including:
+
+- Jenkins CI (self-hosted, behind IP and GitHub team membership restrictions)
+- Travis CI (hosted, paid on a GPC card)
+- Concourse CI (self-hosted in TechOps, behind IP and GitHub team membership restrictions)
+- GitHub Actions CI (hosted, paid through GDS' GitHub Enterprise Cloud subscription)
+
+## Proposal
+
+To switch to GitHub Actions CI for continuous integration (that is, branch and PR builds for tests/linting) as the preferred approach where we do not need the wider platform integration of Jenkins.
+
+This would involve working towards:
+
+ - GOV.UK projects *only* using either GitHub Actions or Jenkins for CI
+ - No projects should use Concourse, Travis or another tool for CI without reasons to justify an exception
+
+This would be a gradual process, because we have many repos, but this decision would deem the usage of Concourse CI and Travis CI as a legacy approach to CI for GOV.UK.
+
+### Examples of existing uses of GitHub Actions
+
+- [GOV.UK Coronavirus Business Volunteer Form][business-volunteer-form], a Rails app with a Postgres database.
+- [GOV.UK AWS linting checks][govuk-aws].
+- [GOV.UK Coronavirus Content YAML validation][govuk-coronavirus-content].
+
+### Pros
+
+- A simpler, more modern approach compared to Jenkins.
+- Using Actions would make it easier to get rid of the CI Jenkins, which would be a huge cost saving - our CI environment is in Carrenza and currently costs ~£6k a month + VAT.
+- GDS as a whole pays for GitHub Enterprise Cloud, which entitles us to unlimited minutes on public repos (both Linux and macOS runners), and 50,000 minutes per month of Actions on private repos.
+- Supported by GitHub, from whom we get Enterprise support via our GitHub Enterprise Cloud subscription.
+- We don't have to host any of our own CI infrastructure.
+- Well documented, easy to [get started][actions-get-started] building a new pipeline. Easier than Concourse for CI as PR builds are built-in to GitHub rather than needing to use [third-party code](https://github.com/telia-oss/github-pr-resource).
+- Doesn't require access to the VPN like our Jenkins does, or membership of any GitHub teams, in order to see logs for public repos.
+- External contributors can see their changes and why they failed, without us having to [push branches][merge-a-pull-request] and paste build logs from Jenkins. This fits more readily with our "code in the open" ethos, and saves on internal developer time.
+- Already in use across some of our production applications. Examples: [business volunteer form][business-volunteer-form], [find support form][find-support-form], [vulnerable people form][vulnerable-people-form]
+- In the future, we can make our internal GitHub Actions into libraries so we’re not duplicating “set up Rails apps” steps across 50 apps and can update them in one place (perhaps `alphagov/actions`?). We already do something similar with `alphagov/govuk-jenkinslib`. See [Homebrew/actions](https://github.com/Homebrew/actions) for an example.
+
+### Cons
+
+- We have less control over what software is on the runners. But most of the time we only need a database, a version of Ruby and some gems. And updating our Jenkins CI instances with new packages is a massive pain.
+- Another system to learn, as well as the replatforming work already ongoing.
+- Duplication of build parameters and steps between all our apps, as they’re all Rails apps. This can be mitigated if we use a shared actions library as proposed earlier.
+
+### Findings from some initial explorations into using GitHub Actions
+
+#### `push` vs `pull_request` build triggers
+
+We considered the following options:
+
+1. Running `on: [pull_request, push]`. This means that internal devs get feedback on their commits when they push them, even if they haven't opened a PR yet (via `push`). External contributors can also get CI feedback on their PRs (via `pull_request`). It does mean that pushing commits to an open PR will waste compute power, time and the planet's precious natural resources, since we'll be running things twice (via `push` and `pull_request`).
+
+2. Running `on: push`. This would persist a current problem we have with running tests on PRs from external contributors. When [a PR is raised from a fork](https://github.com/alphagov/govuk-docker/pull/337), the tests won't run (needs `pull_request`). Actions also [don't run on forked PRs](https://github.com/alphagov/govuk-docker/pull/337). So we'd be back to the annoying state that is [our current process for forked PRs][merge-a-pull-request].
+
+3. Running `on: pull_request`. This has the advantage that it's consistent behaviour for internal and external contributors. Internally we'd have to put up with not having tests run on branches, prior to raising a PR. This would mean changing developer workflows to run more tests locally, or being less worried about "exposing to the world" our force-pushes to get things right on the PR. (Irrespective of this decision, we should always build the `master` branch when merge commits are `push`ed.)
+
+In a [poll of all interested RFC respondents](https://github.com/alphagov/govuk-rfcs/pull/123#issuecomment-620028907), option 1 - running builds on both `pull_request` and `push` events - was the most popular. In that case, going forward, all Actions CI builds will run `on: [push, pull_request]`.
+
+#### `release` tagging via GitHub Actions
+
+Currently, our release tags are pushed [at the end of a Jenkins `master` build](https://github.com/alphagov/govuk-jenkinslib/blob/master/vars/govuk.groovy#L705-L727). The numbers auto-increment each time. These have been reset to 0 on a few occasions during Jenkins upgrades. A consequence of tagging releases on successful master builds via GitHub Actions is that our `release` tags would be reset to 0 once again.
+
+### The state of our repos at the moment
+
+Currently we have:
+
+- The vast majority of our repos run CI checks on Jenkins.
+- The new apps we built recently (the coronavirus forms), and the Seal (as it's not an "app" per-se, it's deployed to Heroku), use GitHub Actions.
+- Some of them (DGU) run CI on Travis.
+- Some GOV.UK Data Labs projects use Concourse for PR checks.
+
+## Outcome and next steps
+
+We will gradually migrate all of the GOV.UK repos to run CI builds with GitHub Actions. To start with, these could be the [`skipDeployToIntegration`][] repos, or tooling. We need to [make `govuk-saas-config` set GitHub Actions checks to "required"][govuk-saas-config].
+
+Gems - or other things that aren't deployed apps on our infrastructure that may currently live on CI Jenkins - will not be migrated to GitHub Actions as part of this RFC. We will address these in a future RFC, potentially along with the GOV.UK replatforming.
+
+[actions-get-started]: https://help.github.com/en/actions/getting-started-with-github-actions
+[business-volunteer-form]: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/.github/workflows/tests.yml
+[find-support-form]: https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/.github/workflows/tests.yml
+[govuk-aws]: https://github.com/alphagov/govuk-aws/blob/master/.github/workflows/ci.yml
+[govuk-coronavirus-content]: https://github.com/alphagov/govuk-coronavirus-content/blob/master/.github/workflows/tests.yml
+[govuk-saas-config]: https://github.com/alphagov/govuk-saas-config/issues/42
+[merge-a-pull-request]: https://docs.publishing.service.gov.uk/manual/merge-pr.html#a-change-from-an-external-contributor
+[skipDeployToIntegration]: https://github.com/search?q=org%3Aalphagov+skipDeployToIntegration&type=Code
+[vulnerable-people-form]: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/.github/workflows/tests.yml


issyl0:
Oops, this should be RFC-123 so it matches the PR number... one moment...

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,67 @@
...
+This would involve working towards GOV.UK projects only using either GitHub Actions or Jenkins for CI, no projects should use Concourse, Travis or another tool for CI without reasons to justify an exception.
+
+This would be a gradual process, because we have many repos, but this decision would deem the usage of Travis CI as a legacy approach to CI for GOV.UK.
+
+### Pros
+
+- A simpler, more modern approach compared to Jenkins.
+- GDS as a whole pays for GitHub Enterprise Cloud, which entitles us to unlimited minutes on public repos (both Linux and macOS runners), and 50,000 minutes per month of Actions on private repos.
+- Supported by GitHub. We get Enterprise support via our GitHub Enterprise Cloud subscription. And we don't have to host any of our own infrastructure.
+- Well documented, easy to [get started][actions-get-started] building a new pipeline. Easier than Concourse for CI as PR builds are built-in to GitHub rather than needing to use third-party code]().
```
injms, line 33: Nitpick - are these brackets meant to be here?

```suggestion
- Well documented, easy to [get started][actions-get-started] building a new pipeline. Easier than Concourse for CI as PR builds are built-in to GitHub rather than needing to use third-party code.
```

issyl0, line 33:
```suggestion
- Well documented, easy to [get started][actions-get-started] building a new pipeline. Easier than Concourse for CI as PR builds are built-in to GitHub rather than needing to use third-party code](https://github.com/telia-oss/github-pr-resource).
```

issyl0: Agh, I thought I meant to include a link but it turns out I didn't.

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,67 @@
...
+This would involve working towards GOV.UK projects only using either GitHub Actions or Jenkins for CI, no projects should use Concourse, Travis or another tool for CI without reasons to justify an exception.
+
+This would be a gradual process, because we have many repos, but this decision would deem the usage of Travis CI as a legacy approach to CI for GOV.UK.
+
+### Pros
+
+- A simpler, more modern approach compared to Jenkins.
+- GDS as a whole pays for GitHub Enterprise Cloud, which entitles us to unlimited minutes on public repos (both Linux and macOS runners), and 50,000 minutes per month of Actions on private repos.
+- Supported by GitHub. We get Enterprise support via our GitHub Enterprise Cloud subscription. And we don't have to host any of our own infrastructure.
+- Well documented, easy to [get started][actions-get-started] building a new pipeline. Easier than Concourse for CI as PR builds are built-in to GitHub rather than needing to use third-party code](https://github.com/telia-oss/github-pr-resource).
```
erino, line 33: I think you're missing an `[` somewhere on this line

issyl0: Argh! Thanks. 😓 I think I fixed it now.

issyl0:
Right, now we've got a correctly rendering Markdown doc (thanks for your patience!)... comment away!

benthorner:
I like this direction we're going in: let's not self-host stuff we don't need to. One minor request: could we put an example of using GitHub actions early on? They're new to me and I wanted to see what it would look like for one of our repos.

On a more major note: a lot of [the current config](https://github.com/alphagov/govuk-jenkinslib/blob/master/vars/govuk.groovy) is taken up with support for the E2E tests, which are the reason for all the Docker stuff. Do we know what the outcome of this will be for the E2E tests? Will they remain separate (on Jenkins)?

issyl0:
@benthorner Thanks, I linked to some existing examples in 9cb2d69.

kevindew:
> On a more major note: a lot of [the current config](https://github.com/alphagov/govuk-jenkinslib/blob/master/vars/govuk.groovy) is taken up with support for the E2E tests, which are the reason for all the Docker stuff. Do we know what the outcome of this will be for the E2E tests? Will they remain separate (on Jenkins)?

I'm imaging they won't be too affected as I'd imagine all the repos that have e2e tests also have integration deployments and thus need to be part of a pipeline - which is outside the scope of this PR. So in theory none of those projects affected until a later, wider post Jenkins discussion kicks in.

How to do the e2e tests does seem something that is hopefully not too difficult but I don't know what the github action approach is. We'd need an action that is set up on either a separate repo or stored somewhere that is shared to be called with different params, and that executed as a particular action. I guess in theory, for lack of a better approach, each repo that needs them could have a command that does the work like:

```
git clone e2e-tests-repo
./run-e2e-tests --for=app --commit=134asdf
```

but that doesn't feel ideal as you'd probably want shared config about the system and environment.

issyl0:
> Thank you for writing this! This is clearly explained, and your pros and cons make sense.

Thank _you_ for your comments! I've only replied to a subset of them now... More tomorrow!

> - When Github is down, that complicates our deploy process.

We do have [a process in place for deploying when GitHub is down](https://docs.publishing.service.gov.uk/manual/github-unavailable.html) as it stands now, because our Jenkinses use GitHub authentication. And that wouldn't change. But for future GitHub outages, we have uptime SLAs from Enterprise support that we have never had before.

> Perhaps we could wait for TechOps to complete their discovery before making a decision?

I've asked @tlwr when the TechOps discovery is likely to start up again, or reach a conclusion.

> Additionally, perhaps a spike could be prepared to move one of our more complex applications over to using Github Actions for CI?

This sounds great! I'm happy to be on that team, but given I've written the vast majority of our GitHub Actions so far, and have been writing Actions for some time outside of work, I definitely shouldn't write the majority of the code.

brucebolt:
I'm assuming the eventual goal of switching all GOV.UK applications to use Github Actions would mean we could massively scale down the capacity of our CI infrastructure.  Have we worked out what this would look like, and how great the financial benefits would be?

issyl0:
@brucebolt That's what it would mean. We've not worked out what the financial benefits would be, but the CI environment is still in Carrenza and therefore it is not cheap. I'll do some digging (with RE's help) and get back to you!

barrucadu:
I like this proposal.

Something I could imagine us doing in the future (beyond the scope of this) is making rubocop, brakeman, and tests separate CI actions, meaning a failure in linting doesn't prevent your tests from running (which is really frustrating).  We can't do this currently because jenkins gets bogged down enough without trebling the number of jobs.

issyl0:
@brucebolt I got some numbers on the costs for our CI environment - thanks, @seb-szy! They're in cfe8a4e.

timwis:
I <3 GitHub Actions. One thing to keep in mind, though, if the usage ever expands to include deployments, is that since the 'steps' are run on docker images, any secrets needed to run that step will be passed to those docker images. That's probably fine for official images (like those in the `actions` namespace), but the flexibility of GH Actions means you can also use third party images. For example, I've used [appleboy/ssh-action](https://github.com/appleboy/ssh-action) and it was simple and easy. But technically I had to pass an SSH private key (or password) to use it. If that image was somehow compromised, the secrets could be stolen. No different than using docker images generally, but I feel like it's possible to use GitHub Actions without realising they're actually just docker images under the hood.

timwis:
I <3 GitHub Actions. One thing to keep in mind, though, if the usage ever expands to include deployments, is that since the 'steps' are run on docker images, any secrets needed to run that step will be passed to those docker images. That's probably fine for official images (like those in the `actions` namespace), but the flexibility of GH Actions means you can also use third party images. For example, I've used [appleboy/ssh-action](https://github.com/appleboy/ssh-action) and it was simple and easy. But technically I had to pass an SSH private key (or password) to use it. If that image was somehow compromised, the secrets could be stolen. No different than using docker images generally, but I feel like it's possible to use GitHub Actions without realising they're actually just docker images under the hood.

timwis:
I <3 GitHub Actions. One thing to keep in mind, though, if the usage ever expands to include deployments, is that since the 'steps' are run on docker images, any secrets needed to run that step will be passed to those docker images. That's probably fine for official images (like those in the `actions` namespace), but the flexibility of GH Actions means you can also use third party images. For example, I've used [appleboy/ssh-action](https://github.com/appleboy/ssh-action) and it was simple and easy. But technically I had to pass an SSH private key (or password) to use it. If that image was somehow compromised, the secrets could be stolen. No different than using docker images generally, but I feel like it's possible to use GitHub Actions without realising they're actually just docker images under the hood.

issyl0:
@timwis Thanks for the input - that context will be useful if (hopefully when) we come to assess the use of Actions for deployments!

issyl0:
There's been some further chat about secrets management in Actions on [my PR that advertised a stretch goal as getting `rubocop-govuk` to publish to RubyGems on new releases](https://github.com/alphagov/rubocop-govuk/pull/41). I'll summarise it in this RFC in the coming days with a case study of Jenkins vs. Actions for publishing gems.

issyl0:
@benthorner Hey! As you've done some very nice work in https://github.com/alphagov/rubocop-govuk/pull/45 (rubocop-govuk now runs tests on PRs with GitHub Actions, and on master publishes the gem!), would you mind writing up a summary as a commit to this RFC?

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,76 @@
...
+- Already in use across [some][business-volunteer-form] of [our][find-support-form] [production][vulnerable-people-form] applications.
+- As we go further down the path, we can make our internal GitHub Actions into libraries so we’re not duplicating “set up Rails apps” steps across 50 apps and can update them in one place - like we do with `alphagov/govuk-jenkinslib` now. Maybe `actions/alphagov`? (See [Homebrew/actions](https://github.com/Homebrew/actions) for an example.)
+- Getting rid of Jenkins would be a huge cost saving - our CI environment is in Carrenza and currently costs ~£6k a month + VAT.
+
+### Cons
+
+- We have less control over what software is on the runners. But most of the time we only need a database, a version of Ruby and some gems. And updating our Jenkins CI instances with new packages is a massive pain.
+- How do we do credentials management for things like pushing draft releases and Docker images to DockerHub?
```
erino, line 49: I think any CI process that uses secrets will have to run on Concourse to get all of the assurances we need.

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,76 @@
...
+- Duplication of build parameters and steps between all our apps, as they’re all Rails apps.
+
+### The state of our repos at the moment
+
+Currently we have:
+
+- The vast majority of our repos run CI checks on Jenkins.
+- The new apps we built recently (the coronavirus forms), and the Seal (as it's not an "app" per-se, it's deployed to Heroku), use GitHub Actions.
+- Some of them (DGU) run CI on Travis.
+- I can't find any uses of Concourse on GOV.UK repos.
```
erino, line 60: A few Data Labs projects use Concourse for regular jobs, not CI

benthorner:
Thanks for the intro @issyl0. Just a few observations from me:

- The test/debug cycle is similar to Jenkins: push it and try it, push it and try it, ...
- Syntax errors are unclear - you have to click down a few screens to see the error

In terms of next steps:

- We should consider having our own actions repo, to DRY-up scripts in other repos
- We could/should clarify if we can set secrets on GitHub

I've copied the RubyGems API key from govuk-secrets into the secrets for the repo. Secrets on GitHub repos can't be read once written, except by the action pipeline, and...

> GitHub automatically redacts secrets printed to the log

rfc-123-github-actions-ci.md
```diff
@@ -50,6 +50,20 @@ This would be a gradual process, because we have many repos, but this decision w
...
+
+- Running `on: pull_request`. Functionally, we run the same tests across the board, whichever of these we choose. But this one has the advantage that it's consistent behaviour for internal and external contributors. But internally we'd have to put up with not having tests run on branches (apart from `master`, we should probably include the `master` branch all the time), which would mean changing developer workflows to run more tests locally (ie RuboCop, a common thing that gets forgotten about), or being less worried about "exposing to the world" our force-pushes to get things right on the PR. We would also probably benefit from opening PRs for long-running branches earlier, using the draft PR functionality, and this would encourage that. This seems like the best option to _Issy_, but please feel free to disagree.
+
```
ChrisBAshton, line 66: Another alternative is wrapping the job in a conditional: https://github.community/t5/GitHub-Actions/don-t-run-actions-on-pull-request-from-fork/td-p/45499

```
jobs:
  prcheck:
    if: github.actor == 'repo-owner'
    runs-on: [ubuntu-latest]
    # etc
```

rfc-123-github-actions-ci.md
```diff
@@ -50,6 +50,20 @@ This would be a gradual process, because we have many repos, but this decision w
...
+Some solutions to this include:
+
+- Running `on: [pull_request, push]`. This means that internal devs get feedback on their commits when they push them, for example if they've forgotten to run RuboCop locally. And also so that external contributors PRs get CI feedback on their changes. It does mean that for long-running test suites, we could waste compute power, time and the planet's precious natural resources running things twice that don't need to be run twice.
+
+- Running `on: push`. Actions don't run [by default on forked PRs](https://github.com/alphagov/govuk-docker/pull/337) - it's a security feature that GitHub has implemented. And [when a PR is raised from a fork](https://github.com/alphagov/govuk-docker/pull/337) and the receiving repo only has a push trigger, the tests don't run. So we'd be back to the annoying state that is [our current process for forked PRs][merge-a-pull-request].
```
kevindew, line 63: Should be flagged here that this doesn't merge with master so we only know the branch passes and not if the integrated branch passes.

rfc-123-github-actions-ci.md
```diff
@@ -50,6 +50,20 @@ This would be a gradual process, because we have many repos, but this decision w
...
+
+- Running `on: pull_request`. Functionally, we run the same tests across the board, whichever of these we choose. But this one has the advantage that it's consistent behaviour for internal and external contributors. But internally we'd have to put up with not having tests run on branches (apart from `master`, we should probably include the `master` branch all the time), which would mean changing developer workflows to run more tests locally (ie RuboCop, a common thing that gets forgotten about), or being less worried about "exposing to the world" our force-pushes to get things right on the PR. We would also probably benefit from opening PRs for long-running branches earlier, using the draft PR functionality, and this would encourage that. This seems like the best option to _Issy_, but please feel free to disagree.
```
kevindew, line 65: This seems to be closest to how Jenkins runs now by always merging master into what's requested. 

It does seem annoying to lose CI though prior to opening a PR and does mean keeping the confusion of branch passing locally but failing on the integrated.

rfc-123-github-actions-ci.md
```diff
@@ -50,6 +50,20 @@ This would be a gradual process, because we have many repos, but this decision w
...
+
+Some solutions to this include:
+
+- Running `on: [pull_request, push]`. This means that internal devs get feedback on their commits when they push them, for example if they've forgotten to run RuboCop locally. And also so that external contributors PRs get CI feedback on their changes. It does mean that for long-running test suites, we could waste compute power, time and the planet's precious natural resources running things twice that don't need to be run twice.
```
kevindew, line 61: I've come around to be in favour of this option since they're slightly different changes. Does seem a pain that we might often have very similar tests run - but at least they'd be in parallel. 

I feel like this would become much more of a pain point though if action queues get long and we find we're waiting for one and not the other.

issyl0: On the Enterprise plan, we get [180 concurrent builds](https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits) across `alphagov`, assuming they're all Linux builds and not macOS.

rfc-123-github-actions-ci.md
benthorner: This is still my preferred option. I think it's confusing to see two sets of tests being run, and people running tests locally just seems easier and more efficient.

Conclusion: is the downside, really a downside, or just a habit we're trying to accommodate?

rfc-123-github-actions-ci.md
issyl0: GitHub does have [an option in the branch protection settings](https://help.github.com/en/github/administering-a-repository/types-of-required-status-checks) to enforce that branches are up to date with `master` before merging. But [we don't currently enforce it](https://github.com/alphagov/govuk-saas-config/blob/master/github/lib/configure_repo.rb#L95).

rfc-123-github-actions-ci.md
issyl0: I was originally in favour of the first option, running on PRs and Push events, but upon writing all the findings up, I came around to this option: only on PRs. We shouldn't be scared of showing off our work, force-pushing, or opening draft PRs - and PRs can always be closed and re-opened!

kevindew: > We shouldn't be scared of showing off our work, force-pushing, or opening draft PRs - and PRs can always be closed and re-opened!

I don't think it's a factor of being scared to show work but more a question of time and efficiency. I can do some work that I expect is fine, push the branch and move onto more work and check that build passed CI. Whereas if I pause working run a full rake task that's going to add more time.

Similarly once I'm putting together a PR it's nice to know the build has passed so that once I click open it's ready for review. It's also nice that all the build process can run while I'm writing the PR.

I'm sure I'd get used to it only being on PR's but it feels like, at least for me, it'll add inefficiencies and extra time to code process. Which is something I'd rather be absorbed by CI machines than myself.

rfc-123-github-actions-ci.md
issyl0: At the end of the day, I don't mind which option we go for (as long as it's one of "only PRs" or "both PRs and pushes", the "on push" option is in no way an improvement to our current processes), but I'd like to solicit some more opinions from others otherwise this feels a bit circular. :-)

theseanything: I can see value in having test run on your branch before a PR is raised, therefore you can immediately notice a failure, rather than waiting after you've raised the PR. I feel like going with both lets us see if there's any value/downsides and we can always tweak these settings in the future.

issyl0:
Here's a makeshift poll. I'd have liked to choose some less "meaningful" emoji (like fruit or something), but we have to go with the reactions that GitHub gives us! Having digested the words and comments about the actions run triggers, if you would like CI checks to run in the following ways, please react with the specified emoji:

 - ❤️ for on `push` only 
- 🚀 for on both `push` and `pull_request`
- 🎉 for on `pull_request` only

rfc-123-github-actions-ci.md
karlbaker02: We're increasingly using Concourse for running PR tests, linting etc before they're able to be merged (nothing is run on merge itself).

issyl0: Ah, I missed you out! My assessments here were done based on the contents of my local `~/govuk` folder, and I don't think I have any of the data labs stuff.

rfc-123-github-actions-ci.md
barrucadu: I don't think many people rebase before merging, so switching that on would be a big disruptive change for a lot of people.

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,90 @@
...
+- Duplication of build parameters and steps between all our apps, as they’re all Rails apps.
+
+### Findings from some initial explorations into using GitHub Actions
+
+#### `push` vs `pull_request` build triggers
+
+There has been some [chatter](https://github.com/alphagov/govuk-developer-docs/pull/2497#discussion_r415572147) about if we want `push` or `pull_request` triggers for tests to run, or both. The GOV.UK Developer Tools group has done some investigation into the differences between this, as there was some concern about if we wanted long-running builds (in the case of apps with large test suites, like whitehall) to run twice when not needed. There is not currently any way to differentiate and say "run on PRs from forks" and "run on push events to main repo branches". We'd have really liked a `pull_request_only_from_forks` setting, but that doesn't exist. It's worth noting that this isn't a problem _only_ with GitHub Actions: [Travis](https://github.com/travis-ci/travis-ci/issues/1147) has had the same pain points for years, and we currently do double builds for the apps we use Travis for.
+
+Some solutions to this include:
```
ChrisBAshton, line 59:
```suggestion
Our options are:
```

rfc-123-github-actions-ci.md
ChrisBAshton: It's worth describing that advantage in the RFC 👍

rfc-123-github-actions-ci.md
ChrisBAshton: I don't think it would be particularly disruptive actually. People on long-running branches regularly rebase against master, in my experience - and for people doing small tweaks, GitHub offers a "Rebase to master" button right on the PR itself.

The only disruption is waiting for the tests to pass again, especially if there's a flaky publishing-e2e-tests downstream too. I guess it might encourage us to make our test suites faster and more reliable.

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,90 @@
...
+Currently we have:
+
+- The vast majority of our repos run CI checks on Jenkins.
+- The new apps we built recently (the coronavirus forms), and the Seal (as it's not an "app" per-se, it's deployed to Heroku), use GitHub Actions.
+- Some of them (DGU) run CI on Travis.
+- I can't find any uses of Concourse on GOV.UK repos.
+
+## Outcome and next steps
+
+The outcome of this is defined in the proposal. In the first instance, I imagine the work to do this will be taken on by the GOV.UK Developer Tools team - please, get involved!
```
ChrisBAshton, line 78: I think we can remove this line.

issyl0:
```suggestion

```

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,90 @@
...
+CI in this case is defined as "running builds for tests and/or linting, on branches and PRs". This does not cover *deployment* of the master branch, which will continue to happen using Jenkins (until such time as we replatform (see RFC 117)) and will be considered as part of wider built pipeline work (such as continuous deployment).
+
+TechOps [started a discovery into CI tooling](https://github.com/alphagov/reliability-engineering/pull/106) a while ago, but it's not reached a conclusion yet. Now we've procured GitHub Enterprise Cloud, making use of the tools available there reduces platform complexity.
+
+## Problem
+
+We currently have many solutions for running branch builds in GOV.UK, including:
+
+- Jenkins CI (self-hosted, behind IP and GitHub team membership restrictions)
+- Travis CI (hosted, paid on a GPC card)
```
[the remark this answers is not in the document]
issyl0: A Government Procurement Card. A credit card (there's only a few in existence within GDS), it allows government to buy things that don't come invoiced.

rfc-123-github-actions-ci.md
issyl0: We should. @bevanloon @richardTowers Can we have a chat about this? It also relates to [this gem publishing comment](https://github.com/alphagov/rubocop-govuk/pull/41#issuecomment-618378751). (We're already doing gem publishing with GitHub Actions, and it's working well.) There's some detailed discussion on GitHub Actions secrets that I linked to in the above `rubocop-govuk` PR - that's worth a read.

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,90 @@
...
+- Doesn't require access to the VPN like our Jenkins does, or membership of any GitHub teams, in order to see logs for public repos.
+- External contributors can see their changes and why they failed, without us having to [push branches][merge-a-pull-request] and paste build logs from Jenkins. This fits more readily with our "code in the open" ethos, and saves on internal developer time.
+- Already in use across [some][business-volunteer-form] of [our][find-support-form] [production][vulnerable-people-form] applications.
+- As we go further down the path, we can make our internal GitHub Actions into libraries so we’re not duplicating “set up Rails apps” steps across 50 apps and can update them in one place - like we do with `alphagov/govuk-jenkinslib` now. Maybe `actions/alphagov`? (See [Homebrew/actions](https://github.com/Homebrew/actions) for an example.)
+- Getting rid of Jenkins would be a huge cost saving - our CI environment is in Carrenza and currently costs ~£6k a month + VAT.
```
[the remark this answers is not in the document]
issyl0: We have deploy Jenkinses in all six environments:

- [Integration Carrenza](https://ci-deploy.integration.publishing.service.gov.uk) (that's where we deploy Terraform from for the rest of the environments, and the CI environment)
- [Integration AWS](https://deploy.integration.publishing.service.gov.uk)
- [Staging Carrenza](https://deploy.staging.publishing.service.gov.uk)
- [Staging AWS](https://deploy.blue.staging.govuk.digital)
- Production Carrenza](https://deploy.publishing.service.gov.uk)
- Production AWS](https://deploy.blue.production.govuk.digital)

rfc-123-github-actions-ci.md
issyl0: That said, can anyone think of any CI (ie linting and app tests - as opposed to *deployment processes*) that need secrets to run?

rfc-123-github-actions-ci.md
benthorner: Have we considered another potential disadvantage: if we're running the tests twice, and they (both) fail, which one do I look at to see the failure? While this may be obvious for us, the architects of the future, I can see it being a source of confusion: which "Details" link should I click on when I see a failure?

issyl0:
The `on: [push, pull_request]` option has a three point lead (reminder: my votes were just to set up the poll and don't count) - is that enough of a majority? If so, I'll write up a decision paragraph.

@ChrisBAshton Thanks so much for your review - I've encorporated your wording changes. I agree on the other things - I'll do some more thinking and write yet more words. 😩

Overall, with three working days left, this is looking pretty positive, so thanks all for your input! Reminder: this closes at 5pm on Monday 5th May.

rfc-123-github-actions-ci.md
benthorner: I'm not aware of anything. However, if we don't allow secrets, do we have an extra consequence that gem releases need to happen <somewhere else>?

richardTowers: I think it would be easier to make the "no secrets" decision for now - that still lets us do the majority of CI things, but as Ben says, we'll need to decide where to do gem releases (because we can't really do those in GitHub Actions without doing the secrets / IA / cyber dance).

benthorner: 👍 - sounds like we'll be wanting to revert part of the demo for [rubocop-govuk](https://github.com/alphagov/rubocop-govuk/pull/45) and go with [some kind of restricted Jenkins build](https://github.com/alphagov/govuk-jenkinslib/pull/70) instead. That does seem like a slightly worse situation than what we had before, but I guess it only applies to a few repos - we only have a few gems.

kevindew: The pact contract testing process requires secrets to publish new pacts (used by Publishing API, GDS API Adapters and maybe Content Store).

I do think we'd be creating quite a pain point if we go down a no secrets route and would mean that we retrospectively need to undo some travis jobs we have that release npm modules. So I do think we should resolve it rather than going down "no secrets" as a simple option.

issyl0: Can we go forwards not backwards, please? I'd also argue that any gem publishing we do on Jenkins (if we must) should be done on the deploy Jenkinses rather than the CI Jenkinses - for separation of concerns.

issyl0: (Also my apologies, that previous comment came across a bit angry.)

rfc-123-github-actions-ci.md
issyl0: If the failures are different then we have more interesting problems. I'd look at the `push` ones, personally, as I'll be able to see quicker that it's failed. Unless it's a forked PR.

rfc-123-github-actions-ci.md
richardTowers: Maybe "no secrets" is the wrong heuristic. I think the gem ones are sensitive because there's a possible vector for compromising GOV.UK in there:

1) steal the rubygems secret from GitHub actions somehow
2) publish a malicious version of a gem that we use
3) wait for dependabot to update it
4) ?????
5) very bad things

Not to say that won't be an acceptable risk once we've spoken to cyber and IA (depends on how we feel about GitHub actions' security), but they definitely need to be in the loop.

The pact example is much less problematic, as it's not going to affect production code.

Also worth noting that I don't actually understand how we used to publish our gems, so I'm not sure what going forwards and backwards actually means...

kevindew: > Maybe "no secrets" is the wrong heuristic. I think the gem ones are sensitive because there's a possible vector for compromising GOV.UK in there:
> 
> 1. steal the rubygems secret from GitHub actions somehow
> 2. publish a malicious version of a gem that we use
> 3. wait for dependabot to update it
> 4. ?????
> 5. very bad things
> 
> Not to say that won't be an acceptable risk once we've spoken to cyber and IA (depends on how we feel about GitHub actions' security), but they definitely need to be in the loop.
> 
> The pact example is much less problematic, as it's not going to affect production code.
> 
> Also worth noting that I don't actually understand how we used to publish our gems, so I'm not sure what going forwards and backwards actually means...

I'd suggest that same vector applies for any gem and even our ones currently as there are already means that either our account or RubyGems is compromised. So that vector should be covered by always verifying what you're merging with dependabot (whether people always do is another matter)

I'd describe our risk as reputational, since were a token to be compromised then GOV.UK's name would be associated with it.

On something of a tangent the move to GitHub actions might also reduce security risks from a different angle as, for apps enabled, we would be building dependabot PRs on GitHubs infrastructure rather than our own.

issyl0: OK so I had a chat with Richard offline (thanks!) and we would like to reiterate that this RFC was only supposed to be about CI (branch builds for testing and linting). For *this case* GitHub Actions is fine. For gems and projects that aren't production code, GitHub Actions is fine for releasing too. For instance, `rubocop-govuk` is only a dev and test gem - and Kevin's `npm` pushes are fine as `html-to-govspeak` isn't a production service either, so it's fine to keep them as-is.

There may well be extensions to this RFC in future when we're at the point where IA/Cyber have done their assessments into GitHub Actions secrets security for *production* secrets, but that's the future. The *now* on this is GitHub Actions for CI builds, to save us a tonne of money and increase developer productivity.

Thank you. Now where's the "also send to channel" button on for GitHub threads...

seb-szy:
As you've mentioned all the GitHub actions within public repos are free and we have 50k minutes for private repos with our current licensing model.
We should probably think how we will track requirements across all GDS teams to run GitHub Actions for private repos. For the time being when GDS hit 50k minutes for private repos these will stop. With extra savings around scaling down Jenkins we could pay/allocate additional minutes for this but this is not set yet.
Keep in mind there is also a limit on number of concurrent GitHub Action jobs running:
https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits

issyl0:
I booked Monday off, and I don't want to be even worse at relaxing than I am normally by interrupting my day off and coming here, soliciting approvals and wrapping up this RFC at 5pm, so I'm extending this RFC deadline to midday Tuesday. Thanks!

rfc-123-github-actions-ci.md
karlbaker02: I'm definitely an advocate of rebasing before pushing, just to make sure that everything runs fine against the latest master code. It can be disruptive in a fast-moving repo though as Chris points out above if in the time tests take to run, more commits are pushed.

karlbaker02:
One of my concerns around this was on a similar line as Bill's, whereby if GitHub is down then this is going to affect all CI tests for the branch that you're running against. In reality though, if GitHub is down then we probably have bigger issues regardless of where CI runs, so GitHub Actions doesn't make any difference to this. Only if GitHub Actions itself were to be unavailable whilst the rest of GitHub were operational would we see any benefit of having CI hosted elsewhere.

Agree with this RFC 👍

kevindew:
> * We'll be moving gem releases to our `deploy` machines. Although some gems are not run in production, they are run on people's machine with access to production. It also makes sense for us to do gem deployments in a consistent way, and it sounds like storing secrets on GitHub is 🚫.

Did we not resolve that releasing gems was ok? It looks like I missed a nuance of https://github.com/alphagov/govuk-rfcs/pull/123#discussion_r417935201 with regard to the word "production". Oh dear, I had thought that was all resolved. It doesn't make sense to distinguish between releases of code being production or not.

issyl0:
> > * We'll be moving gem releases to our `deploy` machines. Although some gems are not run in production, they are run on people's machine with access to production. It also makes sense for us to do gem deployments in a consistent way, and it sounds like storing secrets on GitHub is 🚫.
> 
> Did we not resolve that releasing gems was ok? It looks like I missed a nuance of [#123 (comment)](https://github.com/alphagov/govuk-rfcs/pull/123#discussion_r417935201) with regard to the word "production". Oh dear, I had thought that was all resolved. It doesn't make sense to distinguish between releases of code being production or not.

I had thought we'd resolved this too, to be "releasing gems is OK".

issyl0:
> Thanks for working through everyone's comment @issyl0 ⭐. So many sub-topics!

Thanks for the thanks. It's taken *a lot* of scrolling! 😅

I think Kevin's covered the rest of your points, but on the last one:

> I think we should also extend the scope to include pushing 'release' tags. It would be weird if the `deploy` machines did that. Although we could use [something like `GITHUB_RUN_NUMBER`](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables), we should be clear that a consequence of this would be to reset our current `release_X` counters.

Totally. I'll write a sentence in the doc for this. We have in the past reset `release` counters to 0 via botched Jenkins updates (accidentally overwriting things), so there is precedent! But this would be a better way to reset them - to intentionally do so!

benthorner:
@kevindew @issyl0 I'm pretty sure there is a nuance in that comment:

> For gems and projects that aren't production code, GitHub Actions is fine for releasing too.

So what about `gds-api-adapters`? And `html-to-govspeak` is also production code. Whatever we decide, I think we need to clarify that in the scope, as neither gems nor Docker fall under "running builds for tests and/or linting", and much more under "*deployment* of the master branch".

I'm not fussed either way. Hopefully this discussion illustrates the need to write it down 😸.

kevindew:
> @kevindew @issyl0 I'm pretty sure there is a nuance in that comment:
> 
> > For gems and projects that aren't production code, GitHub Actions is fine for releasing too.
> 
> So what about `gds-api-adapters`? And `html-to-govspeak` is also production code. Whatever we decide, I think we need to clarify that in the scope, as neither gems nor Docker fall under "running builds for tests and/or linting", and much more under "_deployment_ of the master branch".
> 
> I'm not fussed either way. Hopefully this discussion illustrates the need to write it down 😸.

Sure, sounds like something to clarify.

I think the intention and, hopefully, shared understanding is that it is ok for GitHub Actions to talk to other services and release things with our permission so long as those services don't have the ability to change our production code. So releasing a gem is ok because someone still has to merge that to reach production. Whereas if it was common for us to have scripts that blindly run `gem install [x]` with no version constraints (I hope we don't have those) in production this would illustrate a risk.

issyl0:
I just spent half an hour on a call with Richard Towers and we reviewed the comments and more explicitly called out the scope of this RFC in the first instance.

It's because of all your comments that it's become at its end a more defined piece of writing than it was at the start - all your comments have made a lot of sense and it's been interesting and educational trying to group them, define the things we can and cannot do right now, and still incorporate so many.

Hopefully that's enough for this to be overall positive still, and sets more of a defined direction. Thanks again! I'll be back on Tuesday!

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,102 @@
...
+Currently we have:
+
+- The vast majority of our repos run CI checks on Jenkins.
+- The new apps we built recently (the coronavirus forms), and the Seal (as it's not an "app" per-se, it's deployed to Heroku), use GitHub Actions.
+- Some of them (DGU) run CI on Travis.
+- Some GOV.UK Data Labs projects use Concourse for PR checks.
+
+## Outcome and next steps
+
+We will gradually migrate all of the GOV.UK repos to run CI builds with GitHub Actions. To start with, these could be the [`skipdeployToIntegration`][] repos, or tooling. We need to [make `govuk-saas-config` set GitHub Actions checks to "required"][govuk-saas-config].
```
benthorner, line 90:
```suggestion
We will gradually migrate all of the GOV.UK repos to run CI builds with GitHub Actions. To start with, these could be the [`skipDeployToIntegration`][] repos, or tooling. We need to [make `govuk-saas-config` set GitHub Actions checks to "required"][govuk-saas-config].
```

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,102 @@
...
+- Another system to learn, as well as the replatforming work already ongoing.
+- Duplication of build parameters and steps between all our apps, as they’re all Rails apps. This can be mitigated if we use a shared actions library as proposed earlier.
+
+### Findings from some initial explorations into using GitHub Actions
+
+#### `push` vs `pull_request` build triggers
+
+We considered the following options:
+
+1. Running `on: [pull_request, push]`. This means that internal devs get feedback on their commits when they push them, for example if they've forgotten to run RuboCop locally. And also so that external contributors PRs get CI feedback on their changes. It does mean that for long-running test suites, we could waste compute power, time and the planet's precious natural resources running things twice that don't need to be run twice.
```
benthorner, line 67:
```suggestion
1. Running `on: [pull_request, push]`. This means that internal devs get feedback on their commits when they push them, even if they haven't opened a PR yet (via `push`). External contributors can also get CI feedback on their PRs (via `pull_request`). It does mean that pushing commits to an open PR will waste compute power, time and the planet's precious natural resources, since we'll be running things twice (via `push` and `pull_request`).
```

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,102 @@
...
+
+2. Running `on: push`. Actions don't run [by default on forked PRs](https://github.com/alphagov/govuk-docker/pull/337) - it's a security feature that GitHub has implemented. And [when a PR is raised from a fork](https://github.com/alphagov/govuk-docker/pull/337) and the receiving repo only has a push trigger, the tests don't run. So we'd be back to the annoying state that is [our current process for forked PRs][merge-a-pull-request].
```
benthorner, line 69:
```suggestion
2. Running `on: push`. This would persist a current problem we have with running tests on PRs from external contributors. When [a PR is raised from a fork](https://github.com/alphagov/govuk-docker/pull/337), the tests won't run (needs `pull_request`). Actions also [don't run on forked PRs](https://github.com/alphagov/govuk-docker/pull/337). So we'd be back to the annoying state that is [our current process for forked PRs][merge-a-pull-request].
```

rfc-123-github-actions-ci.md
```diff
@@ -0,0 +1,102 @@
...
+
+3. Running `on: pull_request`. Functionally, we run the same tests across the board, whichever of these we choose. But this one has the advantage that it's consistent behaviour for internal and external contributors. But internally we'd have to put up with not having tests run on branches (apart from `master`, we should probably include the `master` branch all the time), which would mean changing developer workflows to run more tests locally (ie RuboCop, a common thing that gets forgotten about), or being less worried about "exposing to the world" our force-pushes to get things right on the PR. We would also probably benefit from opening PRs for long-running branches earlier, using the draft PR functionality, and this would encourage that. This seems like the best option to _Issy_, but please feel free to disagree.
```
benthorner, line 71:
```suggestion
3. Running `on: pull_request`. This has the advantage that it's consistent behaviour for internal and external contributors. Internally we'd have to put up with not having tests run on branches, prior to raising a PR. This would mean changing developer workflows to run more tests locally, or being less worried about "exposing to the world" our force-pushes to get things right on the PR. (Irrespective of this decision, we should always build the `master` branch when merge commits are `push`ed.)
```

richardTowers:
> I guess that means we'll have to revert repos like rubocop-govuk back to running on CI?

I think that depends on how much we care about consistency in the short term. In the rubocop-govuk case I don't think there's any cause for IA concern, so there's no problem with it running on GitHub actions, other than the fact that it's inconsistent with the other gems. I'd be happy for it to stay the way it is for now, unless people think the difference will be too confusing.

In the medium term, I think it's still likely we'll want to adopt GitHub actions for pushing to rubygems - it's just  that we need to bring IA along with us in that conversation.

benthorner:
Thanks for the update @richardTowers! I'm still a bit confused about the distinction between `rubocop-govuk` and other gems. The thing we're worried about seems to be the possible leakage of the API token, which would still be an issue. If someone had the token, they could mess with any of our gems, whether or not they're using GitHub Actions.

Am I missing something subtle?

richardTowers:
Ah no, it's me who's missing the subtlety. I hadn't twigged that the API key to deploy the one rubygem would give you access to deploy the others.

Given that, and strictly speaking, I think you're right that we should revert the rubocop-govuk gem deployment until we've got the nod from IA.

issyl0:
I have returned to an approval! Hurray! Thanks, Ben. I'll take a quick look at the remaining comments today but I think it's pretty much there.

Apologies about the gem publishing thing - I did indeed misinterpret my own comment originally. I read the "gems and production apps" bit as if there were an Oxford comma! Sorry for any wasted effort on the gem side of things, Ben. But it's not really wasted because we learnt a lot along the way and there will be prior art we can use ("Revert Revert" anyone?) _when_ IA say yes to it.

issyl0:
I squashed all the commits into one, so that what's on master (after midday) reflects the full thinking in one place. All the comments will be left here!

kevindew:
Thanks Issy 💯

issyl0:
Hitting that button feels gooood! I've got an hour in my calendar now to go through and work out which GOV.UK repos use what CI solution, what type of app it is (production app, gem, tool, infrastructure), and work out some kind of a migration plan: easiest, less dependent ones first. If anyone would like to join me, feel free!

PR #124, opened 2020-04-26
RFC 124: Use Edge Side Include (ESI) to deploy banners on GOV.UK

alex-ju:
**TL;DR**: Using Edge Side Include (ESI) to deploy banners on GOV.UK will help us move away from `static`/`slimmer` while being able to roll out changes quickly

**Deadline for comments**: 11th May 2020 (~2 weeks)

👉 [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/rfc-124-deploying-banners/rfc-124-deploying-banners.md)

### DIFF
--- a/rfc-124-deploying-banners.md
+++ b/rfc-124-deploying-banners.md
@@ -1,0 +1,81 @@
+# Use Edge Side Include (ESI) to deploy banners on GOV.UK
+
+## Summary
+
+Using Edge Side Include (ESI) to deploy banners on GOV.UK will help us move away from `static`/`slimmer` while being able to roll out changes quickly.
+
+## Background
+A few types of site-wide banners can be activated to convey important information on GOV.UK.
+
+### Emergency banner
+The information for the emergency banner is [stored in Redis](https://github.com/alphagov/static/blob/998d94e1e1c70dabbe86384f565966acc8ef5110/lib/emergency_banner/deploy.rb#L4). [`Static`](https://github.com/alphagov/static/blob/master/app/views/notifications/_emergency_banner.html.erb) is responsible for displaying the information. We use Jenkins to run [rake tasks in `static`](https://github.com/alphagov/static/blob/master/lib/tasks/emergency_banner.rake) to set or delete the appropriate hash in Redis.
+
+### Non-emergency banner
+The information for the non-emergency banner is [hardcoded in a view in `static`](https://github.com/alphagov/static/blob/master/app/views/notifications/_global_bar.html.erb#L2-L11). To update this information and show or hide the banner a new pull request must be created with the changes, merged to master, then deployed across all environments. A list with the pages where the banner should not be displayed is [hardcoded in the JavaScript that initialises the banner](https://github.com/alphagov/static/blob/master/app/assets/javascripts/global-bar-init.js#L27-L39). This banner has [a mechanism based on cookies to show the banner only a certain number of times](https://github.com/alphagov/static/blob/d900b5be5b6176ce66008f6bc368908cd0cda5e3/app/assets/javascripts/modules/global-bar.js#L62-L70).
+
+### Cookies banner
+The information for the [cookie banner is in the shared components library](https://components.publishing.service.gov.uk/component-guide/cookie_banner). To update the information consistently across the public-facing applications in GOV.UK, a new pull request in the components library must be created with the changes, merged to master then publish a new release of the components library. All the public-facing applications need to be updated to use the latest version of the components library. This update must be deployed across all environments for each application.
+
+![The current mechanism for deploying banners on GOV.UK](https://docs.google.com/drawings/d/e/2PACX-1vS-vz5bPpGAOhATHYH6hqY5BSR_T-JPSJkuGeneuZKe6Ae8Z4vShcPw-5Im9_OESK1zapFUZX4gNoX4/pub?w=1504&amp;h=909)
+
+The current mechanism for deploying banners on GOV.UK – [view source](https://docs.google.com/drawings/d/1XgSskX5Ufb6BU_PiaIvAHsj7bTARMMgmWlkfOf3f_uA/edit)
+
+## Problem
+
+The recent events – such as making the cookie banner Privacy and Electronic Communications Regulations (PECR) compliant or showing important information related to Brexit and more recently COVID-19 – have made us more aware about the importance of being able to deploy banners quickly and with confidence without overloading our infrastructure. As described in the previous section, the process for deploying banners on GOV.UK is not consistent and has a certain degree of complexity depending on the banner type.
+
+Moving away from `static`/`slimmer` towards a fully component-based architecture makes the process or rolling out a change across all applications more difficult.
+
+## Proposal
+
+Use Edge Side Include (ESI) to deploy banners on GOV.UK.
+
+Edge Side Include (ESI) is a [web standard](https://www.w3.org/TR/esi-lang) originally proposed by Akamai and Oracle, among other companies. It allows an Edge Server (like [Fastly’s caches](https://www.fastly.com/blog/using-esi-part-1-simple-edge-side-include)) to "mix and match" content from multiple URLs.
+
+In practice, the emergency banner on GOV.UK, for example, will require the following changes:
+
+1. make the emergency banner available as a fragment (e.g. on `/emergency-banner`); this should be hosted on a frontend server, potentially served by a new application called `banners`, similarly with the other public-facing application
+1. add an ESI include tag (pointing to the fragment) in the application layout
+```
+  <esi:include src="/emergency-banner" />
+```
+1. add an ESI remove tag and ESI comment to provide a fallback on environments where ESI is not available
+```
+  <esi:remove>
+    <%= render "/components/emergency_banner" %>
+  </esi:remove>
+  <!--esi
+  <esi:include src="/emergency-banner" />
+  -->
+```
+1. enable ESI in Fastly via Varnish Configuration Language (VCL)
+1. optional: add extra logic in VCL. Fastly gives us the ability to create custom responses at the edge – called "synthetics" in Varnish terminology – which can be used to include personalised, dynamic information that will be served from the edge instead of origin
+
+### Pros
+- performance gains (a higher cache-hit ratio, quicker page loads, less traffic spikes)
+- the include fragment will have its own URL and thus independent VCL logic and caching TTLs
+- due to the above, fragments can be purged independently of the main HTML content – this allows banners to roll-out instantly without having to drop the cache of all public-facing applications in GOV.UK
+- ESI remove can be used to store fallback HTML content (which could be the rendered banner component) and therefore make local development easier and more consistent with production
+
+### Cons
+- origin must serve top-level page HTML as an uncompressed response (or support content-encoding negotiation) in order for Varnish to be able to do streaming ESI replacement
+- it is also not possible for Varnish to compress the response after ESI processing. Fortunately Fastly has a workaround for this which forces H2O (Fastly’s HTTP terminator and HTTP/2-3 server) to do the streaming compression (which supports both `gzip` and `brotli`). This can easily be enabled via a HTTP response header which is subsequently stripped by H2O. However, as Fastly bills from bytes out of Varnish this would result in you getting billed for the uncompressed bytes instead of the compressed version. We are on a flat fee plan and thus this should not be an issue
+- if ESI is not implemented correctly can make us vulnerable to different attacks (SSRF and XSS)
+
+What a consistent mechanism to update banners would look like:
+- content changes and enabling/disabling banners should be done in `banners`. This can be initially done through code changes, YAML configuration or rake tasks (similar to the current process for the emergency banner) then potentially improved to be powered by an admin interface.
+- design and functional changes should be done in `govuk_publishing_components`, as for the rest of the UI components. A new version of the gem must be released then `banners` updated and deployed across all environments.
+
+![The proposed mechanism for deploying banners on GOV.UK](https://docs.google.com/drawings/d/e/2PACX-1vRRMbxkghxpS_SHB706QfcM2M9vwQZZ--bzMxr17Fj1k_GJ7Z_rCoLqba9tJQqEXH2VTPyrxNh_W4rR/pub?w=1503&amp;h=908)
+
+The proposed mechanism for deploying banners on GOV.UK – [view source](https://docs.google.com/drawings/d/1S9VaY6jjpFo3QWQ4CJ4fhSwpWn12BNqunXhT27JdDAY/edit)
+
+
+### References
+- [Edge Side Includes (ESI) standard](https://www.w3.org/TR/esi-lang)
+- [Fastly documentation on ESI configuration](https://docs.fastly.com/en/guides/using-edge-side-includes)
+- [GOV.UK’s emergency banner](https://docs.publishing.service.gov.uk/manual/emergency-publishing.html)
+- [GOV.UK’s non-emergency banner](https://docs.publishing.service.gov.uk/manual/global-banner.html)
+- [RFC 84: Replace `static` by a gem](https://github.com/alphagov/govuk-rfcs/blob/bc8ffe85cdf5cdf5005502cba50d5b64237f1b71/rfc-084-frontend-in-a-gem.md)
+- [RFC 95: Long term future of applications](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-095-long-term-future-apps.md)
+- [RFC 118: Storing global content in `content-store`](https://github.com/alphagov/govuk-rfcs/pull/118/files?short_path=e726d76#diff-e726d76080573a2313d111b44a90f38d)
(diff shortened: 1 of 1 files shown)


rfc-124-deploying-banners.md
```diff
@@ -0,0 +1,81 @@
...
+  <esi:include src="/emergency-banner" />
+```
+1. add an ESI remove tag and ESI comment to provide a fallback on environments where ESI is not available
+```
+  <esi:remove>
+    <%= render "/components/emergency_banner" %>
+  </esi:remove>
+  <!--esi
+  <esi:include src="/emergency-banner" />
+  -->
```
ChrisBAshton, line 49: Might be worth marking this up as a `diff` so it's clearer you're modifying the code you added in the previous step. It took me a little while to figure out you didn't want to have `<esi:include` appearing twice in the document.

```diff
+  <esi:remove>
+    <%= render "/components/emergency_banner" %>
+  </esi:remove>
+  <!--esi
  <esi:include src="/emergency-banner" />
+  -->
```

rfc-124-deploying-banners.md
```diff
@@ -0,0 +1,81 @@
...
+```
+1. enable ESI in Fastly via Varnish Configuration Language (VCL)
+1. optional: add extra logic in VCL. Fastly gives us the ability to create custom responses at the edge – called "synthetics" in Varnish terminology – which can be used to include personalised, dynamic information that will be served from the edge instead of origin
```
ChrisBAshton, line 52: Are you hinting that this could be used to, for example, configure which pages the banner should not display on (instead of using JS for that)?

rfc-124-deploying-banners.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Pros
+- performance gains (a higher cache-hit ratio, quicker page loads, less traffic spikes)
```
ChrisBAshton, line 55: Is the "less traffic spikes"/"higher cache-hit ratio" because the banner can be cachebusted in isolation, as per point 3 in this list? And is the "quicker page loads" because of the reduced JS?

If so, I think it's worth splitting these out into separate points ("improved server-side performance", "improved client-side performance")

rfc-124-deploying-banners.md
```diff
@@ -0,0 +1,81 @@
...
+- the include fragment will have its own URL and thus independent VCL logic and caching TTLs
```
ChrisBAshton, line 56: This isn't a 'pro' in and of itself - could it be merged with the next line?

rfc-124-deploying-banners.md
```diff
@@ -0,0 +1,81 @@
...
+- ESI remove can be used to store fallback HTML content (which could be the rendered banner component) and therefore make local development easier and more consistent with production
+
+### Cons
+- origin must serve top-level page HTML as an uncompressed response (or support content-encoding negotiation) in order for Varnish to be able to do streaming ESI replacement
+- it is also not possible for Varnish to compress the response after ESI processing. Fortunately Fastly has a workaround for this which forces H2O (Fastly’s HTTP terminator and HTTP/2-3 server) to do the streaming compression (which supports both `gzip` and `brotli`). This can easily be enabled via a HTTP response header which is subsequently stripped by H2O. However, as Fastly bills from bytes out of Varnish this would result in you getting billed for the uncompressed bytes instead of the compressed version. We are on a flat fee plan and thus this should not be an issue
+- if ESI is not implemented correctly can make us vulnerable to different attacks (SSRF and XSS)
```
ChrisBAshton, line 63:
```suggestion
- if ESI is not implemented correctly, it could make us vulnerable to different attacks (SSRF and XSS)
```

Could you elaborate on this?

rfc-124-deploying-banners.md
```diff
@@ -0,0 +1,81 @@
+# Use Edge Side Include (ESI) to deploy banners on GOV.UK
+
+## Summary
+
+Using Edge Side Include (ESI) to deploy banners on GOV.UK will help us move away from `static`/`slimmer` while being able to roll out changes quickly.
+
+## Background
+A few types of site-wide banners can be activated to convey important information on GOV.UK.
```
vanitabarrett, line 8: Do we need to mention the survey banner here too? It's not something we tend to toggle on/off in the same way, but I assume we probably need to consider it here too.

kevindew:
Thanks for raising Alex, certainly a good idea 🥇. I haven't heard much about ESI for a long time, and it sounds a good application. Also, I'm all for making progress towards removing static.

So I do have a relatively major concern about this. It feels like this is a small idea that reduces some things from static but may not be suitable for everything. Dealing with this in isolation may well lead to multiple approaches to how we replace static, which ends up being confusing or inconsistent. Since there is already a [draft RFC](https://github.com/alphagov/govuk-rfcs/pull/118) surrounding working out an action plan to remove static I think it'd be best if we try find opportunities for a number of people to collaborate so that we can have a single RFC that draws out the plan to replace static. How do you feel about that?  

On the actual proposal a few thoughts I have are:

- If we did this in Fastly we'd lose the ability to see this on the draft stack
- If we do this in Fastly it makes it harder to consider a move to cloudfront which don't support ESI
- We could do this with our own varnish that is on the draft stack but Fastly would cache all responses so we'd lose some of the cache invalidation bonuses (which is probably my favourite side effect :-)
- We'd lose ability (or have to write hacks) to see usage in local dev or on heroku deploys
- We may end up with inconsistent approaches for static responsibilities (header and footer served one way, banners another)
- Seems like components gem usage could get rather delicate if apps need to include a component that may only be served via ESI, ie all frontend apps need to include CSS/JS that they might not need (unless that was another ESI)

And then I wondered what/where the endpoints for ESI might be? would we code them into a particular frontend app and/or use a content item? would the paths be public?

timblair:
One thing I don't see covered here is how this impacts how we [fall back to static mirrors](https://docs.publishing.service.gov.uk/manual/fall-back-to-mirror.html) in the case of GOV.UK origins being unavailable.

If both the scraping process, and the Fastly VCL, were configured correctly (content scraped without banners, and VCL applying ESIs to content regardless of which backend it was retrieved from), then this could enable us to display/hide/modify the emergency banners even if our origins were down, which would be great from an availability/messaging perspective.

rfc-124-deploying-banners.md
alex-ju: yes, it can be configuring which pages the banner should not display on, it could be the number of times it's being shown.

rfc-124-deploying-banners.md
alex-ju: Updated this point and split it up as suggested. For the client-side, I'm referring to the fact that the cached page assets can still be used by the browser when the banner is enabled/disabled as they won't be re-fingerprinted.

rfc-124-deploying-banners.md
alex-ju: Agree, they should be merged

rfc-124-deploying-banners.md
alex-ju: It's probably somehow similar to the cookie banner, in a sense that we anticipate updates instead of toggling, but it's a good point – I missed it. I'll try and look into it and include it as well.

alex-ju:
Thank you for your review @kevindew!

> So I do have a relatively major concern about this. It feels like this is a small idea that reduces some things from static but may not be suitable for everything. Dealing with this in isolation may well lead to multiple approaches to how we replace static, which ends up being confusing or inconsistent.

Indeed, this is not suitable for everything and it's not meant to be. The remnants we have in `static` at the moment are rather sparse elements that don't fit somewhere else and need to be somehow cetralised. If we try to find a solution that is suitable for everything I'm afraid we'll end up with another `static`. I raised this as a point of discussion for how we deal with banners as I understand they are one of the major concerns in the world of a fully componentised frontend for GOV.UK. 

> Since there is already a [draft RFC](https://github.com/alphagov/govuk-rfcs/pull/118) surrounding working out an action plan to remove static I think it'd be best if we try find opportunities for a number of people to collaborate so that we can have a single RFC that draws out the plan to replace static. How do you feel about that?

I mentioned the RFC 118 in this one too, and I don't think they're excluding each other, on the contrary. Spinning up ideas around different needs that static covers at the moment and how we can approach them with a cleaner and more effective architecture should be beneficial. Happy to join forces and/or collate a unified RFC – be that 118 or a new one.

> On the actual proposal a few thoughts I have are:
> * If we did this in Fastly we'd lose the ability to see this on the draft stack

Good point about the draft stack. I'm not sure how is different from the live stack. Wouldn't this be covered by the ESI fallback described in the RFC?

> * If we do this in Fastly it makes it harder to consider a move to cloudfront which don't support ESI

Indeed. I know close to nothing about the direction or prospects in this area.

> * We could do this with our own varnish that is on the draft stack but Fastly would cache all responses so we'd lose some of the cache invalidation bonuses (which is probably my favourite side effect :-)

Ha, interesting one. I only know found out about our Varnish on draft stack.

> * We'd lose ability (or have to write hacks) to see usage in local dev or on heroku deploys

I mentioned in the RFC how the fallback would work if ESI is not available, local dev and Heroku will render the component as fallback.

> * We may end up with inconsistent approaches for static responsibilities (header and footer served one way, banners another)

We already have inconsistent approaches. This proposal is envisioned with all components sitting in `govuk_publishing_components`, `header` and `footer` part of the `public_layout`, but with the recognition that the banners should have a more efficient mechanism for roll-out.

> * Seems like components gem usage could get rather delicate if apps need to include a component that may only be served via ESI, ie all frontend apps need to include CSS/JS that they might not need (unless that was another ESI)

I understand this concern. An alternative would be to have a scoped style tag pointing to the component's CSS served with the fragment.

> And then I wondered what/where the endpoints for ESI might be? would we code them into a particular frontend app and/or use a content item? would the paths be public?

In the RFC I was suggesting a `banner` application on a frontend machine will serve as endpoint. It could be implemented in an existing application too, but again, separation of concerns – otherwise we'll end up with another `static`.

alex-ju:
> One thing I don't see covered here is how this impacts how we [fall back to static mirrors](https://docs.publishing.service.gov.uk/manual/fall-back-to-mirror.html) in the case of GOV.UK origins being unavailable.

Thanks for pointing this out @timblair. I admit I haven't thought about the mirrors. Will try and have a more in-depth look and expose how that would look like.

Nooshu:
> If we do this in Fastly it makes it harder to consider a move to cloudfront which don't support ESI

I was under the impression a move was no longer being considered due to the time and cost implications involved in the move. As mentioned in [RFC 109](https://github.com/alphagov/govuk-rfcs/pull/109), I'd also not recommend CloudFront from a web performance point of view due to their [broken HTTP/2 prioritisation](https://github.com/andydavies/http2-prioritization-issues#getting-started) and [weird handling of OCSP](https://github.com/alphagov/govuk-rfcs/pull/109#issuecomment-529460598).

kevindew:
> > Since there is already a [draft RFC](https://github.com/alphagov/govuk-rfcs/pull/118) surrounding working out an action plan to remove static I think it'd be best if we try find opportunities for a number of people to collaborate so that we can have a single RFC that draws out the plan to replace static. How do you feel about that?
> 
> I mentioned the RFC 118 in this one too, and I don't think they're excluding each other, on the contrary. Spinning up ideas around different needs that static covers at the moment and how we can approach them with a cleaner and more effective architecture should be beneficial. Happy to join forces and/or collate a unified RFC – be that 118 or a new one.

Thanks that sounds good. I think it's useful to have a balance between a roadmap/loose plan about how a future without static looks and then also to be chipping away at bits iteratively. My concern is that without the loose plan we end up with things that are rather inconsistent, but maybe I'm over concerned? Anyway happy to work with you on that.

We've obviously got the problem with 118 that the team has been paused and we don't know when an implementation might come (I'll go close the PR after this actually). Does this work fit any team which helps bump it along implementation wise?

> Good point about the draft stack. I'm not sure how is different from the live stack. Wouldn't this be covered by the ESI fallback described in the RFC?

Yes, I believe it'd be covered by the fallback so long as ESI is disabled on our internal varnish - and I'll just apologise that I didn't take in the fallback on my first reading 🙈 so I'm sorry some of my comments referenced things covered.

It might be worth having a wrapper function that handles the ESI fallback which, conditional on environment variable, can request the include from frontend. That could allow the draft stack to be the same as live without needing ESI.

> > * We could do this with our own varnish that is on the draft stack but Fastly would cache all responses so we'd lose some of the cache invalidation bonuses (which is probably my favourite side effect :-)
> 
> Ha, interesting one. I only know found out about our Varnish on draft stack.

Sure, it's a real big confusion point that we have two layers of varnish in the stack. I'd be all for getting rid of our internal one.
 
> > * We'd lose ability (or have to write hacks) to see usage in local dev or on heroku deploys
> 
> I mentioned in the RFC how the fallback would work if ESI is not available, local dev and Heroku will render the component as fallback.

Sorry, my bad again - must absorb more information before commenting.

> > * We may end up with inconsistent approaches for static responsibilities (header and footer served one way, banners another)
> 
> We already have inconsistent approaches. This proposal is envisioned with all components sitting in `govuk_publishing_components`, `header` and `footer` part of the `public_layout`, but with the recognition that the banners should have a more efficient mechanism for roll-out.

👍 

> > And then I wondered what/where the endpoints for ESI might be? would we code them into a particular frontend app and/or use a content item? would the paths be public?
> 
> In the RFC I was suggesting a `banner` application on a frontend machine will serve as endpoint. It could be implemented in an existing application too, but again, separation of concerns – otherwise we'll end up with another `static`.

We could maybe use the frontend app rather than creating a new one (I'm of the opinion we already have too many frontend apps with similar responsibilities so take a view towards consolidation over fragmentation)? In RFC 118 we have a similar problem with what to do with error pages without static and using frontend was the leading suggestion.

Where would content for the banners be held, in app or in the component gem?

> > If we do this in Fastly it makes it harder to consider a move to cloudfront which don't support ESI
> 
> I was under the impression a move was no longer being considered due to the time and cost implications involved in the move. As mentioned in [RFC 109 comments](https://github.com/alphagov/govuk-rfcs/pull/109), I'd also not recommend CloudFront from a web performance point of view due to their [broken HTTP/2 prioritisation](https://github.com/andydavies/http2-prioritization-issues#getting-started) and [weird handling of OCSP stapling](https://github.com/alphagov/govuk-rfcs/pull/109#issuecomment-529460598).

Sure, I don't know myself if there are plans or ideas leaning that way. I know we've been using cloudfront for the new coronavirus services and know we've still got the unresolved problems about how to integrate a WAF. However, no problem if we're happy to continue with Fastly for the foreseeable future.

rfc-124-deploying-banners.md
alex-ju: Correct me if I'm wrong, @vanitabarrett, but as far as I can see the survey banner is even more different than the other ones in terms of where the markup sits – [being exclusively generated by JavaScript](https://github.com/alphagov/static/blob/master/app/assets/javascripts/surveys.js#L15-L21).

alex-ju:
> It would be good to see any alternatives considered - e.g. storing banner as a content item and relying on link expansion to update all pages - and their pros and cons.

@ChrisBAshton here are the alternatives I considered from a frontend perspective, before raising this:
- **do nothing** and continue to converge different approaches **and consolidate the frontend code** into components, meaning that `static` will remain the place where the content and toggle mechanism lives, while the design/functional updates will be done in each component in `govuk_publishing_components`. Deployment on this approach will require, at most, update and release of the gem, update and deploy of `static` and running the rake task.
- **use a content item** (as suggested by you and Kevin in the comments and drafted in RFC #118) and continue to converge different approaches **and consolidate the frontend code** into components (similarly to above). We rejected the 'Publishing API link expansion' approach _because a change to the data would take hours to propagate for each item_ and we were leaning more towards using `homepage` for any global data, _since any special content item would need a base path to be present in the content store_. Whatever approach we would take for the content and the toggle mechanism, they would be shifted from `static` in one of the apps (I think `frontend` was suggested) while the design/functional updates will be done in each component in `govuk_publishing_components`. It's harder to sketch how a deploy would look like since it's tightly coupled with the detailed decisions we'd make around the content, but the design/functional changes would still require updates and deployments of each application, which is one of our current pitfalls.

Of course, a similar approach to the one described in this RFC could be achieved without ESI, by using an asynchronous JavaScript request for the fragment, but this won't be in line with our development principles (referring to progressive enhancement and making the info available to everyone) – which I personally highly value.

Happy to hear other people's thoughts or alternatives to this too – there must be a few more, but I might not see them.

rfc-124-deploying-banners.md
alex-ju: _If_ an attacker manages to add an ESI include tag to the HTTP response, through publishing platforms or (reflected) XSS, they can perform SSRF attacks in the context of a surrogate server. Of course, we have cyber to test these things, but it's worth bearing in mind so we close as many doors as possible from the start.

rfc-124-deploying-banners.md
vanitabarrett: @alex-ju That's true, I hadn't realised it worked in that way. But I assume it will still need a central place to live if we're retiring Static?

alex-ju: @vanitabarrett yes. I had a more in-depth look and, with minimal changes, it could be componentised 'as is', then potentially brought more in-line with the other banners and components in general through a few iterations. Since that's outside of the scope for this RFC I only added a commit to cover it in the background section. Thank you again for reminding me about it.

alex-ju:
This RFC is due. I will close it with the main takeaway that we want to tackle `static` holistically rather than trying to address various parts of it.

Thanks to everyone involved, really appreciate it.

PR #126, opened 2020-06-11
Custom configuration for Dependabot

thomasleese:
[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/dependency-management/rfc-126-custom-configuration-for-dependabot.md)

**Deadline:** 2020-06-25

### DIFF
--- a/rfc-126-custom-configuration-for-dependabot.md
+++ b/rfc-126-custom-configuration-for-dependabot.md
@@ -241,7 +241,7 @@ Using [the statistics from above](#Statistics), we can make some estimates on wh
 
 - 7378 PRs wouldn't have been raised (10216 rather than the 17526 we've had so far).
 - The mean time to merge would have decreased from almost 3 days to just under 2 days.
-- The ignored libraries take on average 3.5 days to merge.
+- The ignored libraries alone would have taken on average 3.5 days to merge.
 
 In theory, by limiting the number of libraries that get updated by Dependabot, our applications would end up less up to date. However, by prioritising certain libraries over others, we should end up in a situation where our applications are actually _more_ up to date in the areas we care about (security, frameworks, internal libraries).
 
@@ -250,3 +250,5 @@ We should expect to see the average time for PRs to be merged in (and in particu
 A less measurable metric is that we would expect developers to be spending less time fixing issues in Dependabot PRs.
 
 If we see a situation where most days we don't have any new Dependabot PRs, we could think about losening these restrictions and opening up the number of libraries we want to received updates on.
+
+This RFC proposes that we review the situation six months after the deployment of the global Dependabot configuration to ensure that this has made a difference to the workload. This will be achieved by talking to developers and using some updated statistics. This review will be owned by the Platform Health team.
--- a/rfc-126-custom-configuration-for-dependabot.md
+++ b/rfc-126-custom-configuration-for-dependabot.md
@@ -170,6 +170,7 @@ update_configs:
       - match: { dependency_name: factory_bot_rails }
       - match: { dependency_name: jasmine }
       - match: { dependency_name: rails }
+      - match: { dependency_name: rspec }
       - match: { dependency_name: rspec-rails }
       - match: { dependency_name: sass-rails }
   - package_manager: javascript
@@ -189,7 +190,7 @@ update_configs:
 
 Most of our applications use the same internal and framework libraries, so we'll want a single global configuration across all our apps. At the same time, there might be some libraries used by particular apps which fit within one of the three categories.
 
-Maintaining this file across each of our apps has the potential for it to become out of date with our global configuration. To solve this problem, this RFC proposes that we add functionality to [govuk-saas-config] which ensures the configuration file in each repo is up to date with a single global configuration file. Keeping up to date means ensuring that the individual repo configuration file contains at least the global configuration options plus any extra custom configuration for the app itself. This should allow some apps to include framework libraries that are used by specific apps (for example `sinatra`) but not used wider across all our apps.
+Maintaining this file across each of our apps has the potential for it to become out of date with our global configuration. To solve this problem, this RFC proposes that we add functionality to [govuk-saas-config] which ensures the configuration file in each repo is up to date with a single global configuration file. Keeping up to date means ensuring that the individual repo configuration file contains at least the global configuration options plus any extra framework libraries that are used by specific apps (for example `sinatra`) but not used wider across all our apps.
 
 Due to the fact that the configuration file is stored in the Git repo, [govuk-saas-config] will need to raise a PR to keep the configuration in sync rather than writing to the repo directly. This is similar to a script we have for [upgrading Ruby].
 
@@ -244,6 +245,8 @@ Using [the statistics from above](#Statistics), we can make some estimates on wh
 - The ignored libraries alone would have taken on average 3.5 days to merge.
 
 In theory, by limiting the number of libraries that get updated by Dependabot, our applications would end up less up to date. However, by prioritising certain libraries over others, we should end up in a situation where our applications are actually _more_ up to date in the areas we care about (security, frameworks, internal libraries). There is a risk involved in not being able to access new features or fixes from ignored libraries, however the reason the library has been ignored is because we're making little use of it in the app (otherwise it should be a framework library). We can always manually upgrade these libraries when we want to, and that should still be encouraged when doing a major update (for example a Rails upgrade).
+
+Another potential risk is that people find the global configuration confusing if it contains libraries not relevant to the app. This may lead to people editing the config and then it being reset the next day. One way to mitigate this would be to add an explanatory comment to the file.
 
 We should expect to see the average time for PRs to be merged in (and in particular security PRs) to reduce. We should also expect to see a reduction in the number of open PRs, and also a reduction in the number of deployments.
 
--- a/rfc-126-custom-configuration-for-dependabot.md
+++ b/rfc-126-custom-configuration-for-dependabot.md
@@ -37,11 +37,15 @@ By having so many open PRs, it becomes difficult to be able to prioritise import
 
 I pulled together [some statistics on our Dependabot usage][dependabot-stats]:
 
-A total of 17,526 PRs have been opened since we started using Dependabot. Of which, 7,378 would be considered external libraries (not security fixes, not our own libraries and not core requirements, for example Rails).
-
-The majority of PRs are merged within a few days (on average 2.5 days). However, the PRs that require manual intervention (for example, failing tests) tend to take a significant amount of time to be merged (on order of months). Because of this talking in terms of averages tends to be less representative of the actual situation, but it can be useful to make comparisons.
+- A total of 17,526 PRs have been opened since we started using Dependabot.
+- 8,528 are from [our own internal libraries][dependabot-stats-internal] (for example `govuk_publishing_components`).
+- 1013 are [core requirements for our apps][dependabot-stats-framework] (for example `rails`).
+- 675 are security fixes.
+- 7,378 are other libraries.
 
 [dependabot-stats]: https://github.com/thomasleese/dependabot-stats
+[dependabot-stats-internal]: https://github.com/thomasleese/dependabot-stats/blob/bf2d5848f99516962c1b3742fd16883f34541c86/dependabot_stats/analyse.py#L125
+[dependabot-stats-framework]: https://github.com/thomasleese/dependabot-stats/blob/bf2d5848f99516962c1b3742fd16883f34541c86/dependabot_stats/analyse.py#L126
 
 <details>
     <summary>Full results</summary>
@@ -236,17 +240,13 @@ This would allow us to remove the hardest PRs from updates, but it would require
 
 ## Consequences, risks and measuring success
 
-Using [the statistics from above](#Statistics), we can make some estimates on what affect this would have had if it had been enabled from the beginning:
-
-- 7378 PRs wouldn't have been raised (10216 rather than the 17526 we've had so far).
-- The mean time to merge would have decreased from almost 3 days to just under 2 days.
-- The ignored libraries alone would have taken on average 3.5 days to merge.
+Using [the statistics from above](#Statistics), we can make some estimates on what affect this would have had if it had been enabled from the beginning: 7378 PRs wouldn't have been raised (10216 rather than the 17526 we've had so far).
 
 In theory, by limiting the number of libraries that get updated by Dependabot, our applications would end up less up to date. However, by prioritising certain libraries over others, we should end up in a situation where our applications are actually _more_ up to date in the areas we care about (security, frameworks, internal libraries). There is a risk involved in not being able to access new features or fixes from ignored libraries, however the reason the library has been ignored is because we're making little use of it in the app (otherwise it should be a framework library). We can always manually upgrade these libraries when we want to, and that should still be encouraged when doing a major update (for example a Rails upgrade).
 
 Another potential risk is that people find the global configuration confusing if it contains libraries not relevant to the app. This may lead to people editing the config and then it being reset the next day. One way to mitigate this would be to add an explanatory comment to the file.
 
-We should expect to see the average time for PRs to be merged in (and in particular security PRs) to reduce. We should also expect to see a reduction in the number of open PRs, and also a reduction in the number of deployments.
+We should also expect to see a reduction in the number of open PRs, and also a reduction in the number of deployments.
 
 A less measurable metric is that we would expect developers to be spending less time fixing issues in Dependabot PRs.
 
--- a/rfc-126-custom-configuration-for-dependabot.md
+++ b/rfc-126-custom-configuration-for-dependabot.md
@@ -13,7 +13,7 @@ A few solutions have been implemented in the past to make this better:
 - [govuk-dependencies] shows the list of open PRs grouped by application, team or gem.
 - [govuk-dependencies] has a Slack bot which automatically messages in each team's channel with a list of open PRs.
 - We [allowed merging Dependabot PRs without requiring two approvals][rfc-103].
-- [Sharing the responsibility for Dependabot PRs across multiple times][dependencies-team].
+- [Sharing the responsibility for Dependabot PRs across multiple teams][dependencies-team].
 - We've configured Dependabot to [only update top-level dependencies in some repos][top-level-dependencies].
 - [bulk-merger] was developed to make it easier to review and merge Dependabot PRs in bulk.
 
--- a/rfc-126-custom-configuration-for-dependabot.md
+++ b/rfc-126-custom-configuration-for-dependabot.md
@@ -1,0 +1,246 @@
+# Custom configuration for Dependabot
+
+## Summary
+
+Introduce stricter configuration for Dependabot to reduced number of open PRs, number of deployments and effort required to ensure apps always use the latest version of libraries that we care about.
+
+## Problem
+
+Dependabot was enabled across all our applications in 2018. Due to the number of applications we have and libraries we use, keeping on top of the number of PRs raised by Dependabot has been difficult.
+
+A few solutions have been implemented in the past to make this better:
+
+- [govuk-dependencies] shows the list of open PRs grouped by application, team or gem.
+- [govuk-dependencies] has a Slack bot which automatically messages in each team's channel with a list of open PRs.
+- We [allowed merging Dependabot PRs without requiring two approvals][rfc-103].
+- [Sharing the responsibility for Dependabot PRs across multiple times][dependencies-team].
+
+[govuk-dependencies]: http://govuk-dependencies.herokuapp.com/
+[rfc-103]: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-103-merge-dependabot-pull-requests-with-a-single-review.md
+[dependencies-team]: https://github.com/alphagov/govuk-developer-docs/pull/2168
+
+Although these have made improvements to the process, the number of open PRs remains unmanageable. At the time of writing, there are 135 open PRs, of which there are [8 labelled as security fixes][security-prs], the oldest being 21 days old. Dependabot is currently configured to have at most 5 open PRs per repo, which means this number is not the true number and often new PRs will open the moment existing ones are closed down.
+
+[security-prs]: https://github.com/pulls?q=is%3Apr+author%3Aapp%2Fdependabot-preview+user%3Aalphagov+archived%3Afalse+is%3Aopen+label%3Asecurity
+
+This high number of open PRs is felt particularly hard by the Platform Health team, as they currently [own 55 out of 73 applications][application-owners].
+
+[application-owners]: https://docs.publishing.service.gov.uk/apps/by-team.html
+
+By having so many open PRs, it becomes difficult to be able to prioritise important updates (bug fixes, security fixes) over new versions which simply adds new features (which our apps don't necessarily use anyway). This leads to important updates being left for long periods of time and our desire to close down all the PRs means developers spend time investigating and fixing complex problems without any benefit.
+
+### Statistics
+
+I pulled together [some statistics on our Dependabot usage][dependabot-stats]:
+
+The majority of PRs are merged within a few days (on average 2.5 days). However, the PRs that require manual intervention (for example, failing tests) tend to take a significant amount of time to be merged (on order of months). Because of this talking in terms of averages tends to be less representative of the actual situation, but it can be useful to make comparisons.
+
+The libraries that take the longest time to merge are `acts-as-taggable-on`, `paper_trail`, `rails-controller-testing`, `mini_magick` and `gds-sso`. Our own libraries are merged in on average after 1.5 days, whereas third party libraries are merged in after 4 days.
+
+The most problematic libraries (ones where there is a higher proportion of PRs taking a long time to be merged in) tend to those that are not owned by us and not a core requirement of the app (for example `acts-as-taggable-on`, `paper_trail`, `rails-controller-testing`, `mini_magick` and `excon`.
+
+[dependabot-stats]: https://github.com/thomasleese/dependabot-stats
+
+<details>
+    <summary>Full results</summary>
+
+```
+All PRs
+=======
+Mean time to merge: 2 days, 15:35:42.556088
+Max time to merge: 189 days, 5:50:39
+Top 5 longest libraries to merge: acts-as-taggable-on (146 days, 3:40:01), paper_trail (85 days, 7:45:33), rails-controller-testing (84 days, 3:29:26), mini_magick (41 days, 4:15:20), gds-sso (38 days, 17:12:39)
+Top 5 quickest libraries to merge: actionview (0:05:42), miller-columns-element (0:05:36), faraday_middleware (0:03:30), webrick (0:01:17), nltk (0:01:14)
+
+Security PRs
+============
+Mean time to merge: 1 day, 22:01:29.564444
+Max time to merge: 40 days, 18:54:18
+Top 5 longest libraries to merge: secure_headers (28 days, 4:50:14), doorkeeper (19 days, 7:14:41), excon (15 days, 7:26:22), jquery-rails (13 days, 9:40:30), rack (11 days, 14:55:34)
+Top 5 quickest libraries to merge: json (0:52:52), sprockets (0:38:34), rack-protection (0:19:10), actionview (0:05:42), webrick (0:01:17)
+
+Non-security PRs
+================
+Mean time to merge: 2 days, 16:17:56.279271
+Max time to merge: 189 days, 5:50:39
+Top 5 longest libraries to merge: acts-as-taggable-on (146 days, 3:40:01), paper_trail (85 days, 7:45:33), rails-controller-testing (84 days, 3:29:26), mini_magick (41 days, 4:15:20), excon (41 days, 0:57:48)
+Top 5 quickest libraries to merge: docker-compose (0:12:51), vcr (0:10:44), miller-columns-element (0:05:36), faraday_middleware (0:03:30), nltk (0:01:14)
+
+Security Libraries
+==================
+Mean time to merge: 1 day, 22:01:29.564444
+Max time to merge: 40 days, 18:54:18
+Top 5 longest libraries to merge: secure_headers (28 days, 4:50:14), doorkeeper (19 days, 7:14:41), excon (15 days, 7:26:22), jquery-rails (13 days, 9:40:30), rack (11 days, 14:55:34)
+Top 5 quickest libraries to merge: json (0:52:52), sprockets (0:38:34), rack-protection (0:19:10), actionview (0:05:42), webrick (0:01:17)
+
+Internal Libraries
+==================
+Mean time to merge: 1 day, 12:06:12.839822
+Max time to merge: 154 days, 6:12:55
+Top 5 longest libraries to merge: gds-sso (38 days, 17:12:39), govuk_app_config (13 days, 17:41:26), govuk_test (5 days, 0:17:42), plek (2 days, 8:29:52), govspeak (1 day, 9:19:06)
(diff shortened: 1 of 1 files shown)


rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,246 @@
...
+      - match: { dependency_name: govuk_test }
+      - match: { dependency_name: rubocop-govuk }
+      - match: { dependency_name: plek }
+      - match: { dependency_name: scss_lint-govuk }
+      # Framework gems
+      - match: { dependency_name: factory_bot_rails }
+      - match: { dependency_name: jasmine }
+      - match: { dependency_name: rails }
+      - match: { dependency_name: rspec-rails }
+      - match: { dependency_name: sass-rails }
```
bilbof, line 168: We could make use of wildcards and the `dependency_type` field.

```
      - match:
          dependency_name: "govuk*"
          dependency_type: "production"
```

This will include all gems beginning with "govuk" and that aren't in the "development" group.

Not bumping gems used only for local development might cut down on the number of unnecessary bumps.

benthorner: +1 for an asterisk - that's some nice future proofing in there.

I think we should bump dev dependencies:

- We want to keep up-to-date with framework gems like RSpec, which we write a lot of code with.
- We want to keep up-to-date with internal dev gems, like RuboCop GOV.UK. No one wants to raise all those PRs manually.

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,246 @@
...
+
+[security-prs]: https://github.com/pulls?q=is%3Apr+author%3Aapp%2Fdependabot-preview+user%3Aalphagov+archived%3Afalse+is%3Aopen+label%3Asecurity
+
+This high number of open PRs is felt particularly hard by the Platform Health team, as they currently [own 55 out of 73 applications][application-owners].
+
+[application-owners]: https://docs.publishing.service.gov.uk/apps/by-team.html
+
+By having so many open PRs, it becomes difficult to be able to prioritise important updates (bug fixes, security fixes) over new versions which simply adds new features (which our apps don't necessarily use anyway). This leads to important updates being left for long periods of time and our desire to close down all the PRs means developers spend time investigating and fixing complex problems without any benefit.
```
benthorner, line 30:
```suggestion
By having so many open PRs, it becomes difficult to be able to prioritise important updates (bug fixes, security fixes, internal updates) over new versions which simply adds new features (which our apps don't necessarily use anyway). This leads to important updates being left for long periods of time and our desire to close down all the PRs means developers spend time investigating and fixing complex problems without any benefit.
```

rfc-126-custom-configuration-for-dependabot.md
bilbof: Oh yeah, that's true. Though I wonder if they should be in the `test` group; I'm not sure if `production` includes this (there's no `test` dependabot type).

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,246 @@
...
+
+Dependabot was enabled across all our applications in 2018. Due to the number of applications we have and libraries we use, keeping on top of the number of PRs raised by Dependabot has been difficult.
+
+A few solutions have been implemented in the past to make this better:
+
+- [govuk-dependencies] shows the list of open PRs grouped by application, team or gem.
+- [govuk-dependencies] has a Slack bot which automatically messages in each team's channel with a list of open PRs.
+- We [allowed merging Dependabot PRs without requiring two approvals][rfc-103].
+- [Sharing the responsibility for Dependabot PRs across multiple times][dependencies-team].
+
```
benthorner, line 17: We've also configured Dependabot to only update top-level dependencies, for many repos.

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,246 @@
...
+By having so many open PRs, it becomes difficult to be able to prioritise important updates (bug fixes, security fixes) over new versions which simply adds new features (which our apps don't necessarily use anyway). This leads to important updates being left for long periods of time and our desire to close down all the PRs means developers spend time investigating and fixing complex problems without any benefit.
+
+### Statistics
+
+I pulled together [some statistics on our Dependabot usage][dependabot-stats]:
+
+The majority of PRs are merged within a few days (on average 2.5 days). However, the PRs that require manual intervention (for example, failing tests) tend to take a significant amount of time to be merged (on order of months). Because of this talking in terms of averages tends to be less representative of the actual situation, but it can be useful to make comparisons.
+
+The libraries that take the longest time to merge are `acts-as-taggable-on`, `paper_trail`, `rails-controller-testing`, `mini_magick` and `gds-sso`. Our own libraries are merged in on average after 1.5 days, whereas third party libraries are merged in after 4 days.
```
benthorner, line 38: What's the conclusion here?

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,246 @@
...
+
+[application-owners]: https://docs.publishing.service.gov.uk/apps/by-team.html
+
+By having so many open PRs, it becomes difficult to be able to prioritise important updates (bug fixes, security fixes) over new versions which simply adds new features (which our apps don't necessarily use anyway). This leads to important updates being left for long periods of time and our desire to close down all the PRs means developers spend time investigating and fixing complex problems without any benefit.
+
+### Statistics
+
+I pulled together [some statistics on our Dependabot usage][dependabot-stats]:
+
+The majority of PRs are merged within a few days (on average 2.5 days). However, the PRs that require manual intervention (for example, failing tests) tend to take a significant amount of time to be merged (on order of months). Because of this talking in terms of averages tends to be less representative of the actual situation, but it can be useful to make comparisons.
```
benthorner, line 36: What's the takeaway from this? It seems like a caveat for the subsequent paragraphs. Perhaps we could just provide more info there (mean and max).

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,246 @@
...
+
+The libraries that take the longest time to merge are `acts-as-taggable-on`, `paper_trail`, `rails-controller-testing`, `mini_magick` and `gds-sso`. Our own libraries are merged in on average after 1.5 days, whereas third party libraries are merged in after 4 days.
+
+The most problematic libraries (ones where there is a higher proportion of PRs taking a long time to be merged in) tend to those that are not owned by us and not a core requirement of the app (for example `acts-as-taggable-on`, `paper_trail`, `rails-controller-testing`, `mini_magick` and `excon`.
+
```
benthorner, line 41: Could we also propose an argument in terms of sheer volume here (vs. individual timings)? The sheer volume of external gem PRs may help to support the conclusion of the RFC.

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,246 @@
...
+- We require approvals for all our PRs. Merging in the PR isn't what takes up the time, so having them merge automatically would only shave off a small amount of time.
+- The auto-merge functionality has to be enabled across the entire Dependabot account. We use a shared account across GDS. For this reason, and others, we [weren't able to turn it on][auto-merge-gds-way].
+
+[auto-merge]: https://dependabot.com/blog/automatic-pull-request-merging/
+[auto-merge-discussion]: https://docs.google.com/document/d/18P86EFSZdG8Uv5scbGzVoO8KhA3hkKp80UlzNoIctNE/edit
+[auto-merge-gds-way]: https://github.com/alphagov/gds-way/pull/428#discussion_r426502446
+
+#### Limit updates to lockfile
+
+Another option might be to limit Dependabot to only raise PRs against the `Gemfile.lock` file. The lockfile is used to pin a specific version, whereas the `Gemfile` can contain a wider version range (for example `~> 1.0` means `1.0 all the way up to but not including 2.0`).
```
benthorner, line 206: We could add a little more to this approach: gems we want to update would have little or no version restriction; gems we don't want to update would have a (more) precise restriction, to prevent Dependabot PRs, and signify that we're content with basically any working version of the gem.

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,246 @@
...
+- The mean time to merge would have decreased from almost 3 days to just under 2 days.
+- The ignored libraries take on average 3.5 days to merge.
+
+In theory, by limiting the number of libraries that get updated by Dependabot, our applications would end up less up to date. However, by prioritising certain libraries over others, we should end up in a situation where our applications are actually _more_ up to date in the areas we care about (security, frameworks, internal libraries).
+
+We should expect to see the average time for PRs to be merged in (and in particular security PRs) to reduce. We should also expect to see a reduction in the number of open PRs, and also a reduction in the number of deployments.
+
+A less measurable metric is that we would expect developers to be spending less time fixing issues in Dependabot PRs.
+
+If we see a situation where most days we don't have any new Dependabot PRs, we could think about losening these restrictions and opening up the number of libraries we want to received updates on.
```
benthorner, line 242: What do you think about some kind of review of the situation in X months?

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,242 @@
...
+
+This would allow us to remove the hardest PRs from updates, but it would require us to be more on top of maintaining the list. I'm also not sure how it would work with enabling security updates regardless.
+
+[dependabot-ignored-updates]: https://dependabot.com/docs/config-file/#ignored_updates
+
+## Consequences and measuring success
+
+Using [the statistics from above](#Statistics), we can make some estimates on what affect this would have had if it had been enabled from the beginning:
+
+- 7378 PRs wouldn't have been raised (10216 rather than the 17526 we've had so far).
```
benthorner, line 232: These are really helpful to know! I wonder if we should mention this earlier in the RFC, as it's also an argument for the proposal, as much as a consequence.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Oh, I actually wasn't aware of this. Do we have some sort of evidence for this I can link to?

benthorner: I don't think so, it's just something I noticed, which isn't all that consistent. I enabled it for Whitehall yesterday on the assumption that it's something we _intended_ to do.

Arguably we're just taking this one step further: all the gems we're proposing to monitor are top-level dependencies. Worth noting this setting hasn't prevented security updates.

thomasleese: Added in 4b68907.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Thanks! Added in d984dab.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Good point, I've added that in d28ed61.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Yeah, I couldn't decide on the best way of framing these statistics. I thought it would be clearest to use a single number for comparisons between things, but just clarify that the single number is perhaps not the most representative. Another caveat I've just thought of is that mean time to merge ignores the "time between when the PR is raised and work starts to get the tests passing". This wouldn't be solved by including both the mean and the max. 

Ideally, we'd want to show a graph or exclude the outliers in the results (i.e. the one offs that take a really long time and the ones that get merged in within 1 hour always) each time they're used in the RFC, but at the same time I wanted the statistics to supplement the RFC without being the main argument. This is why I settled with explaining a caveat at the top and then continuing to use a single number that easy to understand where it comes from in the body of the RFC.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: I've added a sentence to this section on the number of PRs in d28ed61. Is this what you had in mind, or would you prefer to go further (i.e. calculate the entire time to merge for each category). I also found it was a bit circular to talk about these statistics in the problem section, because by narrowing it down to external gem PRs, we've already started using the solution proposed to frame our problem...

rfc-126-custom-configuration-for-dependabot.md
thomasleese: That sounds like a good idea to me! I've added a sentence about this in 32511dd.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: I've reworded this section in 7826c68 which I think covers this better now!

thomasleese:
@benthorner @bilbof Thanks for your comments. I've added some clarification to the RFC which I think should cover your general points.

rfc-126-custom-configuration-for-dependabot.md
benthorner: Um, I should have asked: "How does this influence the RFC?".

Perhaps we could delete the original first paragraph, as it doesn't seem to have an effect on the outcome - we're willing to spend the time if it's worth it.

rfc-126-custom-configuration-for-dependabot.md
benthorner: Hmm...I think I'm actually confused about the relevance of timing stats in this RFC, including in the consequences. We don't have any targets for time to merge, and now I'm finding it surprising to see that as a consequence.

Since the time to merge can be due to any and all of volume, variety, and difficulty, I'm not sure we'll see the reduction we're expecting. In any case, it's unclear why "3 days to merge" is much worse than "2 days to merge".

So I'm thinking:

- This is a minor comment on the content here, since it's informational but doesn't seem particularly relevant when you're reading through.
- This is a somewhat minor comment on the importance of mean time to merge as a metric we're looking improve; the count of open PRs seems much easier to work with.

rfc-126-custom-configuration-for-dependabot.md
benthorner: Sure, I forget we want to be unbiased in our statement of the problem! I think it's fair to highlight that "random" external dependencies are a large chunk of the effort. Perhaps we should be even fairer, and highlight our internal libraries as the other main chunk; this helps frame the two main categories of PR we may want to tweak config for.

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -187,7 +189,7 @@ update_configs:
...
 
-Maintaining this file across each of our apps has the potential for it to become out of date with our global configuration. To solve this problem, this RFC proposes that we add functionality to [govuk-saas-config] which ensures the configuration file in each repo is up to date with a single global configuration file. Keeping up to date means ensuring that the individual repo configuration file contains at least the global configuration options plus any extra custom configuration for the app itself.
+Maintaining this file across each of our apps has the potential for it to become out of date with our global configuration. To solve this problem, this RFC proposes that we add functionality to [govuk-saas-config] which ensures the configuration file in each repo is up to date with a single global configuration file. Keeping up to date means ensuring that the individual repo configuration file contains at least the global configuration options plus any extra custom configuration for the app itself. This should allow some apps to include framework libraries that are used by specific apps (for example `sinatra`) but not used wider across all our apps.
```
benthorner, line 192:
```suggestion
Maintaining this file across each of our apps has the potential for it to become out of date with our global configuration. To solve this problem, this RFC proposes that we add functionality to [govuk-saas-config] which ensures the configuration file in each repo is up to date with a single global configuration file. Keeping up to date means ensuring that the individual repo configuration file contains at least the global configuration options plus any extra framework libraries that are used by specific apps (for example `sinatra`) but not used wider across all our apps.
```

benthorner: Pretty minor: the ambiguity before is whether the app could also tweak the top-level configuration, which shouldn't be necessary. I can't see why an app would need to tweak the config apart from for framework dependencies, so it's helpful if we can be more specific and say that.

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -233,15 +235,15 @@ This would allow us to remove the hardest PRs from updates, but it would require
...
 - The ignored libraries alone would have taken on average 3.5 days to merge.
 
-In theory, by limiting the number of libraries that get updated by Dependabot, our applications would end up less up to date. However, by prioritising certain libraries over others, we should end up in a situation where our applications are actually _more_ up to date in the areas we care about (security, frameworks, internal libraries).
+In theory, by limiting the number of libraries that get updated by Dependabot, our applications would end up less up to date. However, by prioritising certain libraries over others, we should end up in a situation where our applications are actually _more_ up to date in the areas we care about (security, frameworks, internal libraries). There is a risk involved in not being able to access new features or fixes from ignored libraries, however the reason the library has been ignored is because we're making little use of it in the app (otherwise it should be a framework library). We can always manually upgrade these libraries when we want to, and that should still be encouraged when doing a major update (for example a Rails upgrade).
```
benthorner, line 246: I could have been even more helpful and said: another risk is that people find the global config confusing, since it will list gems that aren't relevant to the repo. The impact is that people may delete pieces of the config, only to have a PR raised back at them. This seems pretty easy to address, by including an explanatory comment.

Perhaps this is just an implementation detail, more than a risk!

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,254 @@
...
+      - match: { dependency_name: govspeak }
+      - match: { dependency_name: govuk* }
+      - match: { dependency_name: rubocop-govuk }
+      - match: { dependency_name: plek }
+      - match: { dependency_name: scss_lint-govuk }
+      # Framework gems
+      - match: { dependency_name: factory_bot_rails }
+      - match: { dependency_name: jasmine }
+      - match: { dependency_name: rails }
+      - match: { dependency_name: rspec-rails }
```
benthorner, line 173:
```suggestion
      - match: { dependency_name: rspec-rails }
      - match: { dependency_name: rspec }
```

We probably want this as well, to account for non-Rails repos, of which we have quite a few.

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,254 @@
...
+      # Internal gems
+      - match: { dependency_name: gds-api-adapters }
+      - match: { dependency_name: gds-sso }
+      - match: { dependency_name: govspeak }
+      - match: { dependency_name: govuk* }
+      - match: { dependency_name: rubocop-govuk }
+      - match: { dependency_name: plek }
+      - match: { dependency_name: scss_lint-govuk }
+      # Framework gems
+      - match: { dependency_name: factory_bot_rails }
```
benthorner, line 170:
```suggestion
      - match: { dependency_name: factory_bot }
```

Arguably most of our code relies on the core dependency, rather than the smaller Rails-specific layer on top of it.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Yeah, I think this is worth mentioning. I've added a sentence in 19c6952555d3d39563ea49286a6df361d502e9b1.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Sure, added in 6af5beaa9ec1436eba5fda5d37bf15804d0ecac8.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Yep, updated in 19c6952.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: I've added your suggestion in 19c6952 which I think makes this clearer, thanks!

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Yeah, that's fair. I've removed the paragraph in fb3b211f1ecdd8942e20e0c7bf2d3bd808603cc0.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: Yeah, thinking it through, I'm not sure that using the time to merge metric is particularly useful compared to just a count of the number of PRs. I'm going to reword the RFC to focus on the number of PRs rather than the time. I'll leave my statistics up there if anyone is interested though.

rfc-126-custom-configuration-for-dependabot.md
thomasleese: I've reworded the RFC to focus on PR numbers in 95a33e35f34a2afdc96c8a73888aa00010d76f16 which I think addresses this!

rfc-126-custom-configuration-for-dependabot.md
thomasleese: I've reworded the RFC to focus on PR numbers in 95a33e35f34a2afdc96c8a73888aa00010d76f16 which I think addresses this!

rfc-126-custom-configuration-for-dependabot.md
```diff
@@ -0,0 +1,255 @@
...
+## Problem
+
+Dependabot was enabled across all our applications in 2018. Due to the number of applications we have and libraries we use, keeping on top of the number of PRs raised by Dependabot has been difficult.
+
+A few solutions have been implemented in the past to make this better:
+
+- [govuk-dependencies] shows the list of open PRs grouped by application, team or gem.
+- [govuk-dependencies] has a Slack bot which automatically messages in each team's channel with a list of open PRs.
+- We [allowed merging Dependabot PRs without requiring two approvals][rfc-103].
+- [Sharing the responsibility for Dependabot PRs across multiple times][dependencies-team].
```
richardTowers, line 16:
```suggestion
- [Sharing the responsibility for Dependabot PRs across multiple teams][dependencies-team].
```

kevindew:
Thanks for raising this Tom, would be super nice to have less of a backlog of dependabot PRs to deal with. 👍 

A few questions:

- Do we have any idea how this would apply to Python projects?
- Do we know how often we'd run the task to keep this file up-to-date and/or if it might get complex for dealing with open PR's to already apply an update to the config?
- Was the idea of `govuk_*` for defining gems a goer? that seemed pretty nice to make the file more concise?

Overall this sounds awesome and I'm looking forward to seeing the impact. However, the one part I have reservations about is having the exact same file in every app and using "govuk_saas_config" to maintain the contents of that.

My logic for this is:

- The same file everywhere creates a number of surprising situations:
  - projects can have dependabot configurations for languages the project doesn't support (for example an API only app like Publishing API having javascript dependencies)
  - adds a bunch of dependencies that an app doesn't have, some of which may be legacy (for example sass-rails)
- Any changes to the global config may affect only one or two apps but will create a mass of PRs
- It introduces an anomaly, no other files in a repo are managed by a different system and therefore would need something of a "DO NOT TOUCH" esque comment
- If a team is actively working on an app they may want to be keeping closer to the metal on dependencies (for example the Publishing Workflow team were fully on-top of dependencies) so it'd actually be beneficial to allow more dependabot updates.

To illustrate a little, looking at the template we have already the Javascript section essentially has 3 dependencies (markdown-toolbar-element, miller-columns-element, paste-html-to-govspeak) that are essentially Content Publisher only so it seems a bit confusing for 60+ apps to have this recorded.

I feel like we've taken this direction based on assumption that, without an external system maintaining consistency, we'll struggle to keep this file up-to-date. I do wonder if that'll actually be the case though since the dependencies don't change that much and once we've set the initial files I don't think we'll gain many benefits from the auto maintenance of the file.

I wonder if we do a one-shot job, that opened bulk PR's to add a base .dependabot/config file, whether that'll provide the same benefits of a govuk-sass-config approach? I feel that would save the confusion/inflexibility downsides and also potentially be simpler to apply so govuk_saas_config doesn't need to get bogged down in creating PRs?

Thanks for your work on this 👍  on a completely different tangent, have we considered whether switching from dependabot to [depfu](https://depfu.com/) offers any nicer benefits such as batching?

thomasleese:
> * Do we have any idea how this would apply to Python projects?

Hmm, good question, I hadn't thought about that to be honest! This might relate to your comments below, but it feels like we wouldn't want the Python projects to have any references to the Ruby or JavaScript packages we use in the majority of our apps... if we decide to go with using govuk-saas-config, we could have a special case. If we decide not to use govuk-saas-config, then we can write a bespoke config file, in which case dealing with Python projects would be the same as any other.

> * Do we know how often we'd run the task to keep this file up-to-date and/or if it might get complex for dealing with open PR's to already apply an update to the config?

In my mind, the task would run daily, we could add the Rake task to the existing daily Jenkins jobs which configures the GitHub repos: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/jobs/configure_github_repos.yaml.erb

In terms of dealing with open PR's, I don't imagine it to be too complex, I think the logic would be that if there is a PR already open, the task can push commits to the existing PRs.

> * Was the idea of `govuk_*` for defining gems a goer? that seemed pretty nice to make the file more concise?

Yep, it's currently in the example configuration in the RFC: https://github.com/alphagov/govuk-rfcs/pull/126/files#diff-b95f76298374fa0276aaf3c985a7ce4eR167

> Overall this sounds awesome and I'm looking forward to seeing the impact. However, the one part I have reservations about is having the exact same file in every app and using "govuk_saas_config" to maintain the contents of that.
> ...

I don't feel particularly strongly either way on this. Having written the [upgrade-ruby-version](https://github.com/thomasleese/upgrade-ruby-version) script, I'm pretty familiar with using the GitHub APIs to open PRs, which is why I'm was happy proposing a solution that used them to keep everything up to date.

I can see the benefits of having something automatically maintain that file, but you're right, I can't see the file changing that often. I've also just thought of a potential problem with the system is that if we _remove_ a dependency from the global config, because we allow individual apps to _add_ further dependencies to the global config, the task wouldn't detect any changes and wouldn't raise any PRs. It feels like a global config might conflict with the idea of letting apps further customise it.

I'd be happy to pivot towards a manual maintenance of these files in this RFC. We can always raise a new RFC which builds on top of this one to implement automatic updates if we discover it becomes a problem.

I wonder what other people on the RFC think? @benthorner @bilbof @richardTowers 

> Thanks for your work on this 👍 on a completely different tangent, have we considered whether switching from dependabot to [depfu](https://depfu.com/) offers any nicer benefits such as batching?

I did briefly look into alternative software. I think we might run into problems that the Cyber team have requested that Dependabot security updates are enabled across all repos, so we might end up having to run two pieces of software. It also looks like Depfu doesn't currently support Python, so we'd definitely need to run both Depfu and Dependabot.

The batching features in Depfu do look great, and pretty similar to what we're wanting to achieve though. There is a Dependabot issue around adding batching support: https://github.com/dependabot/feedback/issues/185

kevindew:
> > * Was the idea of `govuk_*` for defining gems a goer? that seemed pretty nice to make the file more concise?
> 
> Yep, it's currently in the example configuration in the RFC: https://github.com/alphagov/govuk-rfcs/pull/126/files#diff-b95f76298374fa0276aaf3c985a7ce4eR167

Oh great - sorry I missed that (and presumably a number of other iterations). The rendered version link is to a blob of the file, if you switch it to the [branch link](https://github.com/alphagov/govuk-rfcs/blob/dependency-management/rfc-126-custom-configuration-for-dependabot.md) it'll track the changes.

> > Overall this sounds awesome and I'm looking forward to seeing the impact. However, the one part I have reservations about is having the exact same file in every app and using "govuk_saas_config" to maintain the contents of that.
> > ...
> 
> I don't feel particularly strongly either way on this. Having written the [upgrade-ruby-version](https://github.com/thomasleese/upgrade-ruby-version) script, I'm pretty familiar with using the GitHub APIs to open PRs, which is why I'm was happy proposing a solution that used them to keep everything up to date.

Oh neat, that's a cool project. Despite me not understanding a word of that strange python code 😉 

> I can see the benefits of having something automatically maintain that file, but you're right, I can't see the file changing that often. I've also just thought of a potential problem with the system is that if we _remove_ a dependency from the global config, because we allow individual apps to _add_ further dependencies to the global config, the task wouldn't detect any changes and wouldn't raise any PRs. It feels like a global config might conflict with the idea of letting apps further customise it.

Oh sure, was that part of the plan? I thought we'd not let any apps customise the file?

> > Thanks for your work on this 👍 on a completely different tangent, have we considered whether switching from dependabot to [depfu](https://depfu.com/) offers any nicer benefits such as batching?
> 
> I did briefly look into alternative software. I think we might run into problems that the Cyber team have requested that Dependabot security updates are enabled across all repos, so we might end up having to run two pieces of software. It also looks like Depfu doesn't currently support Python, so we'd definitely need to run both Depfu and Dependabot.
> 
> The batching features in Depfu do look great, and pretty similar to what we're wanting to achieve though. There is a Dependabot issue around adding batching support: [dependabot/feedback#185](https://github.com/dependabot/feedback/issues/185)

Sure, sounds reasonable with Cyber having a mandate. Let's hope dependabot catches up. Some of depfu's approaches seem super helpful.

thomasleese:
> Oh great - sorry I missed that (and presumably a number of other iterations). The rendered version link is to a blob of the file, if you switch it to the [branch link](https://github.com/alphagov/govuk-rfcs/blob/dependency-management/rfc-126-custom-configuration-for-dependabot.md) it'll track the changes.

🤦 Thanks!

> Oh sure, was that part of the plan? I thought we'd not let any apps customise the file?

Yeah, so basically the plan was the global config was supposed to be the _minimum_ that should be in each apps `.dependabot/config.yml` file, but the apps can customise and add more rules if they see fit. For example, Search API might want to include Sinatry as a framework dependency: https://github.com/alphagov/govuk-rfcs/pull/126/files#diff-b95f76298374fa0276aaf3c985a7ce4eR195

It does feel like we might have the worst of both worlds though in this case... we still have to manually curate these files if we delete a dependency from the global config or for an individual apps requirements, plus we also have to write and maintain the code to automatically update these files at the same time.

Thinking about it, I'm coming round to preferring a completely manual approach to these files.

benthorner:
> I wonder what other people on the RFC think? @benthorner @bilbof @richardTowers

I think the main risk here is that our repos drift into inconsistency. What we're really doing in this RFC is agreeing an update policy for all of our apps. And I agree that the gems the policy covers will differ with each repo. 

I'm wondering how we can be confident that our repos will keep updating in the way we expect. Looking at our RuboCop config, we have lots of examples where someone has decided to deviate for dubious reasons, or no reason.

The only alternative I can think of is some kind of periodic audit. On the other hand, we may not be overly worried about repos drifting and becoming out-of-date. If we do change the approach, can we write something about the risk?

thomasleese:
Thanks for your input! I think on balance I'm edging towards a manual process, at least for now.

> I think the main risk here is that our repos drift into inconsistency. What we're really doing in this RFC is agreeing an update policy for all of our apps. And I agree that the gems the policy covers will differ with each repo.
> 
> I'm wondering how we can be confident that our repos will keep updating in the way we expect. Looking at our RuboCop config, we have lots of examples where someone has decided to deviate for dubious reasons, or no reason.

I'm imagining that our Dependabot config is unlikely to change as frequently as our linting rules, which I think works in the favour of not automating the configuration. One of the things I like about the manual method, is that we can decided to _increase_ the number of PRs slowly per repo, if we end up on top of things. I think @kevindew chose an example of Content Publisher, where in general, the PRs are dealt with in good time.

> The only alternative I can think of is some kind of periodic audit. On the other hand, we may not be overly worried about repos drifting and becoming out-of-date. If we do change the approach, can we write something about the risk?

I'm going to change the direction of this RFC towards a manual method, pointing out the risk, and explaining where we could go in the future if we decided it was too much work and we wanted a more automated process.

I'll propose a 6 month audit, and from that we can make a decision on whether to keep the current manual process, or switch to a more automated process.

thomasleese:
@kevindew @benthorner @richardTowers @bilbof I've slightly changed the RFC in e1e500c1c9e1605890240f6c40b78d7353730587 to focus on a manual configuration at the beginning, while also leaving the door open to automation in the future based on what we learn in the first six months. Would you be able to take another look?

thomasleese:
Thanks everyone for their comments!

PR #129, opened 2020-11-17
RFC-119 - additional information

WorkSutton:
Some additional notes following initial discussions with @theseanything


WorkSutton:
Following discussion with @theseanything this will be closed as its implementation details which are not appropriate for the RFC

PR #131, opened 2020-11-20
RFC-119: Add some product background in Appendix A

WorkSutton:
I have added some product background. This may not be approprite as these are technical documents.
I did feel that it adds some value following on from various discussions with interested parties since the reopening of the original RFC from March 2020.


WorkSutton:
Closing for now. To be re-opened and reviewed when we are able to implement and/or pick up RFC-119

PR #132, opened 2020-11-20
RFC-119: definitions of existing topic systems

WorkSutton:
Add some definitions to possibly add clarity to the existing topic systems in existance


WorkSutton:
Closing for now. To be re-opened and reviewed when we are able to implement and/or pick up RFC-119

PR #134, opened 2021-01-29
A GOV.UK-wide session cookie & login

barrucadu:
[Rendered](https://github.com/alphagov/govuk-rfcs/blob/msw/account-session/rfc-134-govuk-wide-session-cookie-and-login.md)

**Deadline:** 2021-02-09.

---

We propose introducing a new cookie, `__Host-govuk_account_session`,
which will be an essential cookie, but only set when a user signs in
to use personalised parts of GOV.UK (currently just the Transition
Checker).

Similarly to how our A/B tests work we will manage this cookie at the
Fastly layer, in Varnish Configuration Language (VCL), and use custom
request headers to pass the cookie value to our apps. We will also use
custom response headers to set a new cookie value.

We will create a new app to provide the internal account-supporting
API, and extend frontend to handle the login and logout process.

### DIFF
--- a/rfc-134-govuk-wide-session-cookie-and-login.md
+++ b/rfc-134-govuk-wide-session-cookie-and-login.md
@@ -1,0 +1,360 @@
+# A GOV.UK-wide session cookie & login
+
+## Summary
+
+We propose introducing a new cookie, `__Host-govuk_account_session`,
+which will be an essential cookie, but only set when a user signs in
+to use personalised parts of GOV.UK (currently just the Transition
+Checker).
+
+Similarly to how our A/B tests work we will manage this cookie at the
+Fastly layer, in Varnish Configuration Language (VCL), and use custom
+request headers to pass the cookie value to our apps. We will also use
+custom response headers to set a new cookie value.
+
+We will create a new app to provide the internal account-supporting
+API, and extend frontend to handle the login and logout process.
+
+## Problem
+
+The GOV.UK Account team have launched an experiment on the Transition
+Checker, allowing users to sign up to save their
+results. [finder-frontend sets a session cookie][ff-session]—an
+encrypted cookie—containing the user ID and OAuth tokens used to
+update the data we hold on them.
+
+This works fine for one app, but there are problems with this approach
+when we try to scale to more than one app.
+
+[ff-session]: https://github.com/alphagov/finder-frontend/blob/5b6101e8eb027028a102ab96bcef6ff752b9c558/app/controllers/sessions_controller.rb#L23-L28
+
+### We need to pass cookies to our apps
+
+Our Nginx configuration blocks most cookies, [which we disabled for
+the Transition Checker][tc-cookie-pr]. If multiple apps need to
+consume the session cookie, then this blocking will be disabled for
+ever-increasing chunks of GOV.UK.
+
+[tc-cookie-pr]: https://github.com/alphagov/govuk-puppet/pull/10788
+
+### Apps need to share the same session cookie
+
+It's no good if the user has to log into each part of GOV.UK
+individually.  For example, say we personalise taxon pages: a user
+shouldn't have to log into finder-frontend (for the Transition
+Checker) and collections (for the taxon pages) separately.
+
+There needs to be one session shared across them all.  If we use Rails
+session cookies for this, we need to make sure all apps use the same
+encryption key.
+
+### Which app handles logging in and out?
+
+If there is one session cookie used for all of GOV.UK, which app sets
+that?
+
+Somewhere there needs to be a login controller and a logout controller
+which manipulate the cookie. These controllers will redirect the user
+to the GOV.UK Account system to do the actual authentication, but we
+need something on www.gov.uk itself to set the cookie.
+
+### Non-personalised parts of GOV.UK won't manipulate the session cookie
+
+It's unlikely that all of GOV.UK will be personalised, so there will
+be islands of personalised content.  Currently there is the Transition
+Checker.  Maybe next will be some guidance pages, or something else.
+We want to keep the user's session alive while they are browsing the
+non-personalised parts of GOV.UK, otherwise we risk this bad
+experience:
+
+1. The user logs in to use some personalised part of GOV.UK.
+
+2. The user then spends 30 minutes (or whatever we use for the session
+   duration) viewing non-personalised parts of GOV.UK, but without
+   leaving the site.
+
+3. The user then tries to use another personalised part of GOV.UK, but
+   their session has expired, because the non-personalised parts
+   weren't bumping the expiration time on every page view.
+
+### We can't cache as effectively
+
+The Fastly docs have some [comments on the risks of cookies][fastly-cookies-risks]:
+
+> Cookies can lead to undesirable outcomes.  At worst, if a cookie
+> header is forwarded to your backend server, the backend uses the
+> cookie value to generate personalized content, and that content is
+> then cached by Fastly, a user may end up receiving content intended
+> for someone else.  A theoretical solution to this, adding a Vary:
+> Cookie header to the response, leads to another bad outcome: the
+> response is most likely not cacheable at all, and Fastly will
+> forward all requests to your backend.
+
+[fastly-cookies-risks]: https://developer.fastly.com/reference/http-headers/Cookie/
+
+## Proposal
+
+### Set a session cookie on www.gov.uk
+
+We'll set a new cookie, `__Host-govuk_account_session`, on www.gov.uk.
+
+This cookie will hold:
+
+- The OAuth access token
+- The OAuth refresh token
+
+The access and refresh tokens are opaque and difficult-to-guess
+strings.
+
+This cookie is a secure, [domain-locked][], session cookie:
+
+```
+Set-Cookie: __Host-govuk_account_session=<value>; secure; httponly; samesite=strict; path=/
+```
+
+It can be expired, logging a user out, by re-setting it with
+`max-age=0`.
+
+This cookie cannot be set on the `gov.uk` domain, so service domains
+will need to re-authenticate and manage their own sessions.  This RFC
+is just about authentication and attribute use on www.gov.uk itself.
+The cross-government single-sign-on part of this work is out of scope.
+
+[domain-locked]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Cookie_prefixes
+
+#### Session duration
+
+The cookie is a session cookie.  We will not implement any server-side
+expiration mechanism, though if needed we can revoke or expire the
+refresh tokens.
+
+#### ID tokens
+
+We will not store an OpenID Connect [ID token][] in the cookie.
+Instead the internal API app will make a UserInfo request to look up
+subject identifiers using the OAuth access token.
+
+[ID token]: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
+
+#### Security considerations
+
+Access tokens have a short lifespan, and will only have access to:
+
+- Read from the OpenID Connect UserInfo endpoint, which returns data
+  about the current user.
+
+- Write to an internal API app (not exposed to the internet) which,
+  like our other API apps, will be authenticated with Signon bearer
+  tokens.
+
+Refresh tokens have a long lifespan, but using one requires the OAuth
+client credentials, which are only made available to the internal API
+app.
+
+### Use custom HTTP headers to pass the cookie around
+
+[Fastly's best practices for cookies][fastly-cookies-practices]
+suggest parsing cookies into custom headers and using these headers
+for caching purposes, rather than caching based on the entire
+`Set-Cookie` header (which in our case also contains
+non-account-related things like A/B test bucket assignment, Google
+Analytics session ID, and cookie consent preferences).
+
+In addition, we have nginx configuration to strip out `Set-Cookie`
+headers.  By using custom headers we keep that logic in place,
+ensuring we can't make a mistake and start setting a cookie from one
+of our apps inappropriately.
+
+We will introduce two new headers:
+
+- `GOVUK-Account-Session`: holds the session cookie value.  Set by
+  Fastly to pass the cookie to our apps, set by our apps to create a
+  new cookie.
+
+- `GOVUK-Account-End-Session`: set by our apps to expire the cookie.
+
+We will need to make two changes to our VCL.
+
+[fastly-cookies-practices]: https://developer.fastly.com/reference/http-headers/Cookie/#best-practices
+
+#### Changes to `vcl_recv`
+
+When receiving a request, extract the cookie value and pass it in the
+header:
+
+```vcl
+set req.http.GOVUK-Account-Session = req.http.Cookie:__Host-govuk_account_session;
+```
+
+#### Changes to `vcl_deliver`
+
+If the response depends on the user session, it must either:
+
+1. Set a `Vary: GOVUK-Account-Session` header, or
+2. Set headers to prevent caching entirely
+
+When delivering a response to the user, set a new cookie with a new
+expiration time, and disable shared caches outside of Fastly (both
+Fastly and the user's browser can still cache the page) if the
+response depended on the session:
+
+```vcl
+if (resp.http.GOVUK-Account-End-Session) {
+  add resp.http.Set-Cookie = "__Host-govuk_account_session=; secure; httponly; samesite=strict; path=/; max-age=0"
+  unset resp.http.GOVUK-Account-End-Session;
+} else if (resp.http.GOVUK-Account-Session) {
+  add resp.http.Set-Cookie = "__Host-govuk_account_session=" + resp.http.GOVUK-Account-Session + "; secure; httponly; samesite=strict; path=/"
+}
+
+if (resp.http.Vary ~ "GOVUK-Account-Session") {
+  unset resp.http.Vary:GOVUK-Account-Session;
+  set resp.http.Cache-Control:private = "";
+}
+
+unset resp.http.GOVUK-Account-Session;
+```
+
+### Extend frontend to manage the auth process
+
+It's weird to have the login and logout controllers for GOV.UK as a
+whole located under `/transition-check`.
+
+We will instead add the following endpoints to frontend:
+
+- `GET /sign-in`: initiates the OAuth flow with the accounts system
+  and sends the user on a consent journey.  Accepts these parameters:
+
+    - `_ga`: cross-domain tracking parameter to pass to the accounts
+      domain (optional)
+    - `redirect_path`: path on GOV.UK to redirect back to after
+      authenticating (optional, default: `/`)
+    - `state_id`: see below (optional)
+
+  This calls `GET /api/oauth2/sign-in` to get the URL to redirect the
+  user to.
+
+- `GET /sign-in/callback`: where the accounts system sends the user
+  back to.  Sets the `GOVUK-Account-Session` response header if the
+  user successfully signed in.  Redirects the user back to the
+  `redirect_path`.
+
+  This calls `POST /api/oauth2/callback` to create the session.
+
+- `GET /sign-out`: sets the `GOVUK-Account-End-Session` response
+  header.  Accepts these parameters:
+
+    - `redirect_path`: path on GOV.UK to redirect back to after
+      signing out (optional, default: `/`)
+
+These endpoints are just part of redirection flows, they have no
+visible response.
+
+### Create a new app to provide internal APIs
+
+We will create a new app—called account-api, which will live on a new
+machine class called personalisation—to manage sessions and to handle
+OAuth interactions with the GOV.UK Account system.
+
+The app will serve these internal endpoints:
+
+- `GET /api/attributes`: looks up and returns some attributes from the
+  user's account.  Accepts these parameters:
+
+    - `session`: the `GOVUK-Account-Session` header value
+    - `attributes`: list of attribute names
+
+  Returns either a 401 (if the access and refresh token have expired /
+  been revoked) or a hash of attribute names and values and a
+  `GOVUK-Account-Session` with a fresh access token, if the old one
+  expired.
+
+- `PATCH /api/attributes`: sets some attribute values on the user's
+  account.  Accepts these parameters:
+
+    - `session`: the `GOVUK-Account-Session` header value
+    - `attributes`: hash of attribute names and values
+
+  This is a partial update.  Attributes *not* named in the hash keep
+  their previous values.
+
+  Returns either a 401 (if the access and refresh token have expired /
+  been revoked) or a new `GOVUK-Account-Session` with a fresh access
+  token, if the old one expired.
+
+- `GET /api/oauth2/sign-in`: returns a URL to redirect the user to, to
+  initiate the OAuth login/consent flow.  Accepts these parameters:
+
+    - `redirect_path`: (optional, default: `/`)
+    - `state_id`: (optional)
+
+- `POST /api/oauth2/callback`: returns a session value, if the user
+  has successfully authenticated.  Accepts the parameters from the
+  OAuth response, which will depend on the flow we use.  For example,
+  if we use the `code` flow, the parameters will be `code` and
+  `state`.
+
+- `POST /api/oauth2/state`: sets some attribute values that will be
+  persisted if the user creates a new account, regardless of whether
+  the user returns to GOV.UK.  Accepts these parameters:
+
+    - `attributes`: hash of attribute names and values
+
+    Returns an ID which can be passed to `/sign-in`.  The record
+    expires after 1 hour.
+
+### How the Transition Checker will work
+
+Here are a few examples of how the Transition Checker will work with
+endpoints moved to frontend and to account-api:
+
+#### Clicking the "Sign in" link in the header
+
+1. The link sends the user to `/sign-in?redirect_path=...&_ga=...`
+
+2. The user is redirected to the GOV.UK Account service domain
+
+3. The user authenticates (registers or signs in)
+
+4. The user is redirected to `www.gov.uk/sign-in/callback?...`
+
+5. The internal API app validates the OAuth response and returns the
+   value for the session header.
+
+6. The `GOVUK-Account-Session` response header is set
+
+7. The user is redirected to `redirect_path`
+
+#### Clicking the "Create a GOV.UK account" button on the results page
+
+1. The button sends the user to a new controller in finder-frontend,
+   which:
+
+   - Calls `/api/oauth2/state` with the user's answers, generating an ID
+   - Redirects the user to `/sign-in?state_id=...&redirect_path=...&_ga=...`
+
+2. The new app passes the state attributes to the accounts system
+
+3. The user is redirected to the accounts system, auths, and is sent
+   back (as in steps 2, 3, and 4 above)—but if the user registers, the
+   accounts system saves the provided attribute values
+
+4. The new app creates the session and sends the user to the
+   `redirect_path` (as in steps 5, 6, and 7 above)
+
+This set-some-attributes-on-register flow is necessary because we
+display a confirmation page after the user signs up.  This page
+welcomes the user, says we've sent them a confirmation email, and
+gives a link back to the service.  We want to persist the attributes
+even if the user does not click that link.
+
+We plan to remove the `state_id` mechanism when we retire the
+Transition Checker experiment.
+
+### Local development
+
+A Fastly-managed cookie won't work when running GOV.UK apps locally.
+
+To support that use-case, if `Rails.env.development?`, the new app
+will set an unencrypted `govuk_account_session` cookie, on the domain
+`dev.gov.uk`, which contains the session information, in addition to
+sending the response headers.


barrucadu:
> If it's a session ID that needs to be looked up in a DB on every request, I wonder "why bother with putting the expiry in the cookie and signing it, when you could just store the expiry in the DB (since you're going to have to look it up anyway)?". So I guess it's not an ID?

I was actually thinking of it being an ID.  Storing the expiration time in the database has the same problem as this:

> 1. The user logs in to use some personalised part of GOV.UK.
>
> 2. The user then spends 30 minutes viewing non-personalised parts of
>    GOV.UK, but without leaving the site.
> 
> 3. The user then tries to use another personalised part of GOV.UK, but
>    their session has expired, because the non-personalised parts
>    weren't bumping the expiration time on every page view.

In this case the parts of GOV.UK visited during (2) would be parts which don't hit the database, and so the expiration time doesn't get updated.  If only a handful of pages on GOV.UK (to start with) use personalisation, then it's pretty likely for this to happen.

We could work around that by setting a long expiration time in the database, though then if an attacker manages to get a user's cookie they'll be able to use it for longer.

We probably *will* want some database-level expiration, to avoid an ever-growing collection of old IDs and OAuth tokens, but I think we still need the tamper-proof cookie expiration, so that these cookies expire in a reasonable time.

---

If in the future we have some sort of personalisation on a majority of pages (eg, showing a sign in / sign out header on every page), then we'll be able to do the expiration purely based on database times.  But we're not there yet.

richardTowers:
Ah yeah, that makes sense. If we want to handle the situation where a user noodles around on entirely cached bits of GOV.UK (even if they're personalised), and we don't want their session to time out due to an idle period, it has to be fastly that updates the expiry.

I guess we could use Fastly's [private edge dictionary](https://docs.fastly.com/en/guides/private-dictionaries) to store the expiry in a database that can be written at the edge. But then that's even more complex.

Makes sense to me! Thanks for writing this up. I'll have a proper look next week :)

VitalieMogoreanu:
CDN approach is probably the only way I can think of to do this

I am very happy we are passing _ga parameter :-). there are 2 _ga cookies currently -we should probably fix this so that there is less confusion about which one to pass (I don't think we need to pass both)

A few questions:
- Why 30 minutes for session expiration? I feel we could have longer sessions on GOV.UK - needs to be reviewed with product people and cyber
- POST /api/attributes to update the attributes should support partial update - not all the attributes. 
- POST /api/attributes do we need to worry about conflicting updates from multiple tabs/devices sharing the same session?
- It would be good to have an interface to analyse all user data (not per user), that may be out of scope of this RFC though

I'll have another look next week

barrucadu:
> Why 30 minutes for session expiration?

It's what we currently have for the account manager app.  I'll add a note to the text that 30m is just an example and we'll figure out an appropriate session duration.

> POST /api/attributes to update the attributes should support partial update - not all the attributes.

Yes, this'll be a partial update.  I'll clarify that.

> POST /api/attributes do we need to worry about conflicting updates from multiple tabs/devices sharing the same session?

I think we can probably leave this until we have more attributes.  Erin is thinking about how a GOV.UK attribute-store-for-personalisation will work so I don't want to delve into that too deeply with this RFC.

> It would be good to have an interface to analyse all user data (not per user), that may be out of scope of this RFC though

Yes, I think that'll tie into Erin's work.

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+that?
+
+Somewhere there needs to be a login controller and a logout controller
+which manipulate the cookie. These controllers will redirect the user
+to the GOV.UK Account system to do the actual authentication, but we
+need something on www.gov.uk itself to set the cookie.
+
+### Non-personalised parts of GOV.UK need to manipulate the session cookie too
+
+It's unlikely that personalisation will roll out to all of GOV.UK in
```
erino, line 62: I would say that the "in one go" is superfluous, as it's unlikely that all of GOV.UK will every be personalised. For example AAIB reports

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+one go.  There will be islands of personalised content.  Currently
+there is the Transition Checker.  Maybe next will be some guidance
+pages, or something else.  We want to keep the user's session alive
+while they are browsing the non-personalised parts of GOV.UK,
```
erino, line 66: Should we think about services here? as users don't realise that service domains are different. Although I can't think of a simple fix.

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+
+    Returns an ID which can be passed to `/sign-in`.
+
+Here are a few examples of how the Transition Checker will work:
+
+#### Clicking the "Sign in" link in the header
+
+1. The link sends the user to `/sign-in?redirect_path=...&_ga=...`
+
+2. The user is redirected to `www.account.publishing....`
```
erino, line 291: or to the GOV.UK Account service domain `www.account.service.gov.uk`

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+
+[fastly-cookies-practices]: https://developer.fastly.com/reference/http-headers/Cookie/#best-practices
+
+#### Preventing tampering with the cookie
+
+We can't rely on client-side cookie timeouts to log users out.  If an
+attacker gets a copy of a user's session cookie, they can include it
+in any requests they send, regardless of whether it has expired.  So
+we need to sign the cookie value and expiration time to prevent
+tampering.
```
kevindew, line 120: Will apps do any validation of the session header(s) at all? Or will they just receive a Session-ID header and trust that this is valid, similar to bearer token authentication?

I guess if we have these concerns about an attacker getting hold of a user's session cookie then we have to consider can they do anything with the Session-ID alone (which can be determined from the cookie)? I guess at a basic level something needs to make sure Fastly VCL always has a `unset req.http.GOVUK-Session-ID;` which isn't behind conditionals so an attacker can't just set the header manually. Then beyond that we have edge cases like the `www-origin` entry (which I believe is just for VPN access) where we'd not have this cookie verification.

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+`GOVUK-Session-ID` value, if one is returned; and end the session, if
+the user has been logged out:
+
+```vcl
+if (beresp.http.GOVUK-Session-ID) {
+  set req.http.GOVUK-Session-ID = beresp.http.GOVUK-Session-ID;
+}
+
+if (beresp.http.GOVUK-End-Session) {
+  set req.http.GOVUK-End-Session = "true";
```
kevindew, line 187: Would apps be listening for this header to delete a session or would this be purely for VCL?

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+2. The user is redirected to `www.account.publishing....`
+
+3. The user authenticates (registers or signs in)
+
+4. The user is redirected to `www.gov.uk/sign-in/callback?...`
+
+5. A session ID is generated, and used to persist the OAuth access and
+   refresh tokens to the app's database
+
+6. The `GOVUK-Session-ID` response header is set
```
kevindew, line 300: A few questions come to mind about the intentions of an app once they have one of these?

I assume an app can't assume that the presence of this header means that a user is signed in? is that something that needs to be confirmed with the account frontend? Would this be confirmed on each request or just once in a period of time? 

Would apps remain having their own sessions (eg _finder_frontend_session) (this seems essential for things like flash messages) and would there be any intersection with this govuk_session and those?

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+- `GET /sign-out`: sets the `GOVUK-End-Session` response header,
+  deletes the OAuth tokens from the app's database, and redirects the
+  user to `/`.
+
+The app will also serve these internal endpoints:
+
+- `GET /api/attributes`: looks up and returns some attributes from the
+  user's account.  Accepts these parameters:
+
+    - `session_id`: the value of the `GOVUK-Session-ID` header
```
kevindew, line 263: I'm assuming a user is a record that references their GOV.UK account user id - so one user per account. And that you use the session-id to determine the account. Can a user have multiple valid session-ids at a time?

kevindew (commented):
This sounds great. A really well considered RFC and sounds like a good improvement on the transition checker flow.

I ended up with more questions than I was expecting. I hope they are helpful.

maxgds:
Apologies if I overlooked it, but I'm not clear on how this fits into the overall GOV.UK cookies strategy as described here: https://www.gov.uk/help/cookies - you say this would be an optional cookie and it looks to me like it would best fit under the "remember your settings" header, in which case what would the outcome be if the user has not consented to that kind of cookie?  Would part of the sign up process maybe insist on them accepting such cookies? In which case what happens to people who signed up already? And does accepting this cookie mean they accept all others under "remember your settings"? Or would you create a new "accounts" category for this particular cookie/set of cookies?

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+
+We will create a new app—called account-frontend, which will live on a
+new machine class called personalisation—to serve those controllers
+and to handle OAuth interactions with the GOV.UK Account
+system. Signing in will set the `GOVUK-Session-ID` response header,
+signing out will set the `GOVUK-End-Session` response header.
+
+This app will serve these public endpoints:
+
+- `GET /sign-in`: initiates the OAuth flow with the accounts system
```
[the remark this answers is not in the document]
barrucadu: They're just part of the redirection flow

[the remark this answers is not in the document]
barrucadu: I've added a sentence clarifying

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+  back to.  Sets the `GOVUK-Session-ID` response header and stores the
+  returned OAuth access and refresh tokens in the app's database, if
+  the user successfully signed in, and redirects the user back to the
+  redirect_path.
+
+- `GET /sign-out`: sets the `GOVUK-End-Session` response header,
+  deletes the OAuth tokens from the app's database, and redirects the
+  user to `/`.
+
+The app will also serve these internal endpoints:
```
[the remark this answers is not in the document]
barrucadu: I think we can use a signon API user to authenticate use of this API, as this app will use (user-specific) OAuth tokens when talking to the main attribute store

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+#### Changes to `vcl_deliver`
+
+When delivering a response to the user, set a new cookie with a new
+expiration time, and disable shared caches outside of Fastly (both
+Fastly and the user's browser can still cache the page) if the
+response depended on the session:
+
+```vcl
+if (req.http.GOVUK-End-Session == "true") {
+    set req.http.GOVUK-Session = "";
```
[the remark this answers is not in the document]
barrucadu: Good point, I'll make it set the cookie to have a max-age of 0 in the logout branch

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+} else if (req.http.GOVUK-Session-ID) {
+  set req.http.GOVUK-Session-Expires = strftime({"%s"}, time.add(now, 1800s));
+  set req.http.GOVUK-Session-Signature = digest.hmac_sha256(table.lookup(env, "session_key"), req.http.GOVUK-Session-ID + "." + req.http.GOVUK-Session-Expires));
+  set req.http.GOVUK-Session = digest.base64url_nopad(req.http.GOVUK-Session-ID) + "." + digest.base64url_nopad(req.http.GOVUK-Session-Expires) + "." + digest.base64url_nopad(req.http.GOVUK-Session-Signature);
+}
+
+if (req.http.GOVUK-Session) {
+  add resp.http.Set-Cookie = "govuk_session=" + req.http.GOVUK-Session + "; secure; httponly; samesite=strict; max-age=1800; path=/";
```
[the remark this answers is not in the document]
barrucadu: Yeah, the cookie has to be for www.gov.uk.  Browsers won't let us set a cookie on `gov.uk` itself, as it's (kind of) a TLD, so we can't share cookies with service domains.

I'll add a comment about that.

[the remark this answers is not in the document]
barrucadu: The attributes are likely to fall into two types: those we use on www.gov.uk for personalisation, and those which are shared cross-government for services to use (proof of identity, passport details, etc); and they'll be stored in different places.  Those latter ones will be accessible.

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: This is purely for VCL

barrucadu:
> you say this would be an optional cookie

I'm not totally clear myself on how to classify this cookie.  It's optional in that you can use www.gov.uk without it (by not using any of the accounts stuff).  But it's required if you *do* want to use the accounts stuff.  I don't think bundling use for accounts in with consent for other cookies is a good idea, as that could be seen as nudging people.

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: I think things off www.gov.uk are out of the scope of this RFC, as they fall under the cross-gov stuff (which is looking likely to sit with Digital Identity?)

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+if (resp.http.Vary ~ "GOVUK-Session-ID") {
+  unset resp.http.Vary:GOVUK-Session-ID;
+  set resp.http.Cache-Control:private = "";
+}
+```
+
+The `1800` here (30 minutes) is just an example.  We will work out an
+appropriate session duration with product and security input.
+
+### Create a new app to manage the auth process
```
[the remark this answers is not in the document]
barrucadu: Yeah, I expected this to be a point of contention.

I think putting it in frontend would also work.  My reasoning for making a new app on a new machine class is that this is likely to have very different traffic to other parts of GOV.UK---lots of direct origin hits, no caching at all---so maybe we want something which is easy to scale quickly.  And given that accounts / personalisation is a big item in the spend review & roadmap it's unlikely the team would vanish and the new app fall to Platform Health.

> gem/engine

Yeah, that's another option.  I think the difficulty there is that we'd need to do without the database of session ID -> OAuth tokens.  So we'd need to store those tokens in the cookie, which suggests to me an encrypted (rather than just signed) cookie.  But we've got to do the encryption in VCL (to avoid the origin hit on every request to bump the cookie expiration), and [Fastly's VCL for crypto][1] is very digest-heavy and not so good for encryption.

[1]: https://developer.fastly.com/reference/vcl/functions/cryptographic/

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+
+- `GET /sign-in/callback`: where the accounts system sends the user
+  back to.  Sets the `GOVUK-Session-ID` response header and stores the
+  returned OAuth access and refresh tokens in the app's database, if
+  the user successfully signed in, and redirects the user back to the
+  redirect_path.
+
+- `GET /sign-out`: sets the `GOVUK-End-Session` response header,
+  deletes the OAuth tokens from the app's database, and redirects the
+  user to `/`.
```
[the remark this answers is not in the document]
barrucadu: Sure, that sounds reasonable

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+  This is a partial update.  Attributes *not* named in the hash keep
+  their previous values.
+
+- `POST /api/state`: sets some attribute values that will be persisted
+  if the user creates a new account, regardless of whether the user
+  returns to GOV.UK.  Accepts these parameters:
+
+    - `attributes`: hash of attribute names and values
+
+    Returns an ID which can be passed to `/sign-in`.
```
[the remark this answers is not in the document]
barrucadu: Since this endpoint is for passing extra information to the auth process, we'll want it to expire.  I'll add a note.

maxgds:
> It's optional in that you can use [www.gov.uk](http://www.gov.uk) without it

I think there's a strong argument that it's essential. If you drill down into what our existing cookies are here https://www.gov.uk/help/cookie-details it seems to match the use case of licensing_session - we don't claim that license cookies are optional as you could potentially apply on paper instead.

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
...
+Checker experiment.
+
+### Local development
+
+A Fastly-managed cookie won't work when running GOV.UK apps locally.
+
+To support that use-case, if `Rails.env.development?`, the new app
+will set an unencrypted `govuk_session` cookie, on the domain
+`dev.gov.uk`, which contains the `GOVUK-Session-ID`, in addition to
+sending the response headers.
```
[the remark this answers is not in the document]
barrucadu: Yeah, this sounds like a good candidate for some middleware or a gem.  I think to start with we'll just end up putting some code straight in the app, and then a couple of apps are on board we'll know what's most useful to pull out as common code.

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: I was thinking of having the app trust the session ID header, as if it's only set by Fastly (which verifies the signature) then we're not opening up any vulnerabilities.  Though as you spotted I missed a place where the header needed to be unset.

An alternative would be passing the signed cookie value through in a header, and doing validation in the app which consumes it.  That means we'd need to keep the secret in sync between Fastly (so we can continue to bump the cookie expiration time through VCL) and puppet (so our app can validate the header).  What do you think?

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: If there's a `GOVUK-Session-ID`, then the user is signed in, because they have a valid session cookie.

However, it's not a *user* ID: all you can do with the session ID is get or set attributes, by querying the endpoints in the (new?) app.  It could be that the OAuth token associated with the session ID has been revoked, in which case any attempt to get/set attributes would fail (a 401 status seems most appropriate).  However if an app wants to display different information to signed in vs signed out users, then it can just use the presence of the session ID without checking attributes.

In fact that's how the account navigation in the header on [/transition](https://www.gov.uk/transition) works: it just checks for the presence of a finder-frontend session cookie.  It doesn't try to read the session cookie because /transition is rendered by collections.

Apps would keep their own sessions for flash messages and the like, if needed.

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: If a user is logged in from multiple devices, then they'll have multiple sessions.  I'll add a note saying that the session ID is a random value generated a log in.

barrucadu:
> I think there's a strong argument that it's essential.

Great, and in fact I see that our current account cookies are listed there as essential, even though they're optional in exactly the same way, I should have checked that first :man_facepalming: 

I'll update the RFC.

rfc-134-govuk-wide-session-cookie-and-login.md
stevenjmesser: Out of interest, would Digital Identity cover the brokering aspect to an account? For example, a logged-in user using a service which needs the user's stored details brokered through the account. If those details can be stored without needing to verify the user's identity, does that fall under Digital Identity's remit?

barrucadu: Currently it's looking like Digital Identity will be doing the cross-gov attribute exchange stuff, yes.  But the details are still being worked out.

rfc-134-govuk-wide-session-cookie-and-login.md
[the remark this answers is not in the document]
kevindew: > Yeah, I expected this to be a point of contention. 

Good good :) It always seems easier to add an app than to remove one.

It strikes me that the main architectural advantage of doing this in a new app is having a shared database of attributes that multiple apps can access - is that an approach we're going to be suggest to other departments if/when they implement accounts? As I guess, this might serve as an implementation example

I did have a ponder about the app, [based off there being no user interface](https://github.com/alphagov/govuk-rfcs/pull/134/files/2d18065973d53799743a5e4ebf4f0b10ac1ca5db#diff-f5a6037308ea735f2a6e63b3080e78792499608bcad7386a49cf94799634d1be) and the [expected later need of shared code](https://github.com/alphagov/govuk-rfcs/pull/134#discussion_r567730827), and wondered if pitching it as accounts-frontend might not be ideal as it'd be the only frontend app without an interface. I wondered whether the sign-in and sign-out parts might actually belong distinct from the API aspects of the app - probably in a different frontend app. It struck me as an architectural smell having an app that is partially private API and partially public, the only app of ours I can of with this is Email Alert API and that hasn't worked out too well in keeping the private API actually private.

rfc-134-govuk-wide-session-cookie-and-login.md
kevindew: > An alternative would be passing the signed cookie value through in a header, and doing validation in the app which consumes it. That means we'd need to keep the secret in sync between Fastly (so we can continue to bump the cookie expiration time through VCL) and puppet (so our app can validate the header). What do you think?

Yup that seems a reasonable solution and would resolve the problem. I can see that it'd be a bit of a pain to manage so might be worth validating the necessity first. 

As I understand the worry is that the way an account could be hijacked through discovering a cookie (is that the only public availability of this session id?) and we're accepting that by discovering that cookie they can immediately use the account regardless of verification. The security feature kicks in after that cookie has aged, so if we're in a world where session id lives indefinitely and a cookie lifetime is short (30m) then it offers a security advantage of limiting the window of a bad actor. Whereas if we were in a world where the cookie had much longer life time (1 week) and a session id only lasted for 24 hours then it'd be an unnecessary feature as it wouldn't change any risk.

Perhaps it's worth considering a think about cookie lifetime (since I understand that's an open question), session id lifetime and how secret session id is?

rfc-134-govuk-wide-session-cookie-and-login.md
kevindew: Right - gotcha. Thanks for explaining. So because of the verification there are assumptions we can make about the session id's validity (because of the signature check) and to do anything substantial there needs to be an API call that's actually authenticated. I guess by that logic it's preferential to not cache any attributes/account stuff in the local session as that might have a very different lifetime.

rfc-134-govuk-wide-session-cookie-and-login.md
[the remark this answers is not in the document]
barrucadu: > It strikes me that the main architectural advantage of doing this in a new app is having a shared database of attributes that multiple apps can access - is that an approach we're going to be suggest to other departments if/when they implement accounts?

We think there are likely to be two classes of attributes: those which are useful only within a specific service (or department) and those which are useful across government.  And those ones which aren't useful across government will live with the service.  For example, if GOV.UK Pay use GOV.UK Accounts to implement a way for users to set up and manage recurring payments, we definitely don't want to end up storing payment information in the cross-gov attribute store.

So there is a need for attributes held within a service.  And if that service is composed of microservices (like GOV.UK Publishing is), then the attributes need to be shared.

>  I wondered whether the sign-in and sign-out parts might actually belong distinct from the API aspects of the app - probably in a different frontend app. It struck me as an architectural smell having an app that is partially private API and partially public, the only app of ours I can of with this is Email Alert API and that hasn't worked out too well in keeping the private API actually private.

That's a very good point, it is a bit of a weird "frontend" app.  Maybe the login / logout functionality should live in `frontend` and the attribute management part live in a separate API service which is kept properly internal.

@erino what do you think? This ties in with your work on a GOV.UK Attribute Service.  For this RFC to work we'll need:

- A database mapping session IDs to user IDs (one user can have multiple sessions)
- A database mapping user IDs to OAuth tokens (so we can use the cross-gov attribute store)

Maybe the GOV.UK Attribute Service can do those things as well as storing the GOV.UK-specific attributes?

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: @OllieJC - do you have any thoughts on this?  To summarise, we're in this situation:

- A user is assigned a random session ID when logging in.
- This session ID is stored in a (secure, httponly) cookie with a short (undecided - 30 minute?) expiration time.
- We sign the cookie value + expiration time so that an attacker can't tamper with it.
- On every request, we bump up the expiration time.

The problem is that the session ID is kind of sensitive.

We *don't* regularly change the session ID, as we don't want to hit our origin on every request (GOV.UK can't cope with every request going to origin).  Which means that if an attacker learns the session ID, they can access the account it corresponds to if they can:

- forge a cookie (which would require being able to sign it with out secret), or
- bypass the cookie mechanism and pass the session ID to our origin directly.

---

We could put an expiration time on the session ID, so that a user would need to log in again even if they had been continuously browsing GOV.UK.  So this would be an upper limit on session duration.  But we'd probably want a long upper limit (days) to avoid being annoying.

We could pass the full signed cookie value and not just the session ID to our origin, so an attacker would need to forge a cookie (which requires knowing a secret).  But we'll need to keep the signing secret consistent between our apps and our Fastly config.

rfc-134-govuk-wide-session-cookie-and-login.md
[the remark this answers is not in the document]
erino: > That's a very good point, it is a bit of a weird "frontend" app. Maybe the login / logout functionality should live in frontend and the attribute management part live in a separate API service which is kept properly internal.

I think that makes sense. Having the attribute management service as an internal app will help us keep the API private. Then if an attribute needs to be accessed by a third-party then that should make the case for the attribute to be available in the GOV.UK Account.

> Maybe the GOV.UK Attribute Service can do those things as well as storing the GOV.UK-specific attributes?

Yes, that would make sense. An attribute service for GOV.UK publishing should be responsive enough to support handling sessions.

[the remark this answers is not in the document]
barrucadu: I'll update the RFC to introduce a new internal API app for the endpoints already described, and say that `frontend` will handle the user-facing redirection flows.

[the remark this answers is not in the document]
barrucadu: d80b53187662942a351375f99a86ea2d91b29667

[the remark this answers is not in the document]
erino: > I do wonder whether a broad app like /frontend might still be a potential home since that's the somewhat custom frontend app (@kevindew's first comment)

If we're going to split login and attribute storage, then might an existing frontend app be a good candidate for the `/sign-in*` and `/sign-out` public endpoints?

OllieJC:
I may have misunderstood some of the reasoning here, but I think there's perhaps a lower complexity way of doing this.
I have reservations about the use of an expiry on the client side, use of symmetric keys for hashing and the use of the CDN to set headers, some of which has already been mentioned in the other comments.

My thoughts in a sort of generic user journey flow:
- Have an endpoint `www.gov.uk/sign-in` that is the only place where a cookie is **created** for the `www.gov.uk` domain
- (consider using the `__Host-` prefix to lock the cookie to the domain ([ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Cookie_prefixes)))
- After successful sign in, the `/sign-in` endpoint generates a session ID, where that ID is checked to see if already in use (to prevent randomly generating the same ID as another session) - or use a GUID that has a time seed component, then encrypt the ID with a private key (only available to that app, not shared with any other app), session ID is set in the database against a user account along with an expiry time (I think this can be a reasonable length - such as an hour)
- Cookie is session-based, and doesn't have an expiry client-side (that's handled in the DB), the value in the cookie cannot be tampered with as it'll be encrypted 
- On an origin of another endpoint, fetch the cookie and decrypt with a public key from another endpoint (like `www.gov.uk/.well-known/session-key.pub`) the key can be truly public as it's checking if the paired private key encrypted it, it doesn't matter if someone can decrypt the session to get the session ID as they won't be able to re-encrypt it - this helps get around needing to share symmetric keys for hashing with other origins 
- (there could be multiple public/private key pairs and a reference used in the cookie, the value could be `1:base64` where `1` is the public key reference, `base64` is of the encrypted session ID, this helps with any future rotations)
- The decrypted session ID shouldn't be considered secret or sensitive on its own
- Other origin apps call an internal (authed) API (`/api/attributes`) with the session ID (or just send the entire encrypted cookie) the API checks the expiry from the database and returns attributes for that app, the expiry time is updated in the DB to extend/roll the session
- (should every personalisation origin request query the API to validate the session ID?)
- Any app can set the session cookie to expired (e.g.: `Set-Cookie: __Host-govuk_account_session=; secure; httponly; samesite=strict; max-age=0; path=/`) if the API returns an expired/invalid state and then either act as logged out in the response, or redirect to a signed-out endpoint

barrucadu:
A session cookie encrypted on the backend and not touched by Fastly would make this simpler.  The only thing I'm not sure about is the expiration time,

> session ID is set in the database against a user account along with an expiry time (I think this can be a reasonable length - such as an hour)

As I think we'd want this to work:

- Someone logs into GOV.UK to use one personalised thing
- They then continuously browse non-personalised parts of GOV.UK (so the session ID isn't used, and the expiration time isn't bumped up)
- They then try to use another personalised thing

I'll check with Daisy what our typical session lengths are like.

barrucadu:
So, we have an average session duration of 2:29, *but* the PAs don't really like session duration as a metric because it's skewed by people who just leave the page open.  So we could probably use 1 hour.

barrucadu:
Since we have config in place to strip out `Set-Cookie` headers from our apps, I'll keep the custom headers for starting / ending the session in place.  But all of the crypto stuff can be removed from VCL.

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: I've implemented Ollie's suggestions now.  I've kept the custom headers, because I think our nginx config which strips out `Set-Cookie` *does* add some value (eg, it ensures we can't make a mistake and accidentally leak cookies from an app, adding cookies has to be considered).

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: With the latest change an app can't assume that because there is a header it's valid; but it can use the public key to check that.

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,354 @@
...
+#### Cookie encryption
+
+The cookie value is encrypted with a private key.  This ensures that
+an attacker cannot tamper with the value.  Apps which consume the
+cookie can use the public key to validate it.  This key can be truly
+public, as the session ID itself is not sensitive.
+
+If the cookie value is malformed, for example if an attacker has tried
+re-encrypting it with a different private key, the consuming app
+should log the user out by expiring the cookie.
```
richardTowers, line 129: If the value of the cookie is a securely generated random number with sufficient entropy, I'm not sure we need any of this cryptography. It should already be tamper proof, by virtue of the attacker not being able to guess another random number.

If we do need crypto, I think we want signing here, not encryption. "This ensures that an attacker cannot tamper with the value" is not true for many encryption schemes (for example AES-CBC is [malleable](https://en.wikipedia.org/wiki/Malleability_%28cryptography%29), AES-GCM doesn't have this problem, but it's still the wrong tool for the job).

You could still use an asymmetric signing algorithm (like ECDSA), but I'm really not sure what that buys us if we assume the attacker is somehow capable of guessing any random numbers we generate.

richardTowers: @OllieJC  - just to check I'm not misunderstanding something?

OllieJC: The purpose of crypto is to assert that sign-in created the cookie, rather than another process iterating through session IDs and trying to see if they're a valid session.

Signatures are fine as long as they're always checked - an upshot of encryption is that it enforces the use of a public key to get the session ID so no origin can use the session ID without decrypting (vs. the signature being potentially optional)

OllieJC:
```suggestion
The cookie value is encrypted with a private key. This ensures that
an attacker cannot re-encrypt the value without knowing the private
key - which is never shared. Apps which consume the cookie can 
use the public key to decrypt, essentially validating the source.
This public key can be truly public, as the session ID itself is not
sensitive, but we do need some way of validating it came from the
right place.

If the cookie value is malformed, for example if an attacker has tried
re-encrypting it with a different private key, the consuming app
shouldn't be able to decrypt it with the public key and should log
the user out by expiring the cookie.
```
Or:
```suggestion
The cookie value is the session ID and a signature of that ID signed
by a private key. This ensures that an attacker cannot re-sign the ID
without knowing the private key - which is never shared with apps.
Apps which consume the cookie must check the signature with the
public key, validating the source. This public key can be truly public,
as the session ID itself is not sensitive, but we do need some way of
validating it came from the right place.

If the cookie value is malformed, for example if an attacker has tried
re-signing it with a different private key or not including a signature,
the consuming app must ignore it and should log the user out by
expiring the cookie.
```

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,354 @@
...
+
+[fastly-cookies-practices]: https://developer.fastly.com/reference/http-headers/Cookie/#best-practices
+
+#### Changes to `vcl_recv`
+
+When receiving a request, extract the cookie value and pass it in the
+header:
+
+```vcl
+set req.http.GOVUK-Account-Session = req.http.Cookie:__Host-govuk_account_session;
```
OllieJC, line 178: What happens if the cookie isn't present? We should unset the `GOVUK-Account-Session` header so a malicious actor can't set that in a custom header, only the expected cookie should set it.
```suggestion
unset req.http.GOVUK-Account-Session;
set req.http.GOVUK-Account-Session = req.http.Cookie:__Host-govuk_account_session;
```

rfc-134-govuk-wide-session-cookie-and-login.md
richardTowers: I'm still not sure it's worth the extra complexity.

Let's say we generate our session IDs with `SecureRandom.hex(128/8)` (a random 128 bit number).

If we've got 2^32 active sessions (that's ~4 billion), and we let an attacker make 2^32 guesses per second (a few guesses per nanosecond), it's still going to take them in the order of 2^64 seconds (585 billion years) to guess a valid session ID. Want more security? Use a 256 bit random number.

I suppose being able to validate the cookie before hitting the database to check it's a real session might be nice, but I would have thought a (symmetric) HMAC would be more appropriate for that. Yes, we'd have to share the HMAC keys between the thing doing the signing and the things doing the validation. But even if an HMAC secret got leaked, it would only let an attacker guess random numbers, which isn't going to help them.

Or perhaps you're thinking about a situation where an attacker has somehow managed to get read access to the database where the session IDs are stored, but doesn't have access to the private key that would let them forge new cookies because that's stored somewhere more secure? To be honest, I haven't really considered this from a "some of the infrastructure is compromised and it still needs to be secure" perspective.

OllieJC: Yeah that's totally reasonable. I don't even think you need a 128bits of entropy and it still be implausible to take a wild guess.

My primary thinking around using a signature is that it asserts to a consuming app verifying it that it came from the signing in mechanism. If in future an ID is leaked in anyway, or can be fetched out using an API etc. then it cannot be used by to gain access, as that requires the secret. As for asymmetric, this means the public key can be truly public and not kept on an app. Knowledge of the public key doesn't allow session cookies can be forged.

For example, have an endpoint like:
`/.well-known/session-public-key-158.pub`
If the value of the cookie was something like:
`158:987654321:hash`
Then the consumer app can fetch the public key, cache it, hash the ID and compare the hashes.

I think there are some nuanced attacks here, it'll be best to threat model the proposal before any implementation - I'll put a Jamboard together tomorrow to try and surface some possible attacks together asynchronously.

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: I think that if the cookie isn't present the current code is equivalent to `set req.http.GOVUK-Account-Session = ""`, but we can check that

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,354 @@
...
+### We need to pass cookies to our apps
+
+Our Nginx configuration blocks most cookies, [which we disabled for
+the Transition Checker][tc-cookie-pr]. If multiple apps need to
+consume the session cookie, then this blocking will be disabled for
+ever-increasing chunks of GOV.UK.
+
+[tc-cookie-pr]: https://github.com/alphagov/govuk-puppet/pull/10788
+
+### Apps need to share the same session cookie
```
galund, line 40: This isn't true in the general case, which is that services will have their own cookies, and will redirect users in an OIDC flow to the central authn service if the user isn't logged in, but needs to be. There will need to be some common standards to make that work - including a way for users to log themselves out, and some agreement about cookie lifetimes.

Not that GOV.UK _has_ to have each of is apps implement authentication against the central service separately: where you draw the lines between your services is up to you. But the architecture must allow for operationally independent services operated by different departments in any event, and the more GOV.UK can itself behave like other services will, the more good patterns can be embedded from the start, and the more the OIDC flows will be properly dog-fooded. And the fewer surprises there will be when we work with other government departments to make their services feel properly integrated with the GOV.UK Account.

barrucadu: Right, this RFC is specifically about `www.gov.uk`, not everything under the `gov.uk` domain.

galund: I get that, I still think that treating different services that the GOV.UK team happen to be running more separately might have some merit, and if you're excluding _in general_ then it's worth stating why. Seeing as there are some benefits which I mentioned :-)

barrucadu: Ok, that makes sense.  I think the main problems with apps on the www.gov.uk domain authenticating separately are:

- It's probably confusing to users if they move between different parts of GOV.UK and their logged-in state changes.  We've already found people being confused at the difference between being logged in to the Transition Checker on www.gov.uk and the OIDC thing on www.account.publishing.service.gov.uk, and they're not even on the same domain.
- It adds friction to personalisation, where we'd have to get users to re-authenticate to see personalised content served by different apps.
- It makes it harder to show some global logged-in state, eg a "logout" link in the header, because the user is individually logged into separate parts of GOV.UK.

But for services *not* on www.gov.uk, eg if there's some new big thing and the GOV.UK team spins up a prepare-for-meteor-impact.service.gov.uk, *that* should authenticate separately, because it's not on the same domain.

My thinking is that the microservices which make up GOV.UK (or prepare-for-meteor-impact.service.gov.uk, or whatever) are an implementation detail which shouldn't be exposed to the user, so you'd ideally want all the microservices that make up one service to share their authenticated-ness.

galund: I think it's confusing for users if logged-in state isn't maintained between _all_ services. This will have to be tackled at some point (although not necessarily for this RFC of course!).

The whole point of gov.uk is that users see a single thing and that the question of which department or team provides _any_ service that central government is responsible for is an implementation detail.

erino: > treating different services that the GOV.UK team happen to be running more separately might have some merit
I was thinking that this approach isn't practical as there are 60+ services which make up GOV.UK Publishing, will the GOV.UK Account want to support that many connections for _one_ client. Then scaling that to each of the micro services which make up the 300+ services across Government.

There's also an OpenID Connect spec for this Open ID [Connect Federation](https://openid.net/specs/openid-connect-federation-1_0.html), which I need to read and see how compatible it is with this RFC.

I would like to think that the GOV.UK Account will consider GOV.UK Publishing as a single service and allow it to manage users sessions in a way is appropriate for it's architecture. An implementation detail as @barrucadu says.

barrucadu: Also, if we do end up wanting to split GOV.UK Publishing into separate clients, that can always happen later.

barrucadu:
It would be really great for the discussion to wrap up tomorrow, so that we can begin implementing things.  Even if what we implement isn't perfect and we need to change it later, we'll need something in place to do our personalisation experiments.

I think the major point of discussion remaining is: **what exactly does the session cookie hold?**  Some sort of randomly-generated (collision resistant) session ID yes, but is that plaintext? or signed? or encrypted?

Are all the other major parts generally considered ok?

- We'll introduce a new session cookie for GOV.UK.
- We *won't* be opening our nginx config to let our frontend apps use cookies, instead the value will be passed around with custom headers and some VCL magic.
- We'll introduce a new API app to do the OAuth and attribute-related interactions.
- We'll introduce a new machine class for that app.
- We'll add some new public endpoints in frontend to do the OAuth redirection flow.
- We'll switch over the Transition Checker to use this new auth mechanism and remove the existing session management code from finder-frontend
- We'll make this work in local development by setting the session cookie directly on the `dev.gov.uk` domain.

richardTowers:
Yep, I'm happy with all of those points - happy to approve once we've worked out what we put in the session cookie.

I think the options on the table are:

* Random session ID, plaintext
* Random session ID + HMAC (symmetric)
* Random session ID + asymmetric signature (e.g. using ECDSA or RSA)
* Asymmetrically encrypted random session ID (e.g. using ECIES or RSA+OAEP)

I'm happy with any of those, so long as there's a rough consensus. They're all based on a strong random ID, so they should all be sufficiently secure.

My preference would be for Random session ID + HMAC (symmetric), as I feel that provides the best balance of security strength in depth and simplicity.

If we go for an asymmetric signature or encryption scheme, I'd like us to be specific about which algorithm we're planning on using, which libraries we're going to use to do the crypto, and what effect that will have on the size of the cookie. The devil's in the detail here, and I'd hate for us to make a decision that ends up costing us a lot of implementation time for little security benefit.

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,337 @@
+# A GOV.UK-wide session cookie & login
+
+## Summary
+
+We propose introducing a new cookie, `govuk_session`, which will be an
+optional cookie, required only for users who wish to use personalised
+parts of GOV.UK (currently just the Transition Checker).
```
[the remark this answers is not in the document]
timblair: If we are likely to have sessions for unauthenticated users, might it be better to have just a single session type, which may or may not be attached to an account?

kevindew:
Thanks Michael for the various amends this is all looking good 👍 

With the cookie concern have we identified what scenario we're worried about? If we're expecting that the only way someone will get the session id is from the cookie leaking then I'm not sure how any of the additional protection measures help as someone could use a leaked encrypted cookie or a leaked signed cookie - so do they offer an advantage over plaintext? or is there another way we're concerned that the session id could leak? (an internal bad actor seems the most likely one but then that same actor would probably have access to secrets/keys)

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: A simple way of generating a strong random collision-resistant session ID would be something like this:

```ruby
def generate_session_id
  "#{Time.zone.now.to_f}-#{SecureRandom.hex(128/8)}"
end
```

- `Time.zone.now.to_f` gives the current Unix time in microseconds
- `SecureRandom.hex(128/8)` gives a 128-bit integer in hexadecimal

So we'd end up with session IDs like `1612452968.1587293-b74b7fd4077caf6cabf6632736313653`

Given that the cookie is using the `__Host-` prefix, it seems unlikely that an attacker would be able to obtain the user's session ID without directly nabbing the user's cookie (eg, through malware on their computer, or a MitM attack), and if they can do that then an encrypted or signed cookie doesn't help us.

So, so long as we don't expose the session ID to javascript (eg, including it in the page body), does encrypting it gain us much?

barrucadu:
A few of us are having a call on Tuesday morning to talk through the remaining issues around the exact auth flow used and what the cookie holds.  I've bumped the deadline for discussion to **2021-02-09**.

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: I'm going to resolve this because we've had a talk about it and changed the RFC.  If the new way is still bad we can start another thread.

rfc-134-govuk-wide-session-cookie-and-login.md
```diff
@@ -0,0 +1,359 @@
...
+update the data we hold on them.
+
+This works fine for one app, but there are problems with this approach
+when we try to scale to more than one app.
+
+[ff-session]: https://github.com/alphagov/finder-frontend/blob/5b6101e8eb027028a102ab96bcef6ff752b9c558/app/controllers/sessions_controller.rb#L23-L28
+
+### We need to pass cookies to our apps
+
+Our Nginx configuration blocks most cookies, [which we disabled for
```
barrucadu, line 33: @richardTowers Our of interest, will we have this sort of thing (a proxy removing undesirable headers, both inbound and outbound) in the replatformed world?

richardTowers: I've jotted it down on my board of things the current platform does, but it's not in the plan at the moment. I'm interested in how valuable we think the feature is.

bevanloon (approved):
Thanks for all the work on this Michael - it looks good to me.

Worth a second ✅  from @erino and/or @richardTowers.

richardTowers (approved):
Thanks for all your work on this @barrucadu!

I'm happy 🚀

erino (commented):
👍🏻 from me. Good work @barrucadu, looking forwarding to signing in across GOV.UK

rfc-134-govuk-wide-session-cookie-and-login.md
barrucadu: I suspect the main value we get out of it is not needing to confirm whether any of our apps try to set session cookies by default when they shouldn't do.

suganya-s:
Hey @barrucadu , I'm really late to this party but just wanted to check if this ID would be passed to Google Analytics?

barrucadu:
@suganya-s No, we've ended up deciding the cookie is going to store session data, not an ID.

suganya-s:
Thanks for clarifying @barrucadu, is there any way to join that data to Google Analytics in the works?

barrucadu:
Not yet.  We do have unique user IDs, so one of the ideas we talked about a while back (like triggering a GA event containing the ID when the user successfully logs in) will be doable.

PR #135, opened 2021-02-12
Updating routes in Router

karlbaker02:
[Rendered](https://github.com/alphagov/govuk-rfcs/blob/updating-routes-in-router/rfc-135-updating-routes-in-router.md)

**Deadline:** 2021-03-05

### DIFF
--- a/rfc-135-updating-routes-in-router.md
+++ b/rfc-135-updating-routes-in-router.md
@@ -1,0 +1,53 @@
+# Updating routes in Router
+
+## Summary
+
+This RFC proposes changing the way that routes are updated in Router, moving away from Router API manually updating Router instances and instead allowing Router instances to keep themselves up-to-date via polling MongoDB for changes.
+
+## Problem
+
+Whenever a route needs to be added or updated, Content Store calls into Router API to perform the update. Router API knows the address of every Router instance from a static file that it reads at start-up; it then uses this information to call the `/reload` endpoint on every Router instance to let them know that routes have been updated, so that those instances reload the routes they hold in memory.
+
+As GOV.UK moves to containerise more applications and host them in AWS Elastic Container Service (ECS), Router API will no longer be able to talk to individual instances of Router without additional changes; these changes can either be to rethink how Router instances are updated, or to implement a hacky solution which attempts to keep Router API's ability to talk to Router instances through making a DNS request to AWS App Mesh, get the IP addresses of Router instances and then _attempt_ to talk to them.
+
+The current implementation of updating routes has led to scaling issues. When scaling Router, Router API takes [5 minutes][] to get an accurate router node list. If Router API fails to update a Router node (i.e. one that is being brought up/down), the API also [fails to reach subsequent nodes in the list][].
+
+We should therefore rethink how these routes are updated within Router instances and to come up with another way of managing the update process.
+
+[5 minutes]: https://github.com/alphagov/govuk-puppet/blob/84689ae2c36bbc7740da6a8d70c7f4e04d0e2357/modules/govuk/manifests/apps/router_api.pp#L110
+[fails to reach subsequent nodes in the list]: https://github.com/alphagov/router-api/blob/2aa74d402624837c87fef136084ae66633e57d46/lib/router_reloader.rb#L46
+
+## Proposal
+
+Router will keep its in-memory cache of routes up-to-date via a two stage process:
+
+  1. Router will perform a simple query against MongoDB returning the state of the database with the current set of routes; ideally this will be a query which contains a checksum or timestamp of last update which can be used by Router instances to check whether their local copy of routes is up-to-date.
+  2. If the check against MongoDB concludes that Router needs to update its in-memory cache of routes, then Router will self-action a reload of all routes, using the same mechanism as it currently does when it receives a call to its `/reload` endpoint.
+
+The polling of changes by Router against MongoDB will take place every few seconds in order to ensure that routes are updated quickly as and when they change, which keeps in line with the current approach of updating routes when comparing how fast route updates propagate through to Router instances. Individual Router instances will drift slightly out of sync with each other, but because we're polling every few seconds and updating as soon as we see changes there won't be any noticeable changes.
+
+The query used to poll MongoDB for changes should be quick to compute, taking advantage of any existing indexes on the database or aggregates already stored by MongoDB; for example, the [db.stats() method](https://docs.mongodb.com/manual/reference/method/db.stats/) could be a useful starting point. The details of the query aren't included the scope of this RFC however and it should be assumed that a suitably performant query can be made which allows for the polling of changes to routes.
+
+This proposal is relatively simple to implement compared to some of the others we have considered and will allow us to move at pace in decoupling Router API and Router. It also puts us in a good place to iterate the solution further by using [the change streams feature](#using-mongodb-change-streams-to-notify-router-instances-of-route-changes) in future, whether that's offered by MongoDB, DocumentDB or any other database technology that we decide to use for storing routes.
+
+### Possible alternatives
+
+There were a number of approaches considered to tackle the problem presented in this RFC and [these approaches have been documented in detail](https://docs.google.com/document/d/1gGRWTmhfcfU1jfBDWYMiDa-6G64tRNdPFI3BLcbCxWc/edit#). Still, it is worth a brief summary of the main approaches discussed for completeness.
+
+#### Removing Router API entirely
+
+The first option available to us was to remove Router API entirely, which has been [discussed previously](https://github.com/alphagov/govuk-rfcs/pull/72) and for which there also exists a [tech debt card](https://trello.com/c/MOuw5ke0). Removal of Router API from the stack would be a good refactoring for GOV.UK to remove complexity, as Content Store already calls Router API, so why not remove this call and just have Content Store update routes directly.
+
+The downside is that this approach doesn't solve the underlying problem of this RFC. When apps are running in ECS / AppMesh, it is going to be difficult to contact individual instances of Router without doing something hacky. Removing Router API just moves the problem to Content Store not being able to talk to Router instances, it doesn't make it go away.
+
+#### Using MongoDB change streams to notify Router instances of route changes
+
+MongoDB has the ability to notify consumers of a dataset when data changes via [change streams](https://docs.mongodb.com/manual/changeStreams/), which could be used to completely decouple Router API and Router as outlined in this RFC.
+
+Currently this option isn't available to us, as the MongoDB database containing routes is running version `2.6.12`, with the change sets feature being introduced from MongoDB version `3.6.0`. There is upcoming work on the roadmap to upgrade this MongoDB version outside of the scope of this RFC and there is also a wider question about whether we might want to adopt a different database technology entirely for storing routes, such as DocumentDB or PostgreSQL. For these reasons we dismissed this approach.
+
+#### Using a message broker to notify Route of route changes
+
+GOV.UK already uses RabbitMQ, an open-source message broker which would allow us to completely decouple Router API and Router. Router API would push route updates to an exchange in RabbitMQ and Router instances would have updates automatically pushed to them from the queues subscribed to the exchange.
+
+There are benefits to using RabbitMQ such as message durability and the low overhead in introducing it to tackle this new problem, given that we already have the infrastructure in place. At the same time, Router API and Router instances both already talk to MongoDB, so adding in a message broker such as RabbitMQ adds yet another layer of complexity for something which could be handled as already outlined by MongoDB's change streams feature. For this reason we dismissed this approach.


rfc-135-updating-routes-in-router.md
```diff
@@ -0,0 +1,44 @@
+# Updating routes in Router
+
+## Summary
+
+This RFC proposes changing the way that routes are updated in Router, moving away from Router API manually updating Router instances and instead allowing Router instances to keep themselves up-to-date via polling MongoDB for changes.
+
+## Problem
+
+Whenever a route needs to be added or updated, Content Store calls into Router API to perform the update. Router API knows the address of every Router instance from a static file that it reads at start-up; it then uses this information to call the `/reload` endpoint on every Route instance to let them know that routes have been updated, so that those instances reload the routes they hold in memory.
```
thomasleese, line 9: I wonder if it's worth adding links to how this works to make it easier to find more information. I'm thinking for example https://github.com/alphagov/govuk-puppet/blob/84689ae2c36bbc7740da6a8d70c7f4e04d0e2357/modules/govuk/manifests/apps/router_api.pp#L94-L118 or https://github.com/alphagov/router-api/blob/2aa74d402624837c87fef136084ae66633e57d46/lib/router_reloader.rb#L8-L36.

rfc-135-updating-routes-in-router.md
```diff
@@ -0,0 +1,44 @@
...
+
+As GOV.UK moves to containerise more applications and host them in AWS Elastic Container Service (ECS), Router API will no longer be able to talk to individual instances of Router. We therefore need to rethink how these routes are updated within Router instances and to come up with another way of managing the update process.
+
+## Proposal
+
+Router will keep its in-memory cache of routes up-to-date via a two stage process:
+
+  1. Router will perform a simple query against MongoDB returning the state of the database with the current set of routes; ideally this will be a query which contains a checksum or timestamp of last update which can be used by Router instances to check whether their local copy of routes is up-to-date.
```
thomasleese, line 17: Do we have any more specifics on how this query might work? I guess we don't want to come up with a solution in the RFC, but I think a little more might be helpful to decide whether the overall idea is a good idea or not. (For example, I'm imagining we might need to think a little bit about database indexes or how long it would take to generate said checksum, etc).

thomasleese (commented):
This seems like a good approach, I've left a couple of comments on the text of the RFC itself.

rfc-135-updating-routes-in-router.md
```diff
@@ -0,0 +1,44 @@
...
+
+#### Removing Router API entirely
+
+The first option available to us was to remove Router API entirely, which has been [discussed previously](https://github.com/alphagov/govuk-rfcs/pull/72) and for which there also exists a [tech debt card](https://trello.com/c/MOuw5ke0). Removal of Router API from the stack would be a good refactoring for GOV.UK to remove complexity, as Content Store already calls Router API, so why not remove this call and just have Content Store update routes directly.
+
+The downside to this approach is that it doesn't solve the underlying problem of this RFC, namely that as part of moving apps to ECS and within AWS App Mesh, it's not going to be possible to talk to individual instances directly. Therefore the problem moves to Content Store not being able to talk to Router instances rather than Router API not being able to talk to Router instances.
```
bilbof, line 32: I think talking to individual instances should be possible - each router ECS task has an IP address. We just don't want to implement this, right? Perhaps we could say something like "we'd need to implement a way for Router API to reach Router instances in ECS, which we think is not a good approach given the incidents".

You could also add something about how removing Router API might not make things simpler, since it's quite a simple service app with essentially one endpoint. Moving Router API's logic into Content Store might introduce complexity (and perhaps unforeseen scaling issues).

bilbof (commented):
I think this approach will work, thanks Karl 👍 

Confusingly, we have both Router API calling Router instances, and Router instances calling Router API's DB, so it'll be good to remove one of those requests. It's a simpler design you've proposed 👏 

It would be nice to know how many router instances concurrently refreshing routes (a full reload) it'll take to overload the MongoDB instance. But we can tweak this design in future if we need to deal with scale mismatch. For example, Router API could push the config to an S3 bucket. There's a good article on scale mismatch in designs like this [here](https://aws.amazon.com/builders-library/avoiding-overload-in-distributed-systems-by-putting-the-smaller-service-in-control/).

I'd also be interested if you have thoughts on how Router polling MongoDB for changes might work?

rfc-135-updating-routes-in-router.md
karlbaker02: I wasn't sure how far to go into detail here, as like you say, we haven't done any real research into this and I imagine it'll be something we look at when it comes to implementation. If we're calling a method such as `db.stats()` (or using some other method which returns aggregates) then we might not need to worry about manually creating indexes etc as the queries _should_ already be performant.

On the note of how long it takes to query, so long as the queries return results in a short (and reasonable) amount of time I feel we will be fine.

rfc-135-updating-routes-in-router.md
karlbaker02: Good point, can do.

rfc-135-updating-routes-in-router.md
karlbaker02: I was under the impression that this _wouldn't_ actually be possible with ECS _and_ App Mesh, perhaps @richardTowers can clarify?

As for removing Router API and moving the logic into Content Store, I think the RFC mentioned in this RFC which discussed removing Router API covers that topic in good detail; I think this move _would_ introduce some additional complexity, but given that it's Content Store already calling Router API, I wouldn't have thought this might introduce scaling issues? Happy to be proven wrong here.

richardTowers: It _may_ be possible, I'm not sure I know enough about the details of how AppMesh does it's "load balancing" to be sure.

In my current mental model of how it works I think we could do:

* make a DNS request for `router.mesh.govuk-internal.digital`
* this will return the IP addresses of all the router instances
* contact each IP individually 🎉 

But ... is that what's really happening? Or is Envoy more involved than I suspect it is? Will it refuse to let us talk to a raw IP that it doesn't know about?

Even if this would work, it feels verrrryy hacky.

Maybe:

> The downside is that this doesn't solve the underlying problem of this RFC. When apps are running in ECS / AppMesh, it is going to be difficult to contact individual instances of router without doing something hacky. Removing Router API just moves the problem to Content Store not being able to talk to Router instances, it doesn't make it go away.

rfc-135-updating-routes-in-router.md
karlbaker02: Have added an additional paragraph about this, let us know your thoughts.

rfc-135-updating-routes-in-router.md
bilbof: Yeah I agree, we don't want to do this, [regardless of whether it's possible](https://www.youtube.com/watch?v=kY-pUxKQMUE). I think the fact that we could achieve it - but with hacks, would be good to note in the problem section.

An aside: we should really find out what Envoy does with requests 😬 . Also, the AWS APIs are an alternative to DNS to get a list of IP address. E.g. `aws ecs describe-tasks` returns the `privateIpv4Address` for a set of Fargate tasks.

karlbaker02: I have added Richard's points to the _problem_ section of this RFC as Bill suggests; take a look and let us know your thoughts.

bilbof: The problem section looks good, thanks @karlbaker02.

rfc-135-updating-routes-in-router.md
bilbof: A couple more ideas to demonstrate that the approach is feasible: [collection.dbHash](https://docs.mongodb.com/v2.6/reference/command/dbHash), or Router API could write a timestamp when the routes collection changes and not use a hash.

timblair:
We have an existing [tech debt card](https://trello.com/c/MOuw5ke0/22-router-api-exists) and [a (non-accepted) RFC](https://github.com/alphagov/govuk-rfcs/pull/72) about doing away with router-api entirely. How does this concept relate to what you're proposing?

richardTowers:
> We have an existing [tech debt card](https://trello.com/c/MOuw5ke0/22-router-api-exists) and [a (non-accepted) RFC](https://github.com/alphagov/govuk-rfcs/pull/72) about doing away with router-api entirely. How does this concept relate to what you're proposing?

It's mentioned in [the other options we considered](https://github.com/alphagov/govuk-rfcs/pull/135/files#diff-1d27edce5f0ef80cff38cf26ea71bad0690f4157b81c6efb51c8ca79a99e8a57R37-R41).

We still think that's a good thing to be doing, but it doesn't actually solve this problem. It just moves the problem from "router-api needs to contact individual router instances" to "content-store needs to contact individual instances".

We could fix both things at the same time (i.e. poll content store's DB instead of router's DB, remove router-api and its DB entirely). But I think it's quite a bit more work. Still worth considering though.

timblair:
🤦 My brain completely skipped over that part of the proposal.

I agree that it doesn't solve the current problem, but I was wondering if it's worth considering as part of this to remove that extra bit of infrastructure (router-api) which is essentially only there to provide an intermediary between the content store and the router.

richardTowers:
Let's do a quick 👍🏻  👎🏻  poll to see who prefers which option:

👍🏻  - clean up the architecture. Remove router-api, have router instances poll content-store's DB
👎🏻  - keep it simple for now. Keep router-api, have router instances poll router-api's DB.

timblair:
I've voted for "clean up the architecture," _but_ I'm not convinced that we actually know what that means, or if it's the right thing to do. What I _would_ like to see is some further thinking as to what the implications would be here. If we're doing significant work on the Router / Router API, I want to make sure that we're making the best use of that effort.

Questions to answer would include:

* Is it reasonable / desirable for the Content Store to be the source of truth for Router, and what would be the impact if it was?
* Are there any edge cases where an app talks directly to the Router API, or other times where something isn't registered in the Content Store in a way which would allow suitable, complete view of routes.

rfc-135-updating-routes-in-router.md
sengi: (Sorry for resurrecting old thread but it seemed like the most relevant place for this.)

Having Router API fetch the list of Router instances via DNS would have been a tempting solution IMO (replace reading the instances file with a DNS query and re-query+retry with backoff when contacting an instance fails), but it turns out that an arbitrary limitation in AWS makes it unviable: Amazon's cluster DNS (Cloud Map + ECS) only returns [8 instances maximum](https://docs.aws.amazon.com/cloud-map/latest/api/API_DnsConfig.html#cloudmap-Type-DnsConfig-RoutingPolicy). Amazon has an API which returns the full set of instances, but tying Router API to a proprietary cloud vendor API seems like a step too far (and having Router API watch the database is a more general solution as it removes unnecessary coupling between Router API and Router and should be less complex overall).

rfc-135-updating-routes-in-router.md
sengi: (Again, sorry for thread necromancy but seemed like the most relevant place to continue the discussion.)

To echo Tom's point, I think the details of the polling mechanism are quite crucial here and to defer their consideration until later as a mere implementation detail would be a mistake. The concept may be simple, but how Router API denotes that there's been a change is a different matter and I think it reduces to a non-trivial classical concurrency problem.

For example, having a "last updated" timestamp on the database would be super convenient for implementing the polling loop in Router, but how would Router API update that timestamp? Just write to a specific record in the database each time a change happens? That means all writes will contend for the same lock. Maybe that's fine, say if we're only writing at most a handful of times a second. How frequently do updates happen in the worst case? (Are there bulk republishing operations which generate a lot of changes all at once, for example?)

I'm pretty sure hashing is a red herring - reading through the entire dataset on every write to compute the hash would be infeasible, and doing something incremental involves fancy techniques which Mongo almost certainly doesn't implement. I don't think it's necessary either - a counter or a timestamp would do just fine (and even that is challenging enough!)

There are a couple of things in our favour. Firstly, Mongo already implements replication, so we may be able to leverage that (for example perhaps the replication counters are suitable to detecting when there's been an update). Secondly, we only need eventual consistency and we don't care about linearizability (I'm fairly sure we don't, anyway).

sengi: (tl;dr: let's figure out if we can get away with just polling `rs.status()`)

karlbaker02: For the updating of a timestamp as the mechanism for Router instances to detect changes, I would hope that Mongo can actually take care of this for us and what we then don't need to worry about Router API instances contending for the same lock; we just use the timestamp returned either in a call to one of the above mentioned methods. If I'm being naive here then please let me know!

rfc-135-updating-routes-in-router.md
```diff
@@ -0,0 +1,53 @@
+# Updating routes in Router
+
+## Summary
+
+This RFC proposes changing the way that routes are updated in Router, moving away from Router API manually updating Router instances and instead allowing Router instances to keep themselves up-to-date via polling MongoDB for changes.
+
+## Problem
+
+Whenever a route needs to be added or updated, Content Store calls into Router API to perform the update. Router API knows the address of every Router instance from a static file that it reads at start-up; it then uses this information to call the `/reload` endpoint on every Route instance to let them know that routes have been updated, so that those instances reload the routes they hold in memory.
```
ChrisBAshton, line 9:
```suggestion
Whenever a route needs to be added or updated, Content Store calls into Router API to perform the update. Router API knows the address of every Router instance from a static file that it reads at start-up; it then uses this information to call the `/reload` endpoint on every Router instance to let them know that routes have been updated, so that those instances reload the routes they hold in memory.
```

rfc-135-updating-routes-in-router.md
```diff
@@ -0,0 +1,53 @@
...
+
+There were a number of approaches considered to tackle the problem presented in this RFC and [these approaches have been documented in detail](https://docs.google.com/document/d/1gGRWTmhfcfU1jfBDWYMiDa-6G64tRNdPFI3BLcbCxWc/edit#). Still, it is worth a brief summary of the main approaches discussed for completeness.
+
+#### Removing Router API entirely
+
+The first option available to us was to remove Router API entirely, which has been [discussed previously](https://github.com/alphagov/govuk-rfcs/pull/72) and for which there also exists a [tech debt card](https://trello.com/c/MOuw5ke0). Removal of Router API from the stack would be a good refactoring for GOV.UK to remove complexity, as Content Store already calls Router API, so why not remove this call and just have Content Store update routes directly.
+
+The downside is that this approach doesn't solve the underlying problem of this RFC. When apps are running in ECS / AppMesh, it is going to be difficult to contact individual instances of router without doing something hacky. Removing Router API just moves the problem to Content Store not being able to talk to Router instances, it doesn't make it go away.
```
ChrisBAshton, line 41:
```suggestion
The downside is that this approach doesn't solve the underlying problem of this RFC. When apps are running in ECS / AppMesh, it is going to be difficult to contact individual instances of Router without doing something hacky. Removing Router API just moves the problem to Content Store not being able to talk to Router instances, it doesn't make it go away.
```

rfc-135-updating-routes-in-router.md
```diff
@@ -0,0 +1,53 @@
...
+
+The query used to poll MongoDB for changes should be quick to compute, taking advantage of any existing indexes on the database or aggregates already stored by MongoDB; for example, the [db.stats() method](https://docs.mongodb.com/manual/reference/method/db.stats/) could be a useful starting point. The details of the query aren't included the scope of this RFC however and it should be assumed that a suitably performant query can be made which allows for the polling of changes to routes.
+
+This proposal is relatively simple to implement compared to some of the others we have considered and will allow us to move at pace in decoupling Router API and Router. It also puts us in a good place to iterate the solution further by using the change streams feature in future, whether that's offered by MongoDB, DocumentDB or any other database technology that we decide to use for storing routes.
```
ChrisBAshton, line 31: Would be good to link to a definition, for people unfamiliar with change streams (me!):

```suggestion
This proposal is relatively simple to implement compared to some of the others we have considered and will allow us to move at pace in decoupling Router API and Router. It also puts us in a good place to iterate the solution further by using [the change streams feature](#using-mongodb-change-streams-to-notify-router-instances-of-route-changes) in future, whether that's offered by MongoDB, DocumentDB or any other database technology that we decide to use for storing routes.
```

ChrisBAshton (approved):
LGTM 👍  nice write-up!

rfc-135-updating-routes-in-router.md
karlbaker02: Another option we have is to use the primary machine's `oplog`. [From the docs](https://docs.mongodb.com/manual/core/replica-set-oplog/)...

- _The oplog (operations log) is a special capped collection that keeps a rolling record of all operations that modify the data stored in your databases._

- _MongoDB applies database operations on the primary and then records the operations on the primary’s oplog. The secondary members then copy and apply these operations in an asynchronous process. All replica set members contain a copy of the oplog, in the local.oplog.rs collection, which allows them to maintain the current state of the database._

- _Each operation in the oplog is idempotent._

From the oplog, we can call the method `rs.printReplicationInfo()` to get the timestamp of the last update to the primary which can then be used to poll for changes in Router instances.

Example output of `rs.printReplicationInfo()` on the Router database:
```
rs.printReplicationInfo()
configured oplog size:   990MB
log length start to end: 130748secs (36.32hrs)
oplog first event time:  Mon Mar 08 2021 03:24:12 GMT+0000 (UTC)
oplog last event time:   Tue Mar 09 2021 15:43:20 GMT+0000 (UTC)
now:                     Tue Mar 09 2021 16:21:54 GMT+0000 (UTC)
``` 

Importantly, as the oplog is a log of all operations (modifications) performed on the primary, there is no need to perform a table scan or other high-workload operation on the primary.

sengi: Aha, so I was talking nonsense about the 1-second precision being a problem. We can just compare Timestamp objects from `rs.status()` and we should be fine?

karlbaker02: Just to clarify what's being said here, we could use the `optime` timestamp that's returned when calling `rs.status()`, which is the internal MongoDB BSON timestamp. Importantly though, it _does_ give us the actual timestamp and the ordinal which together form the most recent update to a given database instance, whether the primary or one of the secondaries. The format it returns this in is:

```
...
"optime" : Timestamp(1615383267, 4),
...
```

We are given this timestamp for all members of the replica set, so as long as we're consistent about which replica we're reading from, Router instances should be able to use this value when polling to determine whether there are any new route updates to fetch and if so, to reload it's routes from MongoDB. This timestamp gives us confidence that there will be no race conditions of the kind described previously, so there's also no need to do any force refreshes of routes in Router.

PR #136, opened 2021-02-25
RFC to remove our Google Cloud CDN backup

richardTowers:
✨ [Rendered](https://github.com/alphagov/govuk-rfcs/blob/remove-the-gcp-backup-cdn/rfc-136-remove-our-backup-cdn-in-gcp.md) ✨ 

**Deadline**: 2021-03-17

### DIFF
--- a/rfc-136-remove-our-backup-cdn-in-gcp.md
+++ b/rfc-136-remove-our-backup-cdn-in-gcp.md
@@ -1,0 +1,130 @@
+# Remove the backup CDN in Google Cloud Platform (GCP)
+
+## Summary
+
+We have two backup content delivery networks (CDNs), in case Fastly have an outage.
+The first is AWS CloudFront, the second is Google Cloud CDN.
+
+A simultaneous outage of both Fastly and AWS CloudFront is very unlikely. So
+the Google Cloud CDN backup provides very little extra reliability.
+
+At the same time, the cost of maintaining the Google Cloud CDN backup is
+non-trivial.
+
+The maintenance cost outweighs the expected reliability benefit, so we should
+remove the Google Cloud CDN backup.
+
+## Problem
+
+GOV.UK's current business continuity plan aims to ensure that members of the
+public can continue to use www.gov.uk, even in the event of a major outage in
+one or more of our infrastructure providers.
+
+Firstly, while our primary CDN (Fastly) is up, any failed requests to our origin servers
+will be retried against three static mirrors - two in AWS, one in GCP.
+
+* origin
+* AWS S3
+* AWS S3 (replica in a different region)
+* Google Cloud Storage
+
+If none of the above are successful, an error is served to the user.
+
+Secondly, if our primary CDN (Fastly) is down, we have two backup CDNs which we
+can manually fail over to (by updating a DNS record). AWS CloudFront and Google
+Cloud CDN.
+
+The mirroring approach works well. This RFC does not propose any change to the
+mirrors, including the Google Cloud Storage mirror.
+
+The AWS CloudFront CDN is reasonably easy to maintain and test, however the
+Google Cloud CDN backup is significantly more difficult.
+
+### Our Google Cloud CDN setup currently broken
+
+The certificate in use on the CDN expired in May 2020.
+
+The Google Cloud Storage bucket the CDN is pointing to is currently private.
+This means even if you ignore certificate warnings you'll see an access
+denied error.
+
+These issues could be fixed as part of the failover process, but we don't
+have any documentation explaining how to do that.
+
+### It is difficult to issue a certificate for Google Cloud CDN
+
+GCP provide managed certificates for their CDN, but these can only be issued
+and renewed if the CDN is serving live traffic (i.e. www.gov.uk's [DNS A
+records point to the CDN's IP](https://cloud.google.com/load-balancing/docs/ssl-certificates/google-managed-certs#update-dns)).
+
+This doesn't work for us because we need our DNS A records to be pointing at
+Fastly for www.gov.uk to work properly.
+
+We could issue certificates from a third party certificate provider (for
+example GlobalSign or LetsEncrypt).
+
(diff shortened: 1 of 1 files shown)


rfc-136-remove-our-backup-cdn-in-gcp.md
```diff
@@ -0,0 +1,70 @@
+# Remove the backup CDN in Google Cloud Platform
+
+## Summary
+
+We have two backup content delivery networks (CDNs), in case Fastly have an outage.
+The first is AWS CloudFront, the second is Google Cloud CDN.
+
+A simultaneous outage of both Fastly and AWS CloudFront is very unlikely. So
+the Google Cloud CDN backup provides very little extra reliability.
```
bilbof, line 9: It might be worth illustrating the probability of two CDNs being down? Both AWS CloudFront's SLA and Fastly's SLA (depending on our account) offer a discount if we get less than 99.99% availability (53 minutes of downtime/service/year). Does running two CDNs give us 6 nines (~30s of simultaneous downtime)? If so, running an additional CDN (presumably with an extra four nines) seems excessive, it'd probably take us longer than 30s to failover, and there are bigger problems affecting GOV.UK's availability.

bilbof (approved):
Excellent. Might be worth linking to the original decision / explanation for why we have 3 CDNs, if there is such a decision document.

rfc-136-remove-our-backup-cdn-in-gcp.md
richardTowers: Added: https://github.com/alphagov/govuk-rfcs/pull/136/files#diff-dec23adcf70b72a15b0c38431315d7edecbbe25c093cdcc7b225fa2b95ee6d68R91-R109

ChrisBAshton (approved):
LGTM 👍

bilbof (approved):
Thanks, great work

rfc-136-remove-our-backup-cdn-in-gcp.md
```diff
@@ -0,0 +1,130 @@
...
+mirrors, including the Google Cloud Storage mirror.
+
+The AWS CloudFront CDN is reasonably easy to maintain and test, however the
+Google Cloud CDN backup is significantly more difficult.
+
+### Our Google Cloud CDN setup currently broken
+
+The certificate in use on the CDN expired in May 2020.
+
+The Google Cloud Storage bucket the CDN is pointing to is currently private.
```
fredericfran-gds, line 47: This is intended behaviour to save money in case someone/thing decides to crawl against the Google bucket. As part of the playbook for activating the Google CDN, we would mark the bucket as public and re-point the www.gov.uk DNS to the Google Cloud CDN fronting the bucket.

The main issue we have is the certificate issuing before the live traffic switch.

richardTowers: It's interesting that we don't bother with this for the CloudFront / S3 backup. What's to stop something from crawling that?

In any case, I think it's still worth mentioning this as something that's broken. If we had documentation on how to fail over which included making the bucket public, that would be fine. But we don't.

barrucadu (approved):
Having a second backup CDN seems mad, especially if we didn't notice that it was broken (expired certificate) for ~10 months.

Putting in the work to fix this and make it testable seems far less valuable than spending the same time improving other parts of GOV.UK... or just on other non-reliability-related work too.

rfc-136-remove-our-backup-cdn-in-gcp.md
fredericfran-gds: Umm.. my understanding is that the S3 buckets should be protected, see here: https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-mirror-bucket/mirror-read-policy.tf 

I don't know what policy they had on the Cloudfont side as I didn't do that work. My thoughts were they were going to be only office IPs allowed to use the Cloudfront endpoint using test and we widen the whitelist once it is made operational.

fredericfran-gds (approved):
unlikely we would ever use Google cloud CDN and even the AWS CloudFront since it will have so little functionality that the higher powers will pile in pressure to fix Fastly etc.

huwd (approved):

rfc-136-remove-our-backup-cdn-in-gcp.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Summary
+
+We have two backup content delivery networks (CDNs), in case Fastly have an outage.
+The first is AWS CloudFront, the second is Google Cloud CDN.
+
+A simultaneous outage of both Fastly and AWS CloudFront is very unlikely. So
+the Google Cloud CDN backup provides very little extra reliability.
+
+The infrastructure cost of the Google Cloud CDN backup is neglibible, but the
```
brucebolt, line 11:
```suggestion
The infrastructure cost of the Google Cloud CDN backup is negligible, but the
```

brucebolt (commented):
This looks good to me.

I have a couple of questions:
- Have we confirmed Fastly do not use any AWS infrastructure behind the scenes (if so, a AWS outage that takes Fastly offline could render the AWS option unusable at the same time).
- I notice assets (mostly images) on www.amazon.co.uk have a `x-cache` header that indicates they use Fastly.  Have we considered why Amazon are using a competitor and not using their own service?  I'm hoping this is just down to better performance, but it may be worth researching if any information about this is public.

rfc-136-remove-our-backup-cdn-in-gcp.md
richardTowers: 🤣  I like my spelling better.

brucebolt (approved):
Thanks @richardTowers.  I couldn't find any further information on either of my questions either, so this all sounds good to me.

bevanloon (approved):

himal-mandalia (approved):
LGTM

richardTowers:
The CDN has now been removed.

PR #138, opened 2021-03-23
Enable Brotli compression on GOV.UK

Nooshu:
✨ [Rendered](https://github.com/alphagov/govuk-rfcs/blob/enable-brotli-compression/rfc-138-enable-brotli-compression.md) ✨ 

**Deadline**: 2021-03-30


rfc-138-enable-brotli-compression.md
```diff
@@ -0,0 +1,88 @@
...
+
+Showing a 5.56% reduction.
+
+### Summary 
+
+In all the pages tested across the frontend apps (17 in total) we see between 3-6% reduction in the number of bytes a browser is having to download over the network. This saving is being converted to improvements in the visual metrics a users is seeing in the browser.
+
+## Proposal
+
+The proposal is to enable Brotli compression at the edge using the [Fastly Brotli Compression LA](https://www.fastly.com/release-notes/q3-2020#brotli). Browsers that support Brotli will receive these versions of the files. Any browsers that don't will receive the gzip version of the file. This can therefore be applied using the progressive enhancement methodology. In doing so we MUST update the configuration VCL so as to automate this process. During testing it was simply enabled via the Fastly user interface.
```
thomasleese, line 86: In terms of how we enable this in VCL, have we done any brief investigation in how we might do this?

I found https://developer.fastly.com/reference/vcl/variables/backend-response/beresp-brotli/ but it says "Setting this will unset beresp.gzip." so does that mean we need to parse the `Accept-Encoding` header to set the right `beresp` variable accordingly?

I realise in an RFC we may not want to go too much into the implementation detail, but I just wondered if we'd already looked into it?

thomasleese (commented):
Definitely like this, and the improvements look good. I've left a few comments, but nothing major.

rfc-138-enable-brotli-compression.md
```diff
@@ -0,0 +1,88 @@
...
+| /browse/driving/driving-licences   | collections             | 278,385    | 265,832   | -4.51    |
+| /contact                           | feedback                | 242,047    | 232,124   | -4.10    |
+| /search/all                        | finder-frontend         | 313,400    | 295,979   | -5.56    |
+| /bank-holidays                     | frontend                | 263,686    | 251,555   | -4.60    |
+| /national-minimum-wage-rates       | government-frontend     | 266,699    | 254,073   | -4.73    |
+| /info/coronavirus/business-support | info-frontend           | 233,133    | 226,382   | -2.90    |
+| /licence-finder/sectors            | licence-finder          | 242,555    | 234,969   | -3.13    |
+| /guidance/immigration-rules        | manuals-frontend        | 250,431    | 240,763   | -3.86    |
+| /service-manual/service-standard   | service-manual-frontend | 254,529    | 245,881   | -3.40    |
+| /additional-commodity-code         | smart-answers           | 257,766    | 247,293   | -4.06    |
```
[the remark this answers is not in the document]
Nooshu: Thanks. I've now run your updated URL through our WPT instance and will update the stats accordingly.

rfc-138-enable-brotli-compression.md
Nooshu: Thanks for flagging @thomasleese , I've now raised a support ticket with Fastly asking for more information on how this is enabled on production and what config needs to be changed.

richardTowers: In staging and integration, I just click-ops'd it in the Fastly UI.

If we wanted to do that in production, we'd need to [stop our deployment script from unsetting the compression settings](https://github.com/alphagov/govuk-cdn-config/pull/328).

Perhaps it's possible to do the same thing in pure VCL, and not touch the UI.

Alternatively, we could consider using the Fastly terraform provider to manage this, but perhaps that's a wider can of worms.

Nooshu: Update from Fastly support:
> ...enabling it is the toggle and the set `beresp.brotli` in VCL since this has already been enabled for your account

rfc-138-enable-brotli-compression.md
```diff
@@ -0,0 +1,88 @@
+# Enable Brotli compression on GOV.UK
+
+## Summary
+
+Brotli compression is a new compression algorithm that is available in modern browsers that can offer 10-20% better compression over gzip compression. By enabling this option on the CDN users with browsers that support it will receive smaller static files over the wire (HTML, CSS, JavaScript), and those that don't will continue to receive assets that are gzipped.
+
+## Problem
+
+On GOV.UK we should always be looking to reduce the number of bytes a user has to download to view our pages. This is both good for web performance and data consumption on user devices. Any browser that supports WOFF2 fonts will also support Brotli compression (since this is the compression used in the fonts). According to the latest stats from [caniuse.com](https://caniuse.com/brotli), the Brotli `Accept-Encoding`/`Content-Encoding` functionality is supported by 95% of user browsers globally, and I expect this figure will actually be huger is we only consider UK based users:
```
ChrisBAshton, line 9:
```suggestion
On GOV.UK we should always be looking to reduce the number of bytes a user has to download to view our pages. This is both good for web performance and data consumption on user devices. Any browser that supports WOFF2 fonts will also support Brotli compression (since this is the compression used in the fonts). According to the latest stats from [caniuse.com](https://caniuse.com/brotli), the Brotli `Accept-Encoding`/`Content-Encoding` functionality is supported by 95% of user browsers globally, and I expect this figure will actually be bigger if we only consider UK based users:
```

rfc-138-enable-brotli-compression.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Proposal
+
+The proposal is to enable Brotli compression at the edge using the [Fastly Brotli Compression LA](https://www.fastly.com/release-notes/q3-2020#brotli). Browsers that support Brotli will receive these versions of the files. Any browsers that don't will receive the gzip version of the file. This can therefore be applied using the progressive enhancement methodology. In doing so we MUST update the configuration VCL so as to automate this process. During testing it was simply enabled via the Fastly user interface.
+
+We SHOULD also purge the static cache once completed is it was found during testing that certain Fastly PoP's continued to serve the uncompressed version of the file since there was a mismatch between `Accept-Encoding` header from the browser and `Content-Encoding` version available on the CDN. Once the cache was purged this issue was rectified. This issue will rectify itself over time anyway even without the manual purging since Rails inserts an MD5 fingerprint into the filename of each file. Once a file is changed it will have a new URL, thus invalidating the old cache.
```
ChrisBAshton, line 88:
```suggestion
We SHOULD also purge the static cache once completed as it was found during testing that certain Fastly PoP's continued to serve the uncompressed version of the file since there was a mismatch between `Accept-Encoding` header from the browser and `Content-Encoding` version available on the CDN. Once the cache was purged this issue was rectified. However, this issue would rectify itself over time even without the manual purging, since Rails inserts an MD5 fingerprint into the filename of each file. Once a file is changed it will have a new URL, thus invalidating the old cache.
```

rfc-138-enable-brotli-compression.md
```diff
@@ -0,0 +1,88 @@
...
+
+And we see the a reduction in bytes for all the expected assets:
+
+![bytes for assets compared](rfc-138/total-bytes-bank-holidays.png)
+
+Showing a 4.60% reduction.
+
+### Search page
+
+The visual progress of the search page has improved by approximately 300ms initially, 200ms at the end. As can be seen in the visual progress graph:
```
ChrisBAshton, line 70: Should this be 400ms initially?

rfc-138-enable-brotli-compression.md
```diff
@@ -0,0 +1,88 @@
...
+
+![Visual progress of the homepage using both compression methods.](rfc-138/visual-progress-homepage.png)
+
+And we see the a reduction in bytes for all the expected assets:
+
+![bytes for assets compared](rfc-138/total-bytes-homepage.png)
+
+### Coronavirus page
+
+The visual progress graph for this page starts off worse by 300ms, but quickly catches up and the viewport for Brotli completes rendering 200ms before gzip:
```
ChrisBAshton, line 46: Should this be "starts off worse by 400ms"?

ChrisBAshton (commented):
Proposal looks good 👍  what a great performance improvement with no apparent downsides!

I was a little thrown by the colours of the charts swapping over halfway through, might be worth fixing that, but I got the idea.

Nooshu:
> I was a little thrown by the colours of the charts swapping over halfway through, might be worth fixing that, but I got the idea.

Good spot! Yes I'll re-order the tests so they are consistent .

rfc-138-enable-brotli-compression.md
Nooshu: Yes! Thanks. Updated.

rfc-138-enable-brotli-compression.md
Nooshu: Yes! Thanks. Updated.

richardTowers (approved):
Sounds like a good plan to me! A nice performance win, without any drawbacks spotted so far.

I'm still a little unsure if this is something we can enable in pure VCL, or if we need to click the toggle in the Fastly UI (and therefore need to update our deployment script to stop it from untoggling the toggle). That's an implementation detail though, it doesn't need to hold the RFC up. 🚢

Nooshu:
Looks like the Fastly UI adds the following VCL config when it is enabled:

```
if ((beresp.status == 200 || beresp.status == 404) && (beresp.http.content-type ~ "^(text/html|application/x-javascript|text/css|application/javascript|text/javascript|application/json|application/vnd\.ms-fontobject|application/x-font-opentype|application/x-font-truetype|application/x-font-ttf|application/xml|font/eot|font/opentype|font/otf|image/svg\+xml|image/vnd\.microsoft\.icon|text/plain|text/xml)\s*($|;)" || req.url ~ "\.(css|js|html|eot|ico|otf|ttf|json|svg)($|\?)" ) ) {
  # always set vary to make sure uncompressed versions dont always win
  if (!beresp.http.Vary ~ "Accept-Encoding") {
    if (beresp.http.Vary) {
      set beresp.http.Vary = beresp.http.Vary ", Accept-Encoding";
    } else {
       set beresp.http.Vary = "Accept-Encoding";
    }
  }
  if (req.http.Accept-Encoding == "br") {
    set beresp.brotli = true;
  } elsif (req.http.Accept-Encoding == "gzip") {
    set beresp.gzip = true;
  }
}
```

As mentioned in @thomasleese [comment above](https://github.com/alphagov/govuk-rfcs/pull/138#discussion_r599553965).

PR #139, opened 2021-03-31
RFC-139: Enabling HTTP/3 on GOV.UK

Nooshu:
✨ [Rendered](https://github.com/alphagov/govuk-rfcs/blob/enable-http3/rfc-139-enable-http3.md) ✨ 

**Deadline**: 2022-07-14


rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+### HTTP/3
+To solve the HOL blocking issue at the network layer, HTTP/3 uses a completely new transport layer protocol called QUIC. QUIC initially stood for "Quick UDP Internet Connections", but the IETF now don't use it as an acronym. It is simply the name of the protocol. QUIC is built upon User Datagram Protocol (UDP). Since QUIC uses UDP, it isn't bound by the restrictions that TCP imposes. So it has now been designed so that the network layer now understands the the concept of independent streams, which was restricted to the application layer under HTTP/2:
+
+![Difference between TCP and QUIC when downloading assets.](rfc-139/QUIC-illustration-final.jpg)
+
+In the illustration above, if packet 2 is lost (using HTTP/2), packets 3-8 are blocked until retransmission occurs. Using QUIC if packet 2 is lost only packets 1 and 3 are blocked since they exist in the same stream. The other streams can progress as normal since they are fully independent. QUIC doesn't require in-order packet transmission across the whole connection, but it does still retain ordering within each resource stream.
```
karlbaker02, line 52: In the illustration and when using QUIC, if packet 2 is lost, would it be both packets 1 and 3 which are blocked (if transmitted out of order), or only packet 3 which is lost (if transmitted in order)? Apologies if I've missed this.

karlbaker02:
It's a very interesting read, very informative about the history of HTTP and the benefits that HTTP/3 brings - thanks for helping me to learn something new! I can see that HTTP/3 is definitely of value to us in making GOV.UK as responsive and as accessible to as broad a range of users as possible. 

Out of curiosity and in no way to diminish the benefits described as we need to cater to all users, do we have any idea of the percentage of users coming to GOV.UK which are likely to be experiencing the slow connection speeds which HTTP/3 will specifically help out with?

On browsers, I noticed that there isn't any mention of Internet Explorer (though there is of Edge). I may be behind the times here, but have we stopped officially supporting Internet Explorer? Related to this, and to the previous point around the users that are experiencing the slow connection speeds, do we have a rough idea of the percentages using HTTP/3 enabled browsers?

rfc-139-enable-http3.md
Nooshu: So both packet 1 and 3 would be blocked. Packet 1 and 3 would exist in a buffer on the client waiting for packet 2 to arrive (under QUIC). The ordering of the packets is retained within independent streams under QUIC. All packets need to be present before the whole stream (and therefor the asset) can be processed by the client.

Nooshu:
> Out of curiosity and in no way to diminish the benefits described as we need to cater to all users, do we have any idea of the percentage of users coming to GOV.UK which are likely to be experiencing the slow connection speeds which HTTP/3 will specifically help out with?

This is a really interesting question. The only way I can see us being able to measure this is via the Fastly CDN logs since this is network level information. They provide a set of metrics [we can use to measure this](https://developer.fastly.com/reference/vcl/variables/client-connection/). Even then I doubt we could pinpoint an actual number of users. But we would have an idea in terms of percentage across all connections. e.g. 1% of connections experience some form of packet loss / connection instability.

> On browsers, I noticed that there isn't any mention of Internet Explorer (though there is of Edge). I may be behind the times here, but have we stopped officially supporting Internet Explorer? Related to this, and to the previous point around the users that are experiencing the slow connection speeds, do we have a rough idea of the percentages using HTTP/3 enabled browsers?

We still support IE11, but it is no longer updated by Microsoft. Being such an old browser IE11 only supports [HTTP/2 on Windows 10+ OS's](https://caniuse.com/http2). It won't be effected by a transition to HTTP/3 since it won't understand the protocol.

As for the percentage of users who could currently use HTTP/3, we are looking at 55% examining the March 2021 stats from GA.

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+Users in these areas still have a need to access government services and the information that GOV.UK offers, but under these conditions it won't be a very pleasant experience. These are the users that will benefit from a transition to HTTP/3.
+
+## Brief history of HTTP
+Each iteration of the HTTP protocol has set out to improve the web performance of the previous version of the protocol.
+
+### HTTP/1.0
+One of the first versions of the HTTP protocol came with a number of web performance issues. A primary issue being that only a single asset could be download on a Transmission Control Protocol (TCP) connection before the connection was then closed. There was no way for these connections to persist for multiple asset downloads. This resulted in a lot of waisted negotiation time to establish separate TCP connections for each asset.
```
ChrisBAshton, line 19:
```suggestion
The first version of the HTTP protocol came with a number of web performance issues, a primary issue being that only a single asset could be downloaded on a Transmission Control Protocol (TCP) connection before the connection was then closed. There was no way for these connections to persist for multiple asset downloads. This resulted in a lot of wasted negotiation time to establish separate TCP connections for each asset.
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+HTTP/1.1 fixed the persistent connections issue, allowing multiple assets to be downloaded over a single connection once established. Unfortunately this didn't fully solve the web performance issues seen with the HTTP protocol (stemming from the TCP layer). 
+
+Under HTTP/1.1 the browser must download 100% of an asset before it can move onto the next asset. Meaning it can only download a single asset at a time per connection, the other assets are blocked. This is called Head-of-Line Blocking (HOL Blocking). Browser vendors quickly tried to solve this issue by allowing a browser to open up to 6 TCP connections per domain, so 6 assets can be downloaded in parallel.
+
+Unfortunately a number of issues still exist with this method:
+
+1. Most sites have many more than 6 assets per page, so HOL blocking is still occurring in almost every page load under HTTP/1.1.
+2. Each TCP connection is independent and have no knowledge of each other. All 6 are fighting for limited bandwidth which in turn can saturate the network.
+3. Many websites resorted to domain 'sharding', so as to allow a browser to open up more that 6 TCP connections, in turn exacerbating the problem mentioned in point 2.
```
ChrisBAshton, line 30: May be worth linking to a definition of domain sharding, or inlining a definition here:

```suggestion
3. Many websites resorted to "domain sharding" (hosting assets across multiple domains), to allow a browser to open up more that 6 TCP connections. This in turn exacerbated the problem mentioned in point 2.
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+![An example of HTTP/2 frames within TCP packets.](rfc-139/5_H2_2files_multiplexed.png)
+
+Source: ["Head-of-Line Blocking in QUIC and HTTP/3: The Details"](https://calendar.perfplanet.com/2020/head-of-line-blocking-in-quic-and-http-3-the-details/) by [Robin Marx](https://twitter.com/programmingart).
+
+Unfortunately this didn't solve all the HOL blocking issues. Since HTTP/2 is built upon TCP, and TCP is a protocol that offers the reliable transmission of packets. TCPs job is to make sure that the packets leaving one device arrive at another device in the exact same order and without any missing packets. When packet loss occurs, this is when we again see HOL blocking. When a packet of data is lost over the network, this then holds up the whole (single) TCP connection. Packets behind the missing packet can't be processed until retransmission occurs.
```
ChrisBAshton, line 43:
```suggestion
Unfortunately this didn't solve all the HOL blocking issues. HTTP/2 is built upon TCP: a protocol that offers the reliable transmission of packets. TCP's job is to make sure that the packets leaving one device arrive on the other device in exactly the same order, and with no missing packets. If packet loss occurs, TCP re-transmits the packets. Packets behind the missing packet cannot be processed until retransmission occurs, so all assets being transferred on that single TCP connection are held up.
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+This is where we are currently at with GOV.UK. The vast majority of users on strong, stable connection will have seen improved performance from enabling HTTP/2. Unfortunately, some users may have seen performance drop if they use a connection with high packet loss. This is due to the HOL blocking that now exists at the network layer. This is a primary issue HTTP/3 sets out to solve.
```
ChrisBAshton, line 45:
```suggestion
This is where we are currently at with GOV.UK. The vast majority of users on strong, stable connections will have seen improved performance from enabling HTTP/2. Unfortunately, some users may have seen performance drop if they use a connection with high packet loss. This is due to the HOL blocking that now exists at the network layer. This is a primary issue HTTP/3 sets out to solve.
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+### HTTP/3
+To solve the HOL blocking issue at the network layer, HTTP/3 uses a completely new transport layer protocol called QUIC. QUIC initially stood for "Quick UDP Internet Connections", but the IETF now don't use it as an acronym. It is simply the name of the protocol. QUIC is built upon User Datagram Protocol (UDP). Since QUIC uses UDP, it isn't bound by the restrictions that TCP imposes. So it has now been designed so that the network layer now understands the the concept of independent streams, which was restricted to the application layer under HTTP/2:
```
ChrisBAshton, line 48:
```suggestion
To solve the HOL blocking issue at the network layer, HTTP/3 uses a completely new transport layer protocol called QUIC. QUIC initially stood for "Quick UDP Internet Connections", but the IETF now don't use it as an acronym. It is simply the name of the protocol. QUIC is built upon the User Datagram Protocol (UDP), so isn't bound by the restrictions of TCP. It has been designed so that the network layer now understands the the concept of independent streams, which was restricted to the application layer under HTTP/2:
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+In the illustration above, if packet 2 is lost (using HTTP/2), packets 3-8 are blocked until retransmission occurs. Using QUIC if packet 2 is lost only packets 1 and 3 are blocked since they exist in the same stream. The other streams can progress as normal since they are fully independent. QUIC doesn't require in-order packet transmission across the whole connection, but it does still retain ordering within each resource stream.
+
+## The effect of packet loss
+
+Using WebPageTest we can roughly simulate the effect that packet loss has on performance for our users. For testing I picking a fairly content heavy page like the [past Prime Ministers](https://www.gov.uk/government/history/past-prime-ministers) page and a low spec Moto G4 Android device. For the connection speed I've picked on one of the worse case scenarios from the [MLab data](https://docs.google.com/spreadsheets/d/1UTRucwLY2VC4944-cKunL9yUPJkjDKXDdIPk6GY3610/edit#gid=0). I'm emulating the connection the village of Colintraive has, which reportedly has a 450/230 Kbps connection speed. This is a connection speed equivalent to a 2G or slow 3G connection.
```
ChrisBAshton, line 56: I don't think we need to link to the data again, as it's already established at the beginning of the document:

```suggestion
Using WebPageTest we can roughly simulate the effect that packet loss has on performance for our users. For testing I picking a fairly content heavy page like the [past Prime Ministers](https://www.gov.uk/government/history/past-prime-ministers) page and a low spec Moto G4 Android device. For the connection speed I've picked on one of the worse case scenarios from the MLab data. I'm emulating the connection that the village of Colintraive has, which is reportedly a 450/230 Kbps connection speed, roughly equivalent to a 2G or slow 3G connection.
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Visual progress graph
+The visual progress graph clearly shows the effect packet loss has on page load times. As the packet loss increases, the page load slows down:
+
+![The visual progress graph comparing the percentage completeness of a page over time.](rfc-139/visual-progress.png)
+
+HTTP/3 + QUIC is designed to help improve the page load speed for users experiencing these connection conditions in the real world. 
+
+### Connection quality measurement
+Is it possible to measure the connection quality that our users are experiencing when visiting GOV.UK. Fastly provide a [range of Varnish Configuration Language (VCL) options](https://developer.fastly.com/reference/vcl/variables/client-connection/) that we can use to monitor the quality of user connections should we wish to capture this data:
```
ChrisBAshton, line 73: I assume this is a statement rather than a question?

```suggestion
It is possible to measure the connection quality that our users are experiencing when visiting GOV.UK. Fastly provide a [range of Varnish Configuration Language (VCL) options](https://developer.fastly.com/reference/vcl/variables/client-connection/) that we can use to monitor the quality of user connections should we wish to capture this data:
```

I'm not quite sure what the aim of this section is, though. Are you proposing that as part of this RFC, we _should_ leverage some VCL to start monitoring the packet loss of our users?

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+- [client.socket.tcpi_rttvar](https://developer.fastly.com/reference/vcl/variables/client-connection/client-socket-tcpi-rttvar/) - Measures the variance in RTT of the connection. The higher the variance value, the less stable the connection.
+- [client.socket.tcpi_delivery_rate](https://developer.fastly.com/reference/vcl/variables/client-connection/client-socket-tcpi-delivery-rate/) - Gives a recent value for the delivery bandwidth to the client in bytes per second.
+
+## Our Users
+
+Since HTTP/3 + QUIC offers better performance for users on unstable connections, it is essentially fixing performance for those who were left behind during the transition to HTTP/2. These users are likely to be those who are in the most need of this performance increase. 
+
+Since COVID-19 many people have had to work from home. Many will be in rural areas where high-speed internet access either isn't available, or is expensive. In more populated area many users could still be on ageing cable / DSL connections with [high contention ratios](https://www.broadbandchoices.co.uk/guides/broadband/what-is-contention-ratio).
```
ChrisBAshton, line 84:
```suggestion
Since COVID-19, many people have had to work from home. Many will be in rural areas where high-speed internet access either isn't available, or is expensive. In more populated areas, many users could still be on ageing cable / DSL connections with [high contention ratios](https://www.broadbandchoices.co.uk/guides/broadband/what-is-contention-ratio). 
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Our Users
+
+Since HTTP/3 + QUIC offers better performance for users on unstable connections, it is essentially fixing performance for those who were left behind during the transition to HTTP/2. These users are likely to be those who are in the most need of this performance increase. 
+
+Since COVID-19 many people have had to work from home. Many will be in rural areas where high-speed internet access either isn't available, or is expensive. In more populated area many users could still be on ageing cable / DSL connections with [high contention ratios](https://www.broadbandchoices.co.uk/guides/broadband/what-is-contention-ratio). 
+
+The aim of this transition to HTTP/3 is to allow an ever more diverse level of user connection quality to access GOV.UK, essentially maximising access for all users, no matter what connection they use. By focusing on the long-tail of our users with poor connections, we can effectively shorten it and in doing so make the internet "good enough" to use in more situations. This is all without effecting performance for users already on a good internet connection.
```
ChrisBAshton, line 86:
```suggestion
The aim of this transition to HTTP/3 is to allow an ever more diverse level of user connection quality to access GOV.UK, essentially maximising access for all users, no matter what connection they use. By focusing on the long-tail of our users with poor connections, we can effectively shorten it and in doing so make the internet "good enough" to use in more situations. This is all without affecting performance for users already on a good internet connection.
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Browser support
+The support for HTTP/3 + QUIC in browsers currently isn't accurate on [Caniuse](https://caniuse.com/?search=http3). A PR to rectify this can be [seen here](https://github.com/Fyrd/caniuse/pull/5836). The browsers support for HTTP/3 at the time of writing is:
```
ChrisBAshton, line 89: I think the link text can be a bit more descriptive:

```suggestion
The support for HTTP/3 + QUIC in browsers [currently isn't accurate on Caniuse](https://caniuse.com/?search=http3), but there is a [PR to rectify this](https://github.com/Fyrd/caniuse/pull/5836). Browser support for HTTP/3 at the time of writing is:
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+- Chrome - Enabled
```
ChrisBAshton, line 91:
```suggestion
- Chrome - Enabled (see [Chromium blog](https://blog.chromium.org/2020/10/chrome-is-deploying-http3-and-ietf-quic.html))
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Browser support
+The support for HTTP/3 + QUIC in browsers currently isn't accurate on [Caniuse](https://caniuse.com/?search=http3). A PR to rectify this can be [seen here](https://github.com/Fyrd/caniuse/pull/5836). The browsers support for HTTP/3 at the time of writing is:
+
+- Chrome - Enabled 
+- Edge (chromium version) - Enabled
+- Firefox - Enabled in the next version (88)
+- Safari 14 - Disabled (currently behind a feature flag)
+
+Looking at the March 2021 analytics, approximately 55% of GOV.UK users are using a browser that can now take advantage of HTTP/3 + QUIC. Information about HTTP/3 and QUIC usage in Chrome can be [found here](https://blog.chromium.org/2020/10/chrome-is-deploying-http3-and-ietf-quic.html).
```
ChrisBAshton, line 96:
```suggestion
Looking at the March 2021 analytics, approximately 55% of GOV.UK users are using a browser that can now take advantage of HTTP/3 + QUIC.
```

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Other advantages to HTTP/3 + QUIC
+
+I've only covered a single advantage related to packet loss with web performance above, but HTTP/3 + QUIC comes with a number of other advantages:
+
+- Improvement in connection startup times via TLS v1.3 + [0-RTT](https://blog.cloudflare.com/introducing-0-rtt/)
+- Better security via end-to-end encryption at the transport layer
+- Establish an ecosystem that is easier to evolve from our current TCP usage
+
+Eventually our users will also be able to take advantage of [Connection Migration](https://tools.ietf.org/id/draft-tan-quic-connection-migration-00.html), where QUIC end-points will be able to migrate connections to different IP addresses and network paths at will. Although realistically, this is likely a number of years away.
```
ChrisBAshton, line 123: Could you explain what benefits that would bring? I know it's just an off the cuff comment, but I don't understand it, and it sounds interesting! I skimmed through the linked document but it talked more about the technical protocol rather than the benefits of the feature.

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
...
+![Comparing the visual filmstrip of the effect of packet loss on a page load](rfc-139/visual-filmstrip.png)
+
+As you can see, the visually complete metric sits at 9.5 seconds with no packet loss, but jumps to 13.5 seconds with 3% packet loss. This is a 42% increase.
+
+### Visual progress graph
+The visual progress graph clearly shows the effect packet loss has on page load times. As the packet loss increases, the page load slows down:
+
+![The visual progress graph comparing the percentage completeness of a page over time.](rfc-139/visual-progress.png)
+
+HTTP/3 + QUIC is designed to help improve the page load speed for users experiencing these connection conditions in the real world.
```
ChrisBAshton, line 70: These charts are great for proving that a problem exists. But I was a bit surprised that there's no HTTP/3 equivalent for the charts - it would be great to see what effect enabling HTTP/3 would have on these page load times. If the answer is "eliminate packet loss", and therefore you'd expect a consistent 9.5 seconds under these conditions, then it would be worth clarifying that here.

I do also wonder if there may be a surprising change in page load time in the switch from TCP to UDP - we may not have packet loss, but could some pages still load more slowly than they would over HTTP/2?

rfc-139-enable-http3.md
```diff
@@ -0,0 +1,135 @@
+# Enabling HTTP/3 on GOV.UK
+
+## Summary
+
+HTTP/3 is the latest version of the HTTP protocol that is currently in use in production by Facebook and Google, as well as many others. It offers web performance improvements over HTTP/2, specifically when connections with high packet loss are being used. Users on unstable mobile connections and rural areas should see the most improvement with this new protocol adoption.  
+
+## Problem
+
+According to [MLab](https://www.measurementlab.net/)'s latest data the UK has a median download speed of 20 MBit/s and 7.2 MBit/s upload speed, as seen in their [dashboard data from October 2020](https://datastudio.google.com/u/0/reporting/12XddYj7__cZOaBF2RwplImw_ZdQQ5Nuw/page/9GTNB?s=tA4mKm65BqY&params=%7B%22df5%22:%22include%25EE%2580%25800%25EE%2580%2580IN%25EE%2580%2580United%2520Kingdom%22%7D). Exporting this dashboard data to a spreadsheet paints a [very different picture](https://docs.google.com/spreadsheets/d/1UTRucwLY2VC4944-cKunL9yUPJkjDKXDdIPk6GY3610/edit#gid=0).
```
ChrisBAshton, line 9:
```suggestion
According to the [latest data (October 2020)](https://datastudio.google.com/u/0/reporting/12XddYj7__cZOaBF2RwplImw_ZdQQ5Nuw/page/9GTNB?s=tA4mKm65BqY&params=%7B%22df5%22:%22include%25EE%2580%25800%25EE%2580%2580IN%25EE%2580%2580United%2520Kingdom%22%7D) by [MLab](https://www.measurementlab.net/), the UK has a median download speed of 20 MBit/s and 7.2 MBit/s upload speed. However, exporting this dashboard data to a [spreadsheet](https://docs.google.com/spreadsheets/d/1UTRucwLY2VC4944-cKunL9yUPJkjDKXDdIPk6GY3610/edit#gid=0) paints a very different picture. 
```

ChrisBAshton (commented):
That was an interesting read - HTTP/3 sounds great, and the direction we should be heading in! 👏 

I can't see the proposal being at all contentious, but I do have a few suggestions to polish up the RFC:

rfc-139-enable-http3.md
Nooshu: Good idea, yes I will add a few bullet points to highlight this.

rfc-139-enable-http3.md
Nooshu: Yes I'd like to be able to conduct the same teat under HTTP/3 conditions. For this we'd need to ask Fastly to enable it on integration / staging first.

There is a question about how true the packet loss simulation is with WebPageTest (WPT), hence why it is a rough estimation. Packet loss tends to occur in bursts. So no PL for much of the connection, then a burst of 90% PL for a few seconds. I believe the way WPT models packet loss is that it takes a number of packets out of every 100, and loses them. e.g. lose 2 packets for every 100 sent for 2%. So the test above is great for showing the problem exists, but it's true representation and the fix HTTP/3 + QUIC could bring may be much harder to demonstrate. I think we will likely only see this in it's true form by looking at the web performance results for our "long-tail" users using real user monitoring (RUM) data.

But yes, it's a good question, and I'd like to see the results we get back.

rfc-139-enable-http3.md
Nooshu: Added info about connection migration and a link to a demo of it in action.

Nooshu:
Added updates to the RFC as Fastly has now made [HTTP/3 + QUIC available to all customers](https://www.fastly.com/blog/http-3-and-quic-are-now-available-for-our-entire-customer-base-at-no-additional-charge), it will just require a CNAME change since we aren't using dedicated Fastly IP's.

rfc-139-enable-http3.md
Nooshu: With RFC-147 implemented before HTTP/3 deployment on production, we should be able to compare H/2 vs H3 later in the year with SpeedCurve RUM data.

Nooshu:
It's now also worth noting that global HTTP/3 availability in browsers has now jumped to [73% according to CaniUse.com](https://caniuse.com/http3).

injms:
> It's now also worth noting that global HTTP/3 availability in browsers has now jumped to [73% according to CaniUse.com](https://caniuse.com/http3).

You can plug the list of browsers tracked by analytics into CanIUse - from the last 30 days HTTP/3 availability in browsers is just under 65%.

richardTowers:
We've started talking to Fastly about this. It turns out that we _are_ using dedicated IPs after all (apologies for spreading misinformation about that). That means their TLS team need to make a change ahead of time, and then we can choose to enable this on a per-service level by calling `h3.alt_scv()` in our VCL [as per Fastly's documentation](https://docs.fastly.com/en/guides/enabling-http3-for-fastly-services#offering-http3-from-your-fastly-service)

injms (approved):
I'm approving this as the deadline for discussion has past, any questions and comments have been resolved - and also because this looks to be a solid improvement.

Thanks for your hard work and investigation on this @Nooshu!

PR #141, opened 2021-04-08
Application Healthchecks

barrucadu:
**Deadline for discussion:** 2021-04-16

[Rendered](https://github.com/alphagov/govuk-rfcs/blob/msw/application-healthchecks/rfc-141-application-healthchecks.md)

---

We currently use application healthchecks for three things:

- For load balancing, to determine if an instance is ready to serve
  requests.
- For continuous deployments, to determine (along with other smoke
  tests) if a release can be automatically pushed to production.
- For alerting, so that 2ndline know if an instance has a problem
  which needs manual intervention to fix.

We implement healthchecks using the `GovukHealthcheck` module in
govuk_app_config.

Currently, a healthcheck response is always served with an HTTP status
of 200, and indicates in the body whether the instance's status is
"ok", "warning", or "critical".  However, AWS load balancers use the
HTTP status, not the response body, to determine whether an instance
is healthy.  So we will continue to send requests to an instance which
is in a "warning" or "critical" state.

This RFC proposes that we standardise on our healthchecks returning a
200 or a 500 status code, which has some further ramifications on
monitoring and alerting (discussed below).

A full list of healthchecks which will need updating is given at the
end of the document.

### DIFF
--- a/rfc-141-application-healthchecks.md
+++ b/rfc-141-application-healthchecks.md
@@ -1,0 +1,261 @@
+# Application Healthchecks
+
+## Summary
+
+We currently use application healthchecks for three things:
+
+- For load balancing, to determine if an instance is ready to serve
+  requests.
+- For continuous deployments, to determine (along with other smoke
+  tests) if a release can be automatically pushed to production.
+- For alerting, so that 2ndline know if an instance has a problem
+  which needs manual intervention to fix.
+
+We implement healthchecks using the [`GovukHealthcheck`][] module in
+[govuk_app_config][].
+
+Currently, a healthcheck response is always served with an HTTP status
+of 200, and indicates in the body whether the instance's status is
+"ok", "warning", or "critical".  However, AWS load balancers use the
+HTTP status, not the response body, to determine whether an instance
+is healthy.  So we will continue to send requests to an instance which
+is in a "warning" or "critical" state.
+
+This RFC proposes that we standardise on our healthchecks returning a
+200 or a 500 status code, which has some further ramifications on
+monitoring and alerting (discussed below).
+
+A full list of healthchecks which will need updating is given at the
+end of the document.
+
+## Problem
+
+We overload the meaning of "healthcheck", and cover both *application
+health* as well as *instance health* with our `/healthcheck`
+endpoints.  This introduces confusion, and means that we cannot
+currently use healthchecks in load balancing.
+
+Here are some examples of healthchecks which are suitable for load
+balancing and for post-deployment checks:
+
+- [`GovukHealthcheck::ActiveRecord`][]
+- [`GovukHealthcheck::Mongoid`][]
+- [`GovukHealthcheck::RailsCache`][]
+- [`GovukHealthcheck::Redis`][]
+- [`GovukHealthcheck::SidekiqRedis`][]
+
+If an app can't talk to a backing service it relies on, then it
+probably can't be trusted to handle any requests.  One of these
+healthchecks failing could indicate a networking or a configuration
+issue.
+
+Here are some examples of healthchecks which are unsuitable for load
+balancing or for post-deployment checks:
+
+- [`GovukHealthcheck::SidekiqQueueCheck`][], this is an abstract check
+  which other checks extend.
+  [`GovukHealthcheck::SidekiqQueueLatencyCheck`][] is an instance of
+  this check.
+- [`GovukHealthcheck::ThresholdCheck`][], this is another abstract
+  check which other checks extend.
+  [`GovukHealthcheck::SidekiqRetrySizeCheck`][] is an instance of this
+  check.
+- [`Healthcheck::ApiTokens`][] in [signon][], which reports that an
+  API token needs rotating.
+
+If one of these healthchecks fail, the app instances themselves are
+probably fine.  The problem is elsewhere.
+
+Two of these indicate capacity problems outside of the instance, and
+the third is an alert that a manual maintenance procedure needs to be
+performed some time in the next two months.  They all share the
+property that if one instance reports a failure, all will.  This makes
+them unsuitable for load balancing purposes.
+
+Even if we don't make our healthchecks suitable for load balancing, as
+GOV.UK moves towards continuous deployments, we will end up in
+situations where an automatic deployment is aborted because of
+something unrelated to the change being deployed.  That's not good.
+
+A healthcheck should check that the instance is running and isn't
+completely broken.  That's all.
+
+## Proposal
+
+I propose that we adopt these definitions:
+
+> A *liveness healthcheck* is an HTTP endpoint which MUST return an
+> HTTP status of 200.
+>
+> A *readiness healthcheck* is an HTTP endpoint which MUST return an
+> HTTP status of either 200 (if the instance may receive requests) or
+> 500 (if it should not).  It MAY also return details in the response
+> body as a JSON object.
+
+And furthermore that we commit to deprecating the `/healthcheck`
+endpoint and implementing the new healthchecks at endpoints
+`/healthcheck/live` and `/healthcheck/ready`.
+
+This new approach to readiness healthchecks is more suitable for our
+three purposes than before:
+
+- For load balancing, an unhealthy app will be taken out of the pool.
+- For continuous deployments, we can check the HTTP status rather than
+  read the response body.
+- For alerting, we can continue to read the response body.
+
+We get no immediate benefit from the liveness healthcheck, but will do
+when we have replatformed.
+
+Adopting these definitions gives us some migration work to do.  We
+can't start serving non-"ok" healthchecks with an HTTP status of 500
+right now, as some of our healthchecks are unsuitable for load
+balancing.
+
+We will need to implement the proposal in stages:
+
+1. List all the healthchecks which need changing, because they don't
+   indicate a critical failure of the app (done, see the appendix)
+2. For each such healthcheck:
+   - remove it if it's not adding value, or
+   - add a separate alert if it is
+3. For each application:
+   - serve a liveness healthcheck on `/healthcheck/live`
+   - serve a readiness healthcheck on `/healthcheck/ready`
+4. Update govuk-puppet and govuk-aws to use `/healthcheck/ready`
+   instead of `/healthcheck`
+5. Remove the `/healthcheck` endpoint from every app.
+6. Change `GovukHealthcheck` to serve the appropriate HTTP status
+   code.
+
+As part of (2), we will remove the "warning" state some of our
+healthchecks return.
+
+### Why separate healthchecks?
+
+Separate liveness and readiness healthchecks are a common best
+practice.  They are separate because they serve different purposes:
+
+- Liveness is used by the container orchestrator (such as [Amazon ECS][],
+  which we are replatforming to) to determine if an instance has
+  crashed or entered some other unrecoverable state, and must be
+  restarted.
+
+- Readiness is used by the network load balancers to determine if an
+  instance should be sent traffic.
+
+For example, let's say we have an application which uses a database,
+and that due to some transient fault (like a network partition) some
+of the instances cannot reach the database.  We do not want to send
+traffic to those instances, so their readiness healthchecks should
+fail.  But restarting the instances won't resolve the problem, as it's
+an issue with the underlying infrastructure, so we don't want their
+liveness healthchecks to fail.  We want the instances to keep running,
+so that when the transient fault recovers, they can quickly begin
+serving traffic again.
+
+On the other hand, let's say an instance exhausts all its memory, and
+can't handle any inbound requests at all.  The liveness (and
+readiness) healthcheck will fail, due to timing out, and ECS will
+restart the instance.
+
+### Remove the "warning" state
+
+There are four possible semantics we could assign the "warning" state:
+
+| # | State    | Allows automatic deployments? | Allows requests to be sent to the instance? |
+| - | -------- | ----------------------------- | ------------------------------------------- |
+|   | ok       | yes | yes |
+| 1 | warning  | yes | yes |
+| 2 | warning  | yes | no  |
+| 3 | warning  | no  | yes |
+| 4 | warning  | no  | no  |
+|   | critical | no  | no  |
+
+- If we go for option 1, then "warning" is the same as "ok".
+- If we go for option 2, then "warning" will let us deploy unusable releases.
+- If we go for option 3, then "warning" will block deployments.
+- If we go for option 4, then "warning" is the same as "critical".
+
+The most sensible option is 3, which is our current behaviour.
+
+But does it really gain us anything over having separate alerts for
+the specific condition we need to know about?
+
+By removing the "warning" state, we will remove some spurious alerts
+which don't add value, and add more specific alerts for ones which do.
+
+### Gaps in metrics
+
+Some of the unsuitable healthchecks correspond to conditions we need
+to alert about.  Some can be directly implemented as Icinga checks by
+drawing on data we already report to graphite, but others will need
+new metrics to be made available first.
+
+[Prometheus][], which we are adopting in the replatforming work, is a
+pull-based metrics gathering tool, which nicely solves the problem of
+how to get application state into an alert.  But we're not
+replatformed yet, so until then we may need to do something like
+email-alert-api, which [has a worker to push metrics to graphite][].
+
+## Appendix: Healthchecks which need changing
+
+The following healthchecks need updating before `GovukHealthcheck` can
+serve an HTTP status of 500:
+
+| App or Gem | Check | Reason |
+| ---------- | ----- | ------ |
+| [govuk_app_config][]  | [`GovukHealthcheck::SidekiqQueueCheck`][]           | checks for a capacity issue |
+| [govuk_app_config][]  | [`GovukHealthcheck::SidekiqQueueLatencyCheck`][]    | checks for a capacity issue |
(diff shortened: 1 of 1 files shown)


barrucadu:
@sengi @richardTowers before I drop this into #govuk-developers, does it look reasonable?

rfc-141-application-healthchecks.md
```diff
@@ -0,0 +1,225 @@
...
+
+If one of these healthchecks fail, the app instances themselves are
+probably fine.  The problem is elsewhere.
+
+Two of these indicate capacity problems outside of the instance, and
+the third is an alert that a manual maintenance procedure needs to be
+performed some time in the next two months.  They all share the
+property that if one instance reports a failure, all will.  And so if
+we use these healthchecks as part of load balancing rules, all running
+instances of the app will be taken down simultaneously.
```
sengi, line 74: This is true for the few remaining scary Classic ELB load balancers (and unfortunately I think there's at least one of these handling actual traffic). Well-designed load balancers like ALBs (which most of ours are) don't do this and will spray to all backends when all backends are simultaneously unhealthy, rather than giving up and just dropping traffic on the floor. Still bad though, because the healthcheck then becomes ineffective and an actually-failed backend will still get traffic.

I'd probably simplify this to something like

```suggestion
property that if one instance reports a failure, all will.  This makes
them unsuitable for load balancing purposes.```

(Or if you wanted to include all the detail, maybe "Depending on the load balancer, either all traffic will be dropped or the healthcheck will become ineffective and traffic will be routed even to instances which have actually failed.")

**Review by sengi — COMMENTED** [review]
Sounds very reasonable to me!

**Review by ChrisBAshton — COMMENTED** [review]
This sounds great to me, Michael. It's in line with the work we're already doing in Platform Health to improve our alerting:

- We have a card at the top of the backlog, to [move Signon's API tokens healthcheck into a separate alert](https://trello.com/c/Ouuf740D/2140-move-apitokens-signon-app-healthcheck-into-separate-alert)
- We've [removed the Sidekiq check from Content Data API](https://trello.com/c/Wn4w66dE/2431-2-simplify-content-data-api-healthcheck), and have a card to [consider what to do about Sidekiq checks more generally](https://trello.com/c/vdEbRKbn/2141-wip-replace-sidekiqredis-check), which you rightly identify as offering no value as a health metric.

Do you have a plan on how to follow through on this RFC after its approval? Perhaps we can team up and figure out how much of this Platform Health can action?

**Comment by barrucadu** [comment]
Thanks @ChrisBAshton , it's great that Platform Health is thinking about this problem, and for the same sorts of reasons, as well.

Currently I don't really have an implementation plan, beyond "do it all myself in my free time", but that might not lead to a timely implementation.  Teaming up to split up the work would be good.

**Comment by barrucadu** [comment]
I hope this won't be very contentious, so I've set the deadline for discussion to a week from now.

File: rfc-141-application-healthchecks.md [hunk 2]
```diff
@@ -0,0 +1,224 @@
...
+
+- If we go for option 1, then "warning" is the same as "ok".
+- If we go for option 2, then "warning" will let us deploy unusable releases.
+- If we go for option 3, then "warning" will block deployments.
+- If we go for option 4, then "warning" is the same as "critical".
+
+The most sensible option is 3, which is our current behaviour.
+
+But does it really gain us anything over having separate alerts for
+the specific condition we need to know about?
```
richardTowers, line 118: Is setting up separate alerts for these things part of this proposal?

If we just removed the warnings without implementing alerts, would that leave us with a problematic monitoring gap?

richardTowers (commented):
👍🏻  this looks sensible and well thought out to me. Happy to approve unless anyone else spots a problem in the next week.

rfc-141-application-healthchecks.md
barrucadu: It is, yes.  Step 2 of the migration path mentions making new alerts for checks which add value.

richardTowers: Ah yep, gotcha. Sorry I missed that.

rfc-141-application-healthchecks.md
```diff
@@ -0,0 +1,224 @@
...
+check any of its dependencies:
+
+```ruby
+get "/healthcheck", to: proc { [200, {}, []] }
+get "/healthcheck/ready", to: GovukHealthcheck.rack_response(...)
+```
+
+When we have replatformed to ECS, this healthcheck can be used to
+determine if a container needs to be automatically restarted, for
+example if the process has crashed.
```
bilbof, line 145: Thanks for considering ECS! The `/healthcheck/ready` endpoint would be _perfect_ for [ECS container health checks](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HealthCheck.html).

Am I correct in thinking that the endpoint `/healthcheck/ready` indicates liveness _and_ readiness, and that the `/healthcheck` endpoint indicates that an app can reach its dependencies?

Do these definitions make sense?

**Healthy**: The application can reach its critical dependencies (e.g. frontend can reach content-store)
**Ready**: The application has finished booting up (e.g. router has loaded routes into memory)

Some examples of health and liveness checks would be useful, to make the distinction between what should go into each health check clearer. It would be great to define how each health check should be used, too, to give us an idea about how to use them in ECS.

I am not 100% clear on the value of the "Healthy"/dependency checks (for use by a load balancer or alerting).

I would recommend removing the Healthy checks, and just keep the the readiness checks for load balancing. Because:

* Dependency health checks will mostly fail due to problems outside of the load balancer's control, and won't be a useful signal during deployments
* We don't plan on using health checks for alerting in ECS. Our intent in ECS is to [change how we do alerting](https://docs.google.com/document/d/1unWtbmh2g2HtXbCsFQ0GSbG1LIVraGsq6DQwX24mTVg/edit#heading=h.7qovbilab9jq) and avoid alerting on health checks. I would prefer that we discover a dependency is failing by letting it fail and alerting on it at a higher level (e.g. 5xx rate for a service), rather than via a health check.

Forgive me if this was already your intent here!

You wrote that good health checks indicate "that the instance is running and isn't completely broken". I reckon you've covered that well with the readiness health check. Whereas you wrote that bad health checks indicate "problems outside of the instance", which is probably pretty much what the "Healthy"/dependency checks are indicating.

I found dependency healthchecks a bit frustrating when trying to bring up apps in ECS, since containers would be stopped because a dependency wasn't running. Health checking dependencies (at any level of load balancing) prevented apps from degrading service gracefully (i.e. 50% of pages will work if half of the dependencies go down).

In the past I think we've actually been rescued during incidents from our health checks that check dependencies by AWS ALB's default behaviour. If all targets have failing health checks, then the health checks are ignored by the ALB - handy when a DB is down and you don't want to take down all the apps that use it, too!

> If all targets fail health checks at the same time in all enabled Availability Zones, the load balancer fails open. The effect of the fail open is to allow traffic to all targets in all enabled Availability Zones, regardless of their health status. [source](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html)

barrucadu: > Am I correct in thinking that the endpoint `/healthcheck/ready` indicates liveness and readiness, and that the `/healthcheck` endpoint indicates that an app can reach its dependencies?

I was actually thinking:

- `/healthcheck` *just* checks that the process is running and reachable over HTTP.  It's just a static route returning a 200, with no additional logic.
- `/healthcheck/ready` checks that the app is ready to serve requests (what that means depends on the app: eg, able to reach backing services, or has loaded routes into memories).

The `/healthcheck` is useful for container-level checks, because all the container runtime cares about is that the app runs.  So this is what ECS would use.

The `/healthcheck/ready` is useful for load balancer-level checks, because the load balancer cares, in addition, that the app is able to serve requests.  So this is what ALBs would use.

> I would prefer that we discover a dependency is failing by letting it fail and alerting on it at a higher level (e.g. 5xx rate for a service), rather than via a health check.

I like alerting on user-visible things, but what about the continuous deployment smoke testing use-case?  If a new version of an app has a bug which prevents it talking to its database, we'd want to catch that before it starts serving actual traffic - which a readiness healthcheck ensuring it can successfully do so gives us.

rfc-141-application-healthchecks.md
```diff
@@ -0,0 +1,224 @@
...
+requests, and takes it out of the load balancer target group if not,
+then it is a *readiness* check.
+
+There is also value in having a separate *liveness* check, which
+simply checks if the instance is running and reachable, and doesn't
+check any of its dependencies:
+
+```ruby
+get "/healthcheck", to: proc { [200, {}, []] }
+get "/healthcheck/ready", to: GovukHealthcheck.rack_response(...)
```
kevindew, line 140: I feel like having two things called healthcheck enters some confusing waters. Particularly if one of the routes isn't named - it could be that the term "healthcheck" interchangeably could refer to either of these and it wouldn't be clear that `/healthcheck` refers to a liveness check rather than a ready check.

I wonder is there anything we can do to avoid 2 healthchecks? Are two definitely needed? If they are could we have one thing that is a healthcheck and one thing that is named something else? Then the term healthcheck doesn't become too convoluted (for instance with this change GovukHealthcheck no longer has anything to do with a liveness check)

If that doesn't work could we maybe use /healthcheck as a prefix and any checks within it have specific name (e.g `/healthcheck/app-running` and `/healthcheck/ready`). That could help by there being nothing that can be described as "the healthcheck" and that it's always a collective.

rfc-141-application-healthchecks.md
bilbof: Thanks for clarifying! This sounds reasonable. I think this is fine provided the readiness health checks fail open, and the ALB checks are less aggressive about removing instances than the container checks, to avoid dependencies becoming 'hard' dependencies. In ECS we'll be using Envoy rather than ALBs for health checks, luckily they also fail open ([panic mode!](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/panic_threshold)) if all health checks fail.

rfc-141-application-healthchecks.md
```diff
@@ -0,0 +1,224 @@
...
+  API token needs rotating.
+
+If one of these healthchecks fail, the app instances themselves are
+probably fine.  The problem is elsewhere.
+
+Two of these indicate capacity problems outside of the instance, and
+the third is an alert that a manual maintenance procedure needs to be
+performed some time in the next two months.  They all share the
+property that if one instance reports a failure, all will.  This makes
+them unsuitable for load balancing purposes.
```
kevindew, line 73: It sounds quite reasonable to remove these, they seemed to make more sense when Icinga was pretty much the only consumer.

It does seem to leave a bit of a gap about how application state can be pulled for monitoring. I guess we were using them as a rather crude method for continuous pulling of state, the alternative approach (in Email Alert API we run a [Sidekiq worker to continuously push metrics to graphite](https://github.com/alphagov/email-alert-api/blob/master/app/workers/metrics_collection_worker.rb)) however seems a bit clunky. I suppose this is something Prometheus would help us with - with its focus on pulling metrics - so it might not be that we have a good alternative until then and it doesn't seem hugely pressing.

kevindew (commented):
Thanks for raising this @barrucadu it sounds well considered, logical and a good improvement 👍  I just have a few concerns around the introduction of a second healthcheck and the semantic consequences of that.

bar:
> Thanks for raising this @bar it sounds well considered, logical and a good improvement 👍 I just have a few concerns around the introduction of a second healthcheck and the semantic consequences of that.

@kevindew it seems you meant to tag another person?

rfc-141-application-healthchecks.md
barrucadu: That's a good point about the second healthcheck, @sengi what do you think?  Could we just have the readiness check (which returns either a 200 or a 500 status) and then the ECS healthchecks disregard the code and just check that the endpoint responds?

kevindew:
> > Thanks for raising this @bar it sounds well considered, logical and a good improvement 👍 I just have a few concerns around the introduction of a second healthcheck and the semantic consequences of that.
> 
> @kevindew it seems you meant to tag another person?

Oh gosh sorry, I failed to work the autocomplete widget. I've fixed it now.

rfc-141-application-healthchecks.md
barrucadu: It looks like the healthcheck is just a command: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HealthCheck.html

In which case, `[ "CMD-SHELL", "curl http://localhost/healthcheck" ]` will work.

I'll update the proposal to just have the one healthcheck.

rfc-141-application-healthchecks.md
```diff
@@ -0,0 +1,224 @@
...
+
+Even if we don't make our healthchecks suitable for load balancing, as
+GOV.UK moves towards continuous deployments, we will end up in
+situations where an automatic deployment is aborted because of
+something unrelated to the change being deployed.  That's not good.
+
+A healthcheck should check that the instance is running and isn't
+completely broken.  That's all.
+
+## Proposal
```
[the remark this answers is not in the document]
barrucadu: I've started this section with a new definition of a healthcheck

rfc-141-application-healthchecks.md
```diff
@@ -0,0 +1,224 @@
...
+determine if a container needs to be automatically restarted, for
+example if the process has crashed.
+
+In a sense we already have both liveness and readiness healthchecks:
+the liveness part comes from `GovukHealthcheck` always returning an
+HTTP status of 200, and the readiness part comes from reading the JSON
+response body.  But our load balancers cannot read the response body,
+so in practice we can't make use of our readiness check.
+
+### Migration path
```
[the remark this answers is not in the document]
barrucadu: Added a note about this

rfc-141-application-healthchecks.md
barrucadu: Added a note about this

rfc-141-application-healthchecks.md
barrucadu: Since there's no longer an ECS-specific part to this proposal, I've added an appendix explaining why we want slightly different healthchecks for ECS, and how we can turn a readiness healthcheck into a liveness healthcheck.

sengi: > Are two definitely needed?

Yes - separate liveness and readiness probes have been best practice for 5+ years now (basically since Kubernetes became popular - and they've been standard practice at Google for 10+ years). There are occasionally circumstances where you can get away with conflating them or omitting one or the other, but those are the exception rather than the rule - and we're talking about a general implementation across all GOV.UK apps here. Following best practices such as this will help to ensure we can run GOV.UK apps in containers reliably and without getting locked into to one particular system.

> I feel like having two things called healthcheck enters some confusing waters. Particularly if one of the routes isn't named - it could be that the term "healthcheck" interchangeably could refer to either of these and it wouldn't be clear that /healthcheck refers to a liveness check rather than a ready check.

I agree - /healthcheck would be a confusing choice of name for one of the healthchecks. I'd recommend deprecating the old endpoint and introducing new ones like you suggested. `/healthcheck/ready` is a good suggestion. At the risk of bikeshedding, I would probably go with `live` rather than `app-running` (so that it's clear that it's supposed to be used for the liveness probe). (If you look at examples of cloud-native open-source software you'll see liveness checks often called things like `/healthz` or `/health` but that's just a Google-ism that escaped into the outside world.)

On a practical note, I don't think ECS (or ALB, more precisely) is able to take 500 as a success response. Even if it could be made to do that, I think it would be surprising and confusing and lead to mistakes in future.

sengi: In principle you could reuse the readiness check as a liveness check in ECS, but again I think that would be highly confusing and there would be no benefit.

barrucadu: Ok, I've updated the proposal to define the two types of healthcheck, say we want to deprecate `/healthcheck` and set up `/healthcheck/live` + `/healthcheck/ready`, and give a motivating example of the sort of condition each check is supposed to detect.

barrucadu:
@kevindew @sengi @richardTowers @ChrisBAshton @bilbof The deadline for discussion is tomorrow, do you have any more comments?

rfc-141-application-healthchecks.md
```diff
@@ -0,0 +1,261 @@
...
+
+As part of (2), we will remove the "warning" state some of our
+healthchecks return.
+
+### Why separate healthchecks?
+
+Separate liveness and readiness healthchecks are a common best
+practice.  They are separate because they serve different purposes:
+
+- Liveness is used by the application runtime (such as [Amazon ECS][],
```
sengi, line 139: Slightly pedantic terminology fix for clarity (sorry to be picky but I think this one's worth it)

```suggestion
- Liveness is used by the container orchestrator (such as [Amazon ECS][],
```

(The specific component of the continainer orchestrator that actually does the liveness probe is typically the _node agent_ (kubelet, borglet, dockerd etc.) but I think that'd be an unnecessary level of detail here.)

sengi (approved):
Sounds great to me 👍  Thanks so much for putting this together - I'm so happy to see this long-neglected area getting some attention (and it'll help Replatforming too!)

kevindew (approved):
This sounds great, thanks @barrucadu 🏆 

When implementing this can we consider the impact of multiple types of healthcheck on the naming of `GovukHealthcheck`, naturally they were authored for a time of singular healthcheck and I don't know if we'll want to consider whether Ready is part of their name/hierarchy and/or having the code for the liveness check in govuk_app_config.

bilbof (approved):
Thanks @barrucadu 👍

PR #142, opened 2021-06-24
Add rfc 142 - interest cohort permission policy

bevanloon:
👉🏻 [Rendered](https://github.com/alphagov/govuk-rfcs/blob/rfc-142-add-interest-cohort-permission-policy/rfc-142-add-interest-cohort-permssion-policy-header.md)

Many thanks to [Matt Hobbs](https://github.com/Nooshu) for a discussion paper on the issue from which many of the words for this RFC have been lifted.

Deadline: 2021-07-01

### DIFF
--- a/rfc-142-add-interest-cohort-permssion-policy-header.md
+++ b/rfc-142-add-interest-cohort-permssion-policy-header.md
@@ -1,0 +1,52 @@
+# Add Interest cohort permission policy header
+
+## Summary
+
+Federated Learning of Cohorts is a new technology that exists in versions of Chrome 89+. Its purpose is to replace third party cookies used for tracking and targeting advertising by assigning Chrome users to groups (or cohorts) depending on their browser history. These cohorts can be used for targeted advertising by Google and other advertising networks.
+
+Once enabled, FLoC calculates a hash based on a user's browser history. This hash is synced with the Google servers and is used alongside other Chrome metrics and a FLoC ID is created. Chrome downloads a global FLoC data set to examine if the FLoC ID should be used for exposing ads to a specific user.
+
+## Problem
+
+In the short term the addition of FLoC actually increases the amount of information available to external observers, as FLoC can be used in conjunction with 3rd party cookies, as well as traditional browser fingerprinting techniques. When these techniques are used together, an observer's ability to identify an individual user increases. FLoC also circumvents some mitigations already put in place to limit data shared across domains. Browsers are rolling out partitioned storage, which restricts observers to only see information from their single domain. As the FLoC ID is generated from the browser history from a range of sites, FLoC actually results in the leaking of more data. This results in the observer seeing more information than they would without FLoC enabled.
+
+The FLoC specification specifically mentions that it “will never be able to prevent all misuse”, and the GOV.UK domain contains guidance and links to subjects that would be considered very sensitive for users, e.g.:
+
+- Universal credit login
+- Bankruptcy
+- Birth, marriage, deaths
+- Disability
+- Crime, justice and the law
+
+The FLoC specification does suggest that for sensitive cagerogies:
+
+> As a first mitigation, the browser should remove sensitive categories from its data collection. But this does not mean sensitive information can’t be leaked.
+
+However it does not guarantee sensitive information can't be leaked.
+
+To ensure information isn't leaked, an [explicit header can be set](https://github.com/WICG/floc#opting-out-of-computation).
+
+## Proposal
+
+To safeguard our users' sensitive browsing data, GOV.UK should explicitly set this header.
+
+We have 3 options for adding the header to the HTTP response: on the CDN, in the application, and in Nginx.
+
+- **Configure in CDN** - configure Fastly so it sends the header on each request.
+- **Configure in app** - configure the Rails apps to send the header.
+- **Configure in Nginx** - configure Nginx so it sends the header on each request.
+
+### Trade-offs
+
+| | Configure in CDN | Configure in app | Configure in Nginx |
+| --- | --- | --- | --- |
+| Deployment | The CDN is relatively easy and fast to deploy | Slow to roll out and iterate. We'd probably add it to `govuk_app_config`, which requires a version bump in multiple applications. Allows staged rollout. | Slow deployments via Puppet |
+| Policies | The header is set consistently for all of the requests, even ones that aren't served from a Rails app like [Licensing](https://github.com/alphagov/licensify) | Allows us to target per app and potentially per document type | Allows setting the header for the frontend apps and publisher apps |
+
+### Preferred option
+
+We'll configure the following header in NGINX to ensure we cover frontend apps as well as publishing apps across GOV.UK.
+
+```
+Permissions-Policy: interest-cohort=()
+```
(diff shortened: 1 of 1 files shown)


rfc-142-add-interest-cohort-permssion-policy-header.md
```diff
@@ -0,0 +1,52 @@
...
+### Trade-offs
+
+| | Configure in CDN | Configure in app | Configure in Nginx |
+| --- | --- | --- | --- |
+| Deployment | The CDN is relatively easy and fast to deploy | Slow to roll out and iterate. We'd probably add it to `govuk_app_config`, which requires a version bump in multiple applications. Allows staged rollout. | Slow deployments via Puppet |
+| Policies | The header is set consistently for all of the requests, even ones that aren't served from a Rails app like [Licensing](https://github.com/alphagov/licensify) | Allows us to target per app and potentially per document type | Allows sharing between the frontend apps and publisher apps |
+
+### Preferred option
+
+We'll configure the following header in the CDN to ensure consistency across GOV.UK
```
kevindew, line 48: Thanks for opening this as per our Slack chat this is just a quick comment, with a couple of food for thought things. 

Might want to call out that this excludes all the internal apps as they aren't behind the CDN. I'd imagine we'd also want them excluded, which may make Nginx an option that provides a higher chance of platform consistency (this would exclude www.gov.uk/alerts though but that is already setting it's own permissions policy).

Also the www.gov.uk/alerts usage gave me a thought, would we append to an existing permissions-policy an app responds with or overwrite it? Rails 6.1 has introduced a [framework for setting them](https://github.com/alphagov/content-publisher/blob/d9de0e3236ed4172a9715f34c13ab37d65eb5ff5/config/initializers/permissions_policy.rb) that none of our apps have embraced yet (to my knowledge) so may be a point to decide whether apps can/should set a permissions policy

bevanloon: Yeah that's a good point about publishing apps Kevin. Updated.

WRT Rails' framework for setting permission-policy, I agree it would be good to have a rule of thumb for GOV.UK, but I wonder whether this PR is the right place to set that guidance? Interested in your thoughts.

rfc-142-add-interest-cohort-permssion-policy-header.md
```diff
@@ -0,0 +1,52 @@
...
+
+The FLoC specification specifically mentions that it “will never be able to prevent all misuse”, and the GOV.UK domain contains guidance and links to subjects that would be considered very sensitive for users, e.g.:
+
+- Universal credit login
+- Bankruptcy
+- Birth, marriage, deaths
+- Disability
+- Crime, justice and the law
+
+The FLoC specification does suggest that for sensitive cagerogies:
```
sengi, line 21: typo
```suggestion
The FLoC specification does suggest that for sensitive categories:
```

rfc-142-add-interest-cohort-permssion-policy-header.md
```diff
@@ -0,0 +1,52 @@
...
+### Trade-offs
+
+| | Configure in CDN | Configure in app | Configure in Nginx |
+| --- | --- | --- | --- |
+| Deployment | The CDN is relatively easy and fast to deploy | Slow to roll out and iterate. We'd probably add it to `govuk_app_config`, which requires a version bump in multiple applications. Allows staged rollout. | Slow deployments via Puppet |
+| Policies | The header is set consistently for all of the requests, even ones that aren't served from a Rails app like [Licensing](https://github.com/alphagov/licensify) | Allows us to target per app and potentially per document type | Allows setting the header for the frontend apps and publisher apps |
+
+### Preferred option
+
+We'll configure the following header in NGINX to ensure we cover frontend apps as well as publishing apps across GOV.UK.
```
sengi, line 48: SGTM but we'll need to remember to add this header by some other means in Replatforming (as we're unlikely to have nginx in the new setup). There's a slight risk that this ends up being forgotten - could mitigate with monitoring and/or just doing our best to track it in the backlog.

sengi (commented):
I'm 100% behind this fwiw. Thanks for proactively looking out for our users' privacy! ❤️

rfc-142-add-interest-cohort-permssion-policy-header.md
sengi: Filed https://trello.com/c/cCIacIKn/387-ensure-were-still-setting-the-permissions-policy-interest-cohort-header-floc-opt-out to track this.

deborahchua:
This sounds really good to me! Do we know if other departments are going to do the same for their services? Specifically the ones that we link off to from GOV.UK.

huwd:
Sounds great, GOV.UK Accounts currently has some stuff on PaaS that might be worth considering (not sure if we're the only ones now, there was some covid work?).

It'd be good to consider, and perhaps document how we can de-FLoC any serices over there?
I understand our accounts stuff is behind cloudfront instead of Fastly, for instance, so we'd need to copy across and maintain multiple configs if that was the route.


Nginx sounds good, might need to check in on the setup on PaaS and who has access to that.
🤔 I wonder if this is something our PaaS colleagues could make default off for any deploy there?

I wonder if that's a solution to @deborahchua's question above?

rfc-142-add-interest-cohort-permssion-policy-header.md
bevanloon: Thanks Chris - good point. Thanks for putting together the trello card to track.

bevanloon:
@deborahchua 

> This sounds really good to me! Do we know if other departments are going to do the same for their services? Specifically the ones that we link off to from GOV.UK.

I think there are other conversations going on for other departments. There is definitely some work going in to updating the service manual around this. That will probably happen after we've added our header and will hopefully that will filter through to other depts and services as well.

rfc-142-add-interest-cohort-permssion-policy-header.md
kevindew: Sure. Yes I don't think much of the `Permission-Policy` header stuff belongs in this PR. The area I think should be covered by this proposal is what Nginx (or CDN if that's decided) should do if a proxied app returns the header.  Options seem to be: assume app knows better do nothing; append to app's header to merge or overwrite the header. All seem to have pro's and con's, I'd maybe lean towards the first option myself for lowest deviation between dev and prod.

bevanloon: I'd be in favour of the simplest thing we can get away with. If we just  think we should just rely on `add_header` at Nginx without imposing any more logic at the moment. As I understand it, that would essentially append to an app's header values.

I think this would give us a reasonable starting point. I suspect if we make use of the Permission-Policy header at an app level it will be for things other than FLoC. In those cases, we at least wouldn't run the risk of removing the explicit opt-out. If we do bump into a situation where we want to enable FLoC for specific apps, we can worry about the logic then.

bevanloon:
> Sounds great, GOV.UK Accounts currently has some stuff on PaaS that might be worth considering (not sure if we're the only ones now, there was some covid work?).

Yes good point @huwd  thanks.

> It'd be good to consider, and perhaps document how we can de-FLoC any serices over there?
> I understand our accounts stuff is behind cloudfront instead of Fastly, for instance, so we'd need to copy across and maintain multiple configs if that was the route.

Setting headers at CloudFront is a bit annoying, so I've amended the PR to say that we'll set headers at the app layer for our PaaS hosted applications. How does that sound?

himal-mandalia (approved):

richardTowers (approved):
Thanks for this @bevanloon and @Nooshu, detailed and well reasoned.

If we need to use the `Permissions-Policy` header for other things in the future we can cross that bridge when we come to it.

I'm happy with this approach.

PR #143, opened 2021-11-19
RFC 143: split our databases out into individual instances

spikeheap:
Adds a proposal to move away from our megadatabases to a single-database-per-instance approach.

I'd particularly appreciate thoughts and wording refinements around the main principle. The intention is for it to unambiguously say "one database per instance, even if your app has 2x different Postgres databases" , except where there's a strong reason to break the rule.

We'll take a couple of weeks to discuss, and decide the fate of this RFC on Monday 6th December.


rfc-142-split-database-instances.md
```diff
@@ -0,0 +1,50 @@
+# One database per RDS instance
+
+## Summary
+
+We currently use a one-instance-many-databases approach to hosting MySQL, Postgres, Mongo and Redis. These mega-databases make it difficult to perform major version upgrades – they need to be coordinated across all applications at the same time, increase the blast-radius of problems in a single database, and make it harder to appropriately size and monitor resources for an application.
+
+This RFC proposes that we split the single RDS instances for MySQL and Postgres, moving to a single RDS instance per database. It also proposes we apply the same principle to our self-hosted Mongo and Redis installations.
+
+This principle _could_ be applied to RabbitMQ, S3, etc., but it's out of scope for this RFC.
```
kevindew, line 9:
```suggestion
This RFC proposes that when applications depend on a supportive service that is not used by other applications, such as a database, the supportive service is distinct from those used by other applications. It is applied particularly to RDS instances, but the principle can apply to other supporting processes such as Redis, Elasticsearch, S3.
```

A suggestion here to try make this a bit less database specific and more of an all-rounder principle.

rfc-142-split-database-instances.md
```diff
@@ -0,0 +1,50 @@
...
+
+2. Problems in one database are likely to impact other unrelated services. For example an app with unexpectedly high load on its database could max-out the instance, causing performance degradation in other databases and their apps.
+
+3. Related to [2], it's significantly harder to size instance resources. Disk is fairly straightforward, but it's harder to understand CPU/RAM/IOPS patterns. We're also likely to over-size resources to reduce the likelihood large-scale problems in [2].
+
+4. There is no clear owner of the instances. Each app is owned by a single team, and each database should be accessed by exactly one app. However many apps with different owners reside on the single instance. Sizing and managing databases is an app-level responsibility, so it doesn't fit with the remit of Platform Reliability/Replatforming/Platform Health. Having multiple databases on a single instance
```
kevindew, line 25: Looks like you've got a premature period at the end of of this: `s/Platform Health. Having/Platform Health having/`

rfc-142-split-database-instances.md
```diff
@@ -0,0 +1,50 @@
...
+## Proposal
+
+We propose that:
+
+1. All new long-lived databases are created in their own individual RDS instances.
+2. Existing MySQL and Postgres databases are migrated out of the central instance to individual managed-service instances as part of the upcoming major upgrade cycle.
+3. Existing Redis and Mongo databases are migrated out of their central instances in-line with team priorities, and prior to any major upgrade.
+
+
+### Consequences
```
kevindew, line 38: There's perhaps a few other consequences to consider:

- there will be less consistency in DB versions across the platform, individual teams could be on different versions
- easier to identify which application has performance problems and an expected reduction in cascading issues
- a potential increase in cost, hopefully offset by having to crank up the machine to huge sizes when there's a problem

rfc-142-split-database-instances.md
```diff
@@ -0,0 +1,50 @@
...
+
+- Creating more RDS instances means that application-owning teams will need to monitor and consider database upgrades as part of their day-to-day work. We can ease this burden through monitoring & reporting, which may be considered by a future Platform Reliability team.
+
+- Zero-downtime migrations take extra planning and effort. Services such as AWS DB Migration Service ease some of this complexity.
```
kevindew, line 42: This one seems a little out of the blue - perhaps it's: There may need to be some service unavailability to apply these principles - however this would happen regardless by AWS ending support (and AWS DB Migration service can help) ?

rfc-142-split-database-instances.md
```diff
@@ -0,0 +1,50 @@
...
+
+- This change affects our production, integration, staging and CI environments. Our setup for CI requires a slightly different user setup so we can create and destroy databases within the instance (where we create databases per test run).
+
+## Questions
+
+1. Should we exclude CI from this proposal, and use a separate RFC to cover changes there?
```
kevindew, line 48: I don't think we need to say a lot about CI so I don't think it needs a separate RFC. I think there should be a distinct consequence for it thought: "CI will need to support a wider range of database versions than it currently supports to cope with teams using different versions"

On the CI front we've got past experience of having different DB versions (we use tags - so right now CI Agent 1-3 have Mongo 2.4 and CI Agent 4 has Mongo 3.2 - best not to comment on the age of those versions mind) so the two things that might be new are: trying to run newer versions of databases and potentially having a reasonable pool of DB options

rfc-142-split-database-instances.md
```diff
@@ -0,0 +1,50 @@
...
+
+2. Do we need another RFC for to discuss the use of managed services (e.g. RDS) in preference to self-hosted?
```
kevindew, line 50: I'd suggest sneaking it in here in a couple of places rather than a separate one, as just a minor tangent - since that is largely a decision that was made a long time ago and, if it needs to be restated, may be something replatforming want to define.

I'd probably add an item to the problems: We still host some database features ourselves (such as Mongo 2.6) which allows us to run very out of date versions of software.

And then in proposal add a point 4 of: 

```
4. Any future supporting services for applications should be distinct to the application and, where applicable, run on a managed service
```

kevindew (commented):
This is looking great. I've chucked a few suggestions in - I may be fighting the tide a bit by trying to keep some broadnesss to wider supporting services, no biggie if they don't make it.

rfc-142-split-database-instances.md
```diff
@@ -0,0 +1,50 @@
...
+
+4. There is no clear owner of the instances. Each app is owned by a single team, and each database should be accessed by exactly one app. However many apps with different owners reside on the single instance. Sizing and managing databases is an app-level responsibility, so it doesn't fit with the remit of Platform Reliability/Replatforming/Platform Health. Having multiple databases on a single instance 
+
+This is coming up now because our versions of MySQL and Postgres are approaching End of Life (notice for [MySQL](https://forums.aws.amazon.com/ann.jspa?annID=8790) and [Postgres](https://forums.aws.amazon.com/ann.jspa?annID=8499)), forcing us to plan major upgrades for both over the next few months.
+
+## Proposal
+
+We propose that:
+
+1. All new long-lived databases are created in their own individual RDS instances.
```
brucebolt, line 33: Could we make this slightly more vague and just refer to `instances` rather than `RDS instances`, to allow individual teams to explore non-RDS options should they wish to?

rfc-142-split-database-instances.md
spikeheap: Thanks, fixed. I had an half-finished sentence dangling there too

rfc-142-split-database-instances.md
spikeheap: Added 👍

rfc-142-split-database-instances.md
spikeheap: I've worked this into consequences 🙇

rfc-142-split-database-instances.md
spikeheap: I've had a go at this. Let me know if we want more (or less) subtlety :D

rfc-142-split-database-instances.md
spikeheap: I've reworded to make the point that we'd have to upgrade anyway. I thought it was worth calling out because the additional complexity of splitting the databases could increase downtime (or at least force people to think about it), but happy to remove if the reworded one doesn't fit with the rest of the RFC

rfc-142-split-database-instances.md
spikeheap: That's a good question @brucebolt. Do you have something in mind, or is this just a concern that we're being overly prescriptive here?

I'm not tied to "RDS" specifically, but there could be a readability problem as we make it more generic. For example, it's easy enough to grok "Split these three MySQL databases into 3 Aurora instances", but as we broaden out to (potentially) all supporting services we're reduced to the RFC saying "distinct supporting services in separate instances". "Instances" has a really specific meaning to me now I've looked at this RFC for a while, but I wonder if it's intuitive.

I'd be tempted to err on the side of specificity if we consistently use AWS services at the moment, with the caveat that teams should always feel able to explore other options if the "standard" way doesn't meet their needs... but if things aren't that consistent then :+1: to removing this. 

📣 @kevindew this ties to your point about broadening the RFC out

rfc-143-split-database-instances.md
```diff
@@ -0,0 +1,66 @@
...
+## Summary
+
+We currently use a one-instance-many-databases approach to hosting MySQL, Postgres, Mongo and Redis. These mega-databases make it difficult to perform major version upgrades – they need to be coordinated across all applications at the same time, increase the blast-radius of problems in a single database, and make it harder to appropriately size and monitor resources for an application.
+
+This RFC proposes that we split the single RDS instances for MySQL and Postgres, moving to a single RDS instance per database. It also proposes we apply the same principle to our self-hosted Mongo and Redis installations, and more generally to all supporting services for an application which aren't shared with other applications.
+
+
+## Naming
+
+The word "database" is often interchangeably used to describe a collection of tables, and an instance of (e.g.) Postgres server running somewhere. For the purposes of this RFC, "database" refers to the collection of tables, schema, etc.. We'll use "instance" to refer to the runtime that hosts the database.
```
richardTowers, line 12: Maybe:

```suggestion
The word "database" is often interchangeably used to describe a collection of tables, and an instance of (e.g.) Postgres server running somewhere. For the purposes of this RFC, "database" refers to the collection of tables, schema, etc.. We'll use "instance" to refer to the runtime that hosts one or more databases.
```

rfc-143-split-database-instances.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Problem
+
+Having a single instance hosting multiple databases has a number of drawbacks:
+
+1. Major upgrades must be applied to all databases in one go, as it's the instance we're upgrading, not the database. Major upgrades are considered risky because they contain potentially breaking, and backward-incompatible changes. This means major upgrades are scarier, so we're less likely to do them until we're forced. They also require coordination across disparate teams so everyone's confident their apps will survive the upgrade.
+
+2. Problems in one database are likely to impact other unrelated services. For example an app with unexpectedly high load on its database could max-out the instance, causing performance degradation in other databases and their apps.
```
sihugh, line 20: [It has happened fairly recently and is also difficult to debug](https://docs.google.com/document/d/10aOHyjO8JjzbIhj5HpowAuDlXvnbkAnDgYpTuCpBv-0/edit#). If all apps are having problems and are being slow, which one is it that's causing the problem? 

Do we want to remove the ambiguity and so replace the last sentence with:

"For example an app with a poorly performing database has caused performance degradation in other databases and their apps."

ChrisBAshton: +1 - would be good to link to the incident report in that sentence too.

rfc-143-split-database-instances.md
```diff
@@ -0,0 +1,66 @@
...
+
+This is coming up now because our versions of MySQL and Postgres are approaching End of Life (notice for [MySQL](https://forums.aws.amazon.com/ann.jspa?annID=8790) and [Postgres](https://forums.aws.amazon.com/ann.jspa?annID=8499)), forcing us to plan major upgrades for both over the next few months.
+
+## Proposal
+
+We propose a new principle to be applied across GOV.UK:
+
+1. Supporting services for an application should not be shared with any other applications at the infrastructure/provisioning level.
+
+> Note: this principle doesn't cover services used for application communication, e.g. where SNS/SQS/RabbitMQ broadcasts events for other applications. The use of the service carries most weight here – e.g. a fictional RabbitMQ queue used solely to manage Sidekiq jobs would be considered a "supporting service", whereas a RabbitMQ queue to broadcast publishing domain events would be accessed by many apps. Both these queues would live in distinct RabbitMQ instances/servers.
```
ChrisBAshton, line 36: This was a bit wordy and doesn't actually affect the outcome (still needs to be on its own distinct server), so I've proposed this change:

```suggestion
> Supporting services for an application should not be shared with any other applications at the infrastructure/provisioning level.

For example, a RabbitMQ queue used solely to manage Sidekiq jobs for an application should never be accessed by another app. An exception to the rule would be when services are designed for application communication, such as a RabbitMQ queue that broadcasts publishing events to be consumed by other apps. Both types of queues should live in distinct RabbitMQ instances/servers.
```

I've also made the principle the 'callout', rather than the side note, as it didn't make much sense as a 1-item numbered list.

rfc-143-split-database-instances.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Migrating to new instances with minimal (or no) downtime takes planning and effort. AWS' forced upgrades of Postgres and MySQL in early 2022 will force this effort regardless.
+
+- This change affects our production, integration, staging and CI environments. Our setup for CI requires a slightly different user setup so we can create and destroy databases within the instance (where we create databases per test run).
+
+- CI will need to support a wider range of database versions than it currently supports to cope with teams using different versions. Ideally, teams will be able to create and manage databases to meet their needs independently, but as a first step we'll need to introduce a pool of "supported" database types and versions. We have experience of doing this with Mongo and tagged agents.
+
+## Questions
+
+1. Can we be specific about databases. If we increase scope to all supporting services, does this RFC become less actionable now?
```
ChrisBAshton, line 66: It adds a bit of complexity, yes - I wouldn't be against removing supporting services from the scope, and introducing that as a follow-up RFC once the DBs are migrated.

ChrisBAshton (commented):
Thanks for putting this together! Definitely on-board with a separate DB per application.

rfc-142-split-database-instances.md
brucebolt: I think my concern is just about this being too prescriptive.  In future, teams may look back at this and think we'd specified RDS for a reason that we didn't note, but could've been important.  But I'm happy for RDS to remain specified, as that's what we're currently using.

rfc-143-split-database-instances.md
```diff
@@ -0,0 +1,66 @@
...
+
+2. Problems in one database are likely to impact other unrelated services. For example an app with unexpectedly high load on its database could max-out the instance, causing performance degradation in other databases and their apps.
+
+3. Related to [2], it's significantly harder to size instance resources. Disk is fairly straightforward, but it's harder to understand CPU/RAM/IOPS patterns. We're also likely to over-size resources to reduce the likelihood large-scale problems in [2].
+
+4. There is no clear owner of the instances. Each app is owned by a single team, and each database should be accessed by exactly one app. However, many apps with different owners reside on the single instance. Sizing and managing databases is an app-level responsibility, so it doesn't fit with the remit of Platform Reliability/Replatforming/Platform Health (i.e. for a single, central resource).
+
+5. We still host some database features ourselves (such as Mongo 2.6) which has a maintenance burden and allows us to run very out of date versions of these services.
```
barrucadu, line 26: Nitpick, but I don't think this is a problem due to "Having a single instance hosting multiple databases"

sengi (commented):
I think the essential idea is a good one. I would highly recommend:

1. Proposing specific changes: which databases are the most important to split out? I wouldn't be at all surprised if you can get 80% of the benefit from 20% of the work if you pick the right ones. I'd probably use two criteria: which ones are noisy neighbours and which ones now belong to different teams who want to manage their instances themselves.
2. Gathering data to support those specific proposed changes. E.g.:
    * Which apps have caused noisy-neighbour incidents (i.e. resource isolation problems) in, say, the past year — and how often.
    * The expected increase in running costs from the specific changes being proposed etc.
    * Roughly the expected effort involved in splitting out these databases, and how this varies depending on the extent to which they're split up (e.g. just the noisy ones vs everything).
3. Fully explore the drawbacks of taking this approach too far
    * E.g. resource stranding is kinda hinted at but isn't fully explained
    * Also increased operational overheads, presumably? (A couple of extra RDS clusters isn't going to move the needle, but 20 would probably give pause for thought.)

I think it's reasonable to propose a general rule-of-thumb of avoiding sharing databases across team boundaries and system boundaries, but I would caution against overgeneralisation. There will always be exceptions where the right choice for the organisation needs to be made on the merits of the individual case rather than mechanically by rule-of-thumb. "If in doubt, keep your databases separate from other apps' databases and especially other teams' databases. It's ok to make exceptions to this if you have supporting data and sound rationale for doing so." is probably about as general as I'd recommend.

rfc-143-split-database-instances.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Creating more RDS instances means that application-owning teams will need to monitor and consider database upgrades as part of their day-to-day work. We can ease this burden through monitoring & reporting, which may be considered by a future Platform Reliability team.
+
+- There will be less consistency in database versions across the programme. Individual teams are responsible for their upgrade appetite and prioritisation.
+
+- There is a potential increase in costs caused by having dedicated instances, particularly when they are initially split (because it's harded to accurately estimate resource requirements). This may be offset by being able to optimise individual databases to their needs.
+
+- Migrating to new instances with minimal (or no) downtime takes planning and effort. AWS' forced upgrades of Postgres and MySQL in early 2022 will force this effort regardless.
+
+- This change affects our production, integration, staging and CI environments. Our setup for CI requires a slightly different user setup so we can create and destroy databases within the instance (where we create databases per test run).
```
kerin, line 60: There's likely a strong argument for using local disposable databases in non-production-like environments like integration or CI, to skip the long RDS provisioning times - the move to kubernetes should in principle make this pretty straightforward.

rfc-143-split-database-instances.md
```diff
@@ -0,0 +1,66 @@
...
+
+- It will be easier to identify the causes of performance problems in an application.
+
+- The chance of a problem with one service cascading to other services will be reduced, because there is no contention for resources which aren't shared.
+
+- Creating more RDS instances means that application-owning teams will need to monitor and consider database upgrades as part of their day-to-day work. We can ease this burden through monitoring & reporting, which may be considered by a future Platform Reliability team.
+
+- There will be less consistency in database versions across the programme. Individual teams are responsible for their upgrade appetite and prioritisation.
+
+- There is a potential increase in costs caused by having dedicated instances, particularly when they are initially split (because it's harded to accurately estimate resource requirements). This may be offset by being able to optimise individual databases to their needs.
```
bilbof, line 56: Do we have a rough idea of what this isolation will cost?

The consequences of the principle (redis per app, memcached per app, RDS per app, etc) may lead to a very large increase in AWS costs and underutilisation of AWS resources.

sengi:
As example of the sort of thing I mean by gathering supporting data, here are the sizes of the databases in the shared Postgres and MySQL RDSs in prod:

```
mysql> SELECT table_schema, ROUND(SUM(data_length)/1024/1024) AS data_mb, ROUND(SUM(index_length)/1024/1024) AS indices_mb FROM information_schema.tables GROUP BY table_schema ORDER BY data_mb DESC;
+----------------------------------+---------+------------+
| table_schema                     | data_mb | indices_mb |
+----------------------------------+---------+------------+
| whitehall_production             |   28011 |       2968 |
| signon_production                |    3422 |       2973 |
| contacts_production              |      17 |          1 |
| release_production               |      13 |          8 |
| collections_publisher_production |       6 |          1 |
| mysql                            |       4 |          2 |
| search_admin_production          |       1 |          0 |
| sys                              |       0 |          0 |
| performance_schema               |       0 |          0 |
| information_schema               |       0 |          0 |
+----------------------------------+---------+------------+
10 rows in set (0.04 sec)
```

```
postgres=> SELECT pg_database.datname AS "database_name", pg_database_size(pg_database.datname)/1024/1024 AS size_in_mb FROM pg_database ORDER BY size_in_mb DESC;
            database_name            | size_in_mb 
-------------------------------------+------------
 email-alert-api_production          |     354454
 publishing_api_production           |      98037
 ckan_production                     |      65263
 support_contacts_production         |      20641
 link_checker_api_production         |       1651
 local-links-manager_production      |        322
 account-api_production              |        136
 content_tagger_production           |         87
 content_publisher_production        |         81
 service-manual-publisher_production |         44
 content_data_admin_production       |          9
 template0                           |          7
 content_audit_tool_production       |          7
 template1                           |          7
 ckan_pycsw_production               |          7
 rdsadmin                            |          7
 postgres                            |          7
(17 rows)
```

There are quite a few tiny ones and I expect those ones have very light workloads as well. That's likely to lead to a lot of stranded resources if we were to split out literally every database into its own RDS instance, especially where replicas are needed for availability. It's worth trying to quantify that to get some clarity on how far it makes sense to go in order to get the most benefit while minimising the drawbacks.

rfc-143-split-database-instances.md
spikeheap: Good point, I'll remove this one

rfc-143-split-database-instances.md
spikeheap: This was such a good point I've lifted it verbatim into the RFC.

rfc-143-split-database-instances.md
spikeheap: 🙌 this question was left over from an initial draft. I _think_ we've got a reasonable balance that lets us be specific about databases while also saying it applies to supporting services now.

rfc-143-split-database-instances.md
```diff
@@ -57,10 +53,12 @@ Applying this initially to our databases, we propose that:
...
 
 - CI will need to support a wider range of database versions than it currently supports to cope with teams using different versions. Ideally, teams will be able to create and manage databases to meet their needs independently, but as a first step we'll need to introduce a pool of "supported" database types and versions. We have experience of doing this with Mongo and tagged agents.
 
 ## Questions
 
-1. Can we be specific about databases. If we increase scope to all supporting services, does this RFC become less actionable now?
\ No newline at end of file
+1. Can we be specific about databases. If we increase scope to all supporting services, does this RFC become less actionable now?
```
sengi, line 62: I would highly recommend limiting the scope to just databases and proposing a set of specific, measurable changes and quantifying the benefits where possible.

rfc-142-split-database-instances.md
kerin: Is the general point here to move away from shared persistent storage, whether it be an RDS instance, DBs on VMs or S3 buckets? So should the wording here be something more like "persistent storage should not be shared across application boundaries, e.g. separate RDS instances, S3 buckets, elasticache instances, etc"?

spikeheap: I think that's a valid point too @kerin. Here I'm referring to long-lived compute as well as storage – the wording is a bit awkward but I was alluding to CI, where we might have long-lived instances which house short-lived databases that are created & destroyed in a test cycle.

rfc-143-split-database-instances.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Problem
+
+Having a single instance hosting multiple databases has a number of drawbacks:
+
+1. Major upgrades must be applied to all databases in one go, as it's the instance we're upgrading, not the database. Major upgrades are considered risky because they contain potentially breaking, and backward-incompatible changes. This means major upgrades are scarier, so we're less likely to do them until we're forced. They also require coordination across disparate teams so everyone's confident their apps will survive the upgrade.
+
+2. Problems in one database are likely to impact other unrelated services. For example an app with a poorly performing database has caused performance degradation in other databases and their apps[^1].
+
+3. Related to [2], it's significantly harder to size instance resources. Disk is fairly straightforward, but it's harder to understand CPU/RAM/IOPS patterns. We're also likely to over-size resources to reduce the likelihood large-scale problems in [2].
```
sengi, line 21: This is certainly true for databases beyond a certain (large-ish) size and/or workload — and it's an important consideration. The converse is true at the other end of the spectrum, though: we have a quite a few very small / lightly-loaded databases which would be very hard to size efficiently if they all ran on separate RDS instances and you'd end up with poor utilisation (which means poor cost-effectiveness).

spikeheap:
Thanks for your comments & suggestions :bow:.

@sengi & @bilbof: 🙌 thanks, these are good points to consider. I've written a wall of text that hopefully comes across as "I'm thinking about this" rather than "here's a wall of text to put you off" :D. Follow-up comments welcome!

### Operational costs 

:+1: to attempting to quantify the breakdown of existing resource requirements, however I'm struggling to think of a way we can accurately predict load on each database once it's split out. Do you have any suggestions?

It's not safe to assume our operational costs will increase. The estimated baseline costs for our 2 shared production multi-AZ instances are:

- Postgres `db.m5.12xlarge`, 4500Gb: $9300/month
- MySQL `db.m5.xlarge`, 100Gb: $600/month

These are sitting at around 0% CPU utilisation for Postgres (with a single spike in 12 hours up to 50%), and <40% for MySQL (with a burst to 100% around 4am). That's probably a crass oversimplification, and there may be less frequent spikes, but it hints heavily towards massive overprovisioning to reduce the chance of resource contention.

As a comparison, for a tiny, relatively unused database we _could_ be looking at $33/month[^1] for a multi-AZ `t2.micro` instance. This is an extreme case, but even several multi-AZ `m4.large` at $360/month wouldn't be increasing our bill significantly. In MoJ we found that teams were able to have a higher tolerance to under-resourcing – trivially identifing database performance issues and increasing resources as needed, without risking impacting other services (or portions of a service).

A positive side-effect of this split is also that if one application continues to demand a `db.m5.12xlarge` instance for a CPU spike once a day, we'll likely be motivated to reduce the impact of that spike.

We discussed the increase in operational costs in the senior technologists call, and agreed we're happy to accept the risk of increased operational costs for the benefit of resource isolation.

### Taking things to the extreme

I'm not a fan of the idea of applying this principle only to the riskiest/noisiest/most expensive databases – in fact this RFC is advocating for taking this approach as far as we can. Although resources _could_ become stranded on old versions, the premise of splitting things up is to reduce the scariness of applying upgrades, which in turn makes upgrades more likely to happen.

At MoJ we (almost) exclusively followed the principle of one-database-one-instance for our multi-tenant cloud infrastructure. This applied to both monolithic legacy apps _and_ our emerging microservices architecture. The approach was successful enough that we started retrofitting that approach for our giant, legacy offender management system which was sharing databases (this was originally done to limit Oracle license costs).

I'd like to explore the risk of overgeneralisation. Aside from cost, do we have any examples (theoretical is fine) where applying this as a strictly one-database-one-RDS-instance rule would cause problems, and what those might be?

### Caveats

> It's ok to make exceptions to this if you have supporting data and sound rationale for doing so.

As with any RFC, there's always the opportunity for exceptions. I'd absolutely support subsequent discussions & RFCs that refine (or overturn) this one when we've got new use cases or feel the need to document exceptions. Is that how we've approached RFC's in the past?

🙇 Thanks again for taking the time to talk through this. If it's easier to jump on a call to chat, ping me on Slack 🙂. 

[^1]: All costs are based on the AWS costs calulator, and don't take into account agreed discounts/reserved instances, etc. Take with a pinch of salt...

sengi:
Thanks for the detailed response! Might indeed be easier to chat on a higher-bandwidth medium about some of this :)

It sounds like we're very much on the same page about what we're trying to achieve here — I just have some concerns that things may not turn out how we want if we're not careful.

The over-provisioning point is an important one, but I think we need to be careful that we're making a like-for-like comparison. There is undoubtedly some right-sizing that needs to happen for some of these instances, but that's (mostly) orthogonal to the question of splitting up the databases for isolation reasons. If we compare a rightsized, everything-separate scenario to a non-rightsized monolithic scenario, we're conflating two different changes and not comparing like-for-like. Since it's going to be very hard to predict the post-rightsizing sizes (since it'd involve some trial and error to find out what really works reliably in production), I'd recommend considering the two issues separately for cost estimation purposes.

Let's suppose that we no longer need the 12xlarge to cope with the email-alert-api workload (which may be the case since there have been some improvements to the app and we're probably not sending out enormous travel advice mailshots at quite the rate we were a year ago!) In order to fairly attribute the savings from rightsizing it, we'd need to show that its large size is still preventing noisy-neighbour problems and therefore it's only by splitting up the databases that we'd be able to downsize it.

As an aside, there are a bunch of other low-hanging-fruit cost savings to be had e.g. getting rid of unused or low-value replicas (particularly in the non-prod environments), reducing EBS volume sizes and switching from `gp2` to `gp3`, etc. but those savings also shouldn't be counted for the purpose of deciding on how much to split up.

A thought experiment: imagine that RDS had really good resource isolation between users. Would that change how / how much you'd choose to split things up?

I think we may have crossed wires about resource stranding — this refers to the wasted space (CPU/RAM/disk etc.) when partitioning resources (and conversely, the inability of one user to "borrow" unused resources from another that would otherwise go to waste).

My main concerns here are that I think we may be underestimating (perhaps greatly) the increased operational effort of adding a large number of extra RDS instances and overestimating the benefits of splitting out the really tiny databases. Happy to talk in more detail about this — at this point I think I should stop before this post gets too long 😅

spikeheap:
@kevindew pointed out that all good decisions come with deadlines to make them (and this is particularly pressing given the AWS-mandated maintenance window). We'll give it until the end of next week to work through comments, with a view to deciding this RFC's fate the week after (Monday 6th December).

sengi:
Had a chat with @spikeheap out-of-band which I'll try to summarise here. I think we're much closer aligned on this than it might seem from the discussion here so far 😅

I think it makes complete sense that in the long-term we'll inevitably want all of these databases to be on separate instances (even the ones which are owned by the same team and pose minimal noisy-neighbour risks) if only because we'll probably want to start managing these via Kubernetes operators/[CRDs](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) at some point (which @kerin also pointed out). It also makes handing things off to other teams easier, as the org will inevitably change over time.

The only thing I'm not 100% convinced of myself is front-loading all of the work to split out every database including the small ones right away — but that's really up to you and the folks doing the project. The main reason I have this concern is that the config management side of things is not in great shape — it's an unloved Terraform 0.11 codebase for the RDS config and a very bespoke and even more unloved Puppet 3 codebase for things like configuring database users, roles, passwords, permissions etc. With a bit of luck that stuff might not cause you too much trouble, but it's something to be aware of.

I'm sure in practice you'd probably start by moving one (or a few) of the databases out first anyway, and in doing so quickly get a feel for how straightforward or not those migrations turn out to be.

So in short, I support the idea of aiming for all these databases to be separate. I also don't object if folks want to do all the splitting up front, but I'd just advise prioritising the big/noisy ones and being prepared to defer some of the ones which might not need immediately splitting out (as a contingency in case the going gets tougher than anticipated).

kevindew (approved):
We're reached the deadline now and it looks like a broad consensus has formed that this is a principle GOV.UK would like to adopt.

Thanks for putting this together @spikeheap.

PR #144, opened 2021-11-24
RFC 144 More reliable emergency banner deployment

bilbof:
This proposes making the emergency banner deployment more reliable, by replacing the "display banner immediately" feature with a guarantee to display the banner within 15 minutes.

Rendered RFC: https://github.com/alphagov/govuk-rfcs/blob/bilbof/consistent-cache-expiry/rfc-144-consistent-cache-expiry.md
Deadline for feedback: Wed 1 December 2021.

The tl;dr is that the "display banner immediately" feature is implemented by purging four layers of caches. Our cache invalidation implementation is fragile and has caused problems, notably [this Nov 2021 incident](https://docs.google.com/document/d/1yXGQxQv0IdeWxVegf8UHatnIrCTbd-ZslZB8AxN11QU/edit#) (GDS internal doc).

We do not have an SLA to display the banner instantly. @nicholsj has confirmed that it is acceptable for the banner to appear on the site within 15 minutes. Therefore we can remove this feature, and make the banner deployment significantly more reliable by ensuring our caches expire within 5 minutes.

This will make the banner deployment more reliable.

Thanks to @kevindew for reviewing drafts of this!

### DIFF
--- a/rfc-144-consistent-cache-expiry.md
+++ b/rfc-144-consistent-cache-expiry.md
@@ -1,0 +1,184 @@
+# 144 More reliable emergency banner deployment
+
+Author: Bill Franklin\
+Date: November 2021\
+Deadline: 1 December 2021
+
+## Summary
+
+When we publish the [emergency banner], we clear several caches which contain
+the Static templates, with the intention that the banner will appear on all
+www.gov.uk pages immediately.
+
+Cache invalidation is complex and hard to debug, and is not a suitable tool for
+displaying content immediately across the site.
+
+Importantly, GOV.UK is not required to display the banner _immediately_.
+There can be a 15 minute delay between publishing the banner and the banner
+appearing on all www.gov.uk pages.
+
+Proposed in this RFC:
+
+1. we will reduce TTLs on pages to a consistent 5 minutes
+1. we will not invalidate caches when updating the emergency banner, instead we will wait for 10 minutes
+
+[emergency banner]: https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#header
+
+## Context
+
+For greater reliability GOV.UK runs several caches.
+
+At the edge, we run two main caches:
+
+- Fastly (CDN) caches all www.gov.uk pages, which honours the [Cache-Control]
+  header provided by Origin applications.
+- The [mirror cache] crawls www.gov.uk and stores all pages in multiple S3 buckets
+  and Google GCS for failover purposes when an Origin service is down and the
+  Fastly grace period for cached content has ended.
+
+On our Origin servers, we make use of several caches:
+
+- Varnish (self-hosted on cache machines) also caches all www.gov.uk pages
+  served by Frontend apps.
+- Frontend apps use the Rails cache (backed by Memcached, Redis, or File cache)
+  to [cache templates] served by Static for [60 seconds][slimmer-cache].
+- Static templates are [cached][static-cache] a fourth time by the Static
+  application, to enable Nginx to serve the templates.
+
+## Problem
+
+The emergency banner deploy process is unnecessarily fragile and hard to debug
+due to the "display banner instantly" feature.
+
+The "display banner instantly" feature is implemented using cache invalidation.
+After updating the Static templates to include the emergency banner text,
+the Jenkins job does an in-order purge of four layers of caches in which the
+Static templates are stored.
+
+The feature is necessary only because www.gov.uk pages generally set a cache TTL
+of 30 minutes and we have assumed that emergency banners need to be displayed instantly.
+
+The "display banner instantly" feature is not required, and when our
+implementation (cache invalidation) fails it has been hard to debug.
+
+[mirror cache]: https://docs.publishing.service.gov.uk/manual/fall-back-to-mirror.html
+[Cache-Control]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#the_cache-control_header
+[cache templates]: https://github.com/alphagov/slimmer#caching
+[slimmer-cache]: https://github.com/alphagov/slimmer/blob/b66fcbe9b667b2d946c5922ef42e619a6366c278/lib/slimmer.rb#L10
+[static-cache]: https://github.com/alphagov/static/blob/main/app/controllers/root_controller.rb#L8
+
+## Proposal
+
+The following commitments aim to make the emergency banner deploy process more
+reliable:
+
+* We will remove the "display banner instantly" feature from the
+  Deploy Emergency Banner job.
+  This will involve removing the four cache invalidation sub-jobs.
+* We will ensure that the emergency banner is displayed within 15 minutes.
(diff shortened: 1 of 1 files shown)


fredericfran-gds (approved):
nice work, I agree on the RFC. Thanks for keeping an eye on the traffic increase due to reduced TTL.

rfc-144-consistent-cache-expiry.md
```diff
@@ -0,0 +1,186 @@
...
+
+This is an operational change for GOV.UK Content Designers, so we will need to
+do appropriate comms when making this change.
+
+## Thoughts on future
+
+This RFC could go further and address other cache challenges. These problems
+are beyond the scope of the problem this RFC is seeking to solve.
+
+When TTLs on pages are lower, there will be a stronger argument for retiring
```
ChrisBAshton, line 180: Interesting side-effect of this work!

ChrisBAshton (approved):
The RFC makes sense to me 👍  thanks for putting this together.

rfc-144-consistent-cache-expiry.md
```diff
@@ -0,0 +1,186 @@
+# 144 More reliable emergency banner deployment
+
+Author: Bill Franklin\
+Date: November 2021\
+Deadline: 26 November 2021
+
+## Summary
+
+When we publish the [emergency banner], we clear several caches which contain
+the Static templates, so that the banner will appear on www.gov.uk immediately.
```
nicholsj, line 10: I suggest this is changed to
>with the intention that the banner will appear across www.gov.uk immediately

because a)we know it currently doesn't happen immediately and b)this could be taken to mean the homepage only

rfc-144-consistent-cache-expiry.md
```diff
@@ -0,0 +1,186 @@
...
+## Proposal
+
+The following commitments aim to make the emergency banner deploy process more
+reliable:
+
+* We will remove the "display banner instantly" feature from the
+  Deploy Emergency Banner job.
+  This will involve removing the four cache invalidation sub-jobs.
+* We will ensure that the emergency banner is displayed within 15 minutes.
+  This will involve reducing the default caching time for GOV.UK pages from 30m to 5m.
```
barrucadu, line 78: Not in scope of this RFC, but do you think this means we could possibly get rid of the cache clearing service in the future?

barrucadu: Ah yes, you mention this at the end

barrucadu (approved):
Sounds good to me.  Relying on explicitly purging cache entries has always felt like a bit of a hack.

dgm34:
Nice one @bilbof! No objections to the general approach, but one small related observation: we don't want to wait 5-10 minutes expecting the banner to appear only to realise when it doesn't eventually appear that we made a mistake and need to start again. 

As part of the process of putting the banner live is there a cache-busting URL we can visit (such as https://www.gov.uk/coronavirus?cache=false) for reassurance that we have deployed the banner correctly, so we know we just need to be patient while caches expire? If so we could consider adding that to the [drill Trello card](https://trello.com/c/cgzyi5e4/1246-global-banner-emergency-publishing-drill-22-11-2021)

rfc-144-consistent-cache-expiry.md
```diff
@@ -0,0 +1,186 @@
...
+if this becomes a requirement.
+If we want the banner to appear immediately on all pages, we should not use
+cache invalidation to achieve this.
+Cache invalidation is hard to get right, particularly when the objects we wish
+to invalidate exist in four layers of caching (and more than a dozen caches),
+as we have seen in incidents related to cache invalidation.
+
+Rather than using cache invalidation to immediately display a banner on all pages
+we should implement this feature in another way.
+For example, using Javascript (like a Breaking News banner) or an [edge-side include].
```
injms, line 160: As discussing how this can be done is out of scope of this RFC, should the examples be mentioned at all?

---

For context, the JavaScript use in the breaking news banner is not desirable - from memory, this was a quick and temporary tweak to a dismissable banner done this way because it needed to be done super speedily. It's not a good example of how things could be done when there's time.

I'd hope that any future site-wide banners that need to appear immediately wouldn't only use JavaScript - whilst it'd only exclude a low percentage of visitors, that'd still be a very large number of people.

aldavidson, line 160:
```suggestion

```

aldavidson: I agree with @injms - we should not be relying on Javascript for essential features - and I'd argue that a banner that falls under the definition of 'emergency' also falls under the definition of 'essential'.

I think it's maybe best to just leave it as '..we should implement this feature in another way.' and delete the 'For example'

bilbof:
> As part of the process of putting the banner live is there a cache-busting URL we can visit for reassurance that we have deployed the banner correctly, so we know we just need to be patient while caches expire? 

@dgm34 adding a query param (ie www.gov.uk/help?cachebust=1) will bypass the CDN and Varnish caches, but won't bypass the frontend or Static internal caches (since these don't use query params in their cache keys). With the current proposal we'll need to wait a maximum of 6 minutes to find out if the deploy failed ([frontend cache TTL](https://github.com/alphagov/slimmer/blob/master/lib/slimmer.rb#L10) + [static cache TTL](https://github.com/alphagov/static/pull/2659/files#diff-719014e0b73d01f90fc1a0dfbc5029e8eb614c68c87400b0cbd1883f1036614dR11)).

We could go further and remove the two internal layers of caching, so we could find out if the deploy failed immediately. These caches are a little hard to justify, but removing them I think is probably out of scope. To meet our SLA of a 15 minute deploy we could lower the TTL on Static's cache to 60s, so at worst we'd just have a couple of minutes to wait to find out if the deploy failed. What do you think?

rfc-144-consistent-cache-expiry.md
bilbof: Thanks, I've made this change.

bilbof:
@injms @aldavidson thanks, I've removed the sentence proposing how we might properly implement the "instantly display banner" feature.

dgm34:
Thanks @bilbof. I think that the changes in https://github.com/alphagov/govuk-puppet/pull/11326 are sufficient to keep the suggestion of removing the two internal caches out of scope for the time being. We can always see how it works out and revisit.

kevindew (approved):
This sounds good. I do think 30 minutes is impractically long and I find cache invalidation of HTTP something of an anti-pattern.

The aspect I'm worried about is whether this will lead to any incidents due to the increase in traffic on origin, so as long as we're cautious in that respect then this sounds good.

bilbof:
Hi folks, I'll merge this RFC PR at 4pm today unless there are objections. If you'd like to review further but don't have time today I'd be happy to extend the deadline, just let me know.

bilbof:
I'll merge this in now. Thank you for your reviews, folks! Please reach out if you'd like to see the implementation doc. :shipit:

PR #146, opened 2022-01-07
RFC-145 unarchive govuk admin template

spikeheap:
This PR introduces RFC-145.

The [govuk-admin-template](https://github.com/alphagov/govuk_admin_template) gem has been deprecated for a while, and new GOV.UK admin applications should be built using the [layout component in govuk_publishing_components](https://govuk-publishing-components.herokuapp.com/component-guide/layout_for_admin). Ideally, we would have migrated all our applications to use the new Design System and publishing components, however there are still several in active use which still rely on govuk-admin-template. The GitHub project is archived, and is therefore read-only.

 This RFC proposes un-archiving the project to allow iterative accessibility fixes to be applied to live applications.

 Importantly, this RFC _does not_ propose removing the deprecation notice, or advocate for significant development of govuk_admin_template. It is intended as a stop-gap that acknowledges the slow pace of migration to the Design System, and its intention is to yield accessibility fixes for users prior to those migrations taking place.

### DIFF
--- a/rfc-145-unarchive-govuk_admin_template.md
+++ b/rfc-145-unarchive-govuk_admin_template.md
@@ -1,7 +1,7 @@
 ---
 Author: Ryan Brooks
 Date: January 2022
-Deadline for decision: 14th January 2022
+Deadline for decision: 21st January 2022 # Placeholder, we'll give a week once the PR is ready for review
 ---
 # Unarchive govuk_admin_template Bootstrap project
 ## Summary
--- a/rfc-145-unarchive-govuk_admin_template.md
+++ b/rfc-145-unarchive-govuk_admin_template.md
@@ -26,9 +26,11 @@ With the intention of enabling quick wins for some perennial accessibility issue
 
 1. We will un-archive [govuk-admin-template](https://github.com/alphagov/govuk_admin_template) on GitHub, making it possible to release new versions of the gem with accessibility fixes.
 
-2. GOV.UK developers may release new versions of the gem to address accessibility and usability issues.
+2. GOV.UK developers may release new versions of the gem to address **critical** accessibility and usability issues.
 
-3. Developers are still advised to migrate pages and applications to the GOV.UK Design System where practical.
+3. Teams **MUST** endeavour to migrate pages and applications to the GOV.UK Design System where practical.
+
+4. (Unchanged) Teams **MUST NOT** create new applications, user journeys or significant new features using `govuk-admin-template`.
 
 ## Consequences
 
--- a/rfc-145-unarchive-govuk_admin_template.md
+++ b/rfc-145-unarchive-govuk_admin_template.md
@@ -26,7 +26,7 @@ With the intention of enabling quick wins for some perennial accessibility issue
 
 1. We will un-archive [govuk-admin-template](https://github.com/alphagov/govuk_admin_template) on GitHub, making it possible to release new versions of the gem with accessibility fixes.
 
-2. GOV.UK developers may release new versions of the gem to address **critical** accessibility and usability issues.
+2. GOV.UK developers may release new versions of the gem to address **critical** accessibility, usability, and security issues.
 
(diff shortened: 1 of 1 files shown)


maxgds:
I feel like the template was probably archived specifically to enforce the migration and unarchiving it seems to defy that aim. I can understand the need here but I'm not particularly happy about it. Can it be allowed ONLY on the proviso that the migration work is scheduled, and the unarchiving be timeboxed so that these quick wins come through as one set of quick wins and then it's re-archived? I'm keen to avoid it staying un-archived or the next seemingly quick win from leading to another unarchiving and so on and so forth for all eternity.

spikeheap:
Thanks for taking a look @maxgds. You highlight an important tension here. I don't think anyone wants the old Bootstrap-based templates to live longer than they need to, and we'd all much rather migrate to the Design System. 

> Can it be allowed ONLY on the proviso that the migration work is scheduled, and the unarchiving be timeboxed so that these quick wins come through as one set of quick wins and then it's re-archived?

Absolutely, we can continue to push for this work to be scheduled, and the creation of the long-lived Publishing service area/team should help with this. We're also recruiting for a contract Frontend Engineer for Publishing. Prior to that work being completed, we've really got the choice between enabling accessibility fixes by un-archiving `govuk-admin-template`, or continuing with the current (3.5 year old) approach of leaving these issues unaddressed until a migration happens. My current view is that it's more important to enable the smallest incremental fixes, as there are a _lot_ of competing priorities and uncertainty at the moment, so it's harder to have confidence that larger pieces of work will be consistently prioritised. 

FWIW I see this shift to incrementally fixing issues in our older systems as just part of a bigger journey. By incrementally fixing the issues in `govuk-admin-template` we'll gain a better understanding of the old templating system, which in turn makes it easier to iteratively migrate portions of Whitehall (etc.) to the new components/Design System. In turn, that makes other changes and migrations easier as our understanding of the system increases and we fix the bigger perennial problems.

spikeheap:
I've hardened the language in the proposal bullets to make it clearer that we shouldn't be using `govuk-admin-template`, and added a **MUST** point about migrating to the Design System.

I've also updated the valid changes to scope it down to "critical" fixes.

chao-xian:
Is unarchiving the only option? I mean could we fork to a `govuk-admin-template-stop-gap-die-die-died` project and switch apps to that for a time, or is that a terrible idea?

richardTowers (approved):
I'm in favour of unarchiving (but leaving the deprecation notice).

The past 3.5 years have been a lesson that archiving a repository doesn't result in the organisation increasing the priority of the work to replace it.

I agree that we should still be aiming to replace it. But how exactly we go about that is a complex prioritisation decision (should we migrate whitehall to the design system? or should we migrate all of our content from whitehall to content-store and archive whitehall? etc.).

Depending on how the publishing team's roadmap shakes out, we could be stuck with whitehall / govuk_admin_template for some time yet. We shouldn't let our users suffer unnecessarily in the meantime.

spikeheap:
> Is unarchiving the only option? I mean could we fork to a govuk-admin-template-stop-gap-die-die-died project and switch apps to that for a time, or is that a terrible idea?

🤔 good question @chao-xian. Forking would leave this repository read-only, but would introduce a new copy that was editable. That feels like it increases the complexity and chance of confusion. The repository name would more clearly show its deprecation status, but I think that's already clear from the repository description and note in the header. 

In this context I think forking would effectively be "unarchiving in a roundabout way", for example what would we do when we finally removed it from our apps? Would we keep archives of both the main repository and the fork?

In terms of messaging and confusion, this RFC would only really affect the archive banner:
Screenshot 2022-01-11 at 16 17 20

spikeheap:
We could add a [Pull Request template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) that heavily dissuades you from adding changes, and links back to this RFC. That would provide an additional step and reminder. 

Would that help alleviate concerns about the longevity of this?

PR #147, opened 2022-02-22
RFC 146 - Implement "Production Deploy Access"

richardTowers:
✨  [Rendered](https://github.com/alphagov/govuk-rfcs/blob/production-deploy-access/rfc-146-production-deploy-access.md) :sparkles:

**Deadline for discussion**: 2022-03-09

### DIFF
--- a/rfc-146-production-deploy-access.md
+++ b/rfc-146-production-deploy-access.md
@@ -72,9 +72,6 @@ We will set up the staging and production deploy Jenkins instances to allow peop
 
 The bottleneck of needing people with production access to merge and deploy code will be substantially widened, since it will be much easier to give people deploy access.
 
-Teams may be less willing to have new engineers doing 2ndline shifts, since the carrot of production access will be a bit less appetising if people can get deploy access without the shifts. We'll need to keep an eye on this, since our current position is that 2ndline shifts are independently valuable things for engineers to do.  
-We may need to model this level of access in kubernetes / argo as well as Jenkins as replatforming progresses (unless things have changed further by that time).
-
 ## Appendices
 
 ### Access Levels
--- a/rfc-146-production-deploy-access.md
+++ b/rfc-146-production-deploy-access.md
@@ -51,7 +51,7 @@ This level of access should not include:
 - admin access to other systems (e.g. Fastly etc.)
 - superadmin access to GOV.UK Signon (create and edit all user types and edit applications)
 
-Engineers should be granted this level of access once they've got the required level of security clearance and have been working on GOV.UK for at least one full month. Access should be granted at the discretion of the engineer's tech lead. Before approving access, tech leads should ensure that the engineer:
+Access should be granted at the discretion of the engineer's tech lead, once the engineer has the required level of security clearance. Before approving access, tech leads should ensure that the engineer:
 
 - is aware of our processes and standards around code review
 - understands the responsibilities that releasing code brings with it
--- a/rfc-146-production-deploy-access.md
+++ b/rfc-146-production-deploy-access.md
@@ -1,0 +1,82 @@
+# RFC 146 - Implement "Production Deploy Access"
+
+## Summary
+
+GOV.UK engineers currently have one of two levels of access: "Integration access" or "Production access".
+
+This RFC proposes a middle level of access - "Production deploy access" - which allows engineers to deploy code but not administer related systems. This level of access should be granted to both civil servants and contractors as needed.
+
+Engineers should be given "Production deploy access" once they've been working with us long enough for their tech lead to have confidence they understand how to use the access and where to get help if they need it.
+
+To avoid confusion, the existing "Production" access level should be renamed to "Production Admin" access.
+
+## Problem
+
+There is [a defined set of rules for getting production access](https://docs.publishing.service.gov.uk/manual/rules-for-getting-production-access.html) which requires engineers to do two 2ndline support shifts before they can be granted access.
+
+This can take a long time even in the best case, and doesn't work well in several common situations:
+
+- contractors tend not to do 2ndline shifts, as it's outside their scope of work
+- we make ad hoc access exceptions for senior hires (such as lead technical architects) who need some access to systems, but don't necessarily need full production admin access
+- experience of doing 2ndline shifts is not necessarily the best preparation for working with production systems, particularly for frontend developers
+- our process for when junior developers and apprentices should be given production access isn't clear
+
+There have been numerous cases where teams with only a few people with production access are slowed down because deployments become a bottleneck.
+
+[RFC 128 - Continuous Deployment](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128-continuous-deployment.md) has in some ways made the situation worse. We wanted to make sure that rolling out continuous deployment didn't inadvertently weaken a policy, so we disabled merge access for people without production access on repositories which were continuously deployed.
+
+There were [attempts at formalising this process around 2016](https://docs.google.com/document/d/1lo1JAkFIeWrIl-7bzkzps8rYi2hscV6IOpDzc8dhFLs/edit#). Unfortunately we didn't go into detail on why 2ndline shifts were required. The theory is that the logic went something like this:
+
+- In the early days of GOV.UK our deployment pipeline and monitoring was immature
+- This meant that debugging issues with deployments often required ssh access
+- In turn, this meant that only people with ssh access were allowed to do deployments
+- Because it's easy to make mistakes when using highly privileged access through ssh, we required people to do a couple of 2ndline shifts to make sure they knew what they were doing
+
+Note that by this logic, the 2ndline requirements are defending against accidental mistakes, not against malicious access. We have separate defences against malicious employees (security clearance, probation etc.).
+
+## Proposal
+
+We should implement a new level of access - "Production Deploy Access". This level of access should include:
+
+- Permission to deploy apps in Jenkins (ideally without Jenkins admin permission)
+- Permission to merge continuously deployed applications
+- Readonly access to AWS, Fastly, logging systems, etc.
+- admin access to GOV.UK Signon (create and edit normal users)
+
+This level of access should not include:
+
+- ssh permission to production or staging
+- admin access to the Deploy Jenkins in production or staging
+- admin or poweruser access to AWS
+- admin access to other systems (e.g. Fastly etc.)
+- superadmin access to GOV.UK Signon (create and edit all user types and edit applications)
+
+Engineers should be granted this level of access once they've got the required level of security clearance and have been working on GOV.UK for at least one full month. Access should be granted at the discretion of the engineer's tech lead. Before approving access, tech leads should ensure that the engineer:
+
+- is aware of our processes and standards around code review
+- understands the responsibilities that releasing code brings with it
+- knows how to roll back to an older release if there are any issues
+- knows how to get help from someone with more access if they need it
+
+These rules should be the same for contractors and civil servants at all levels. Junior and apprentice developers should be given production deploy access once their tech lead feels it's appropriate.
+
+The rules for gaining Production Admin Access should remain the same - engineers will still be required to complete two 2ndline shifts.
+
+## Consequences
+
+We will update the [Rules for Getting Production Access](https://docs.publishing.service.gov.uk/manual/rules-for-getting-production-access.html) to match the outcome of this RFC.
+
+We will create a new GitHub Team for the middle level of access, so we'll have "GOV.UK", "GOV.UK Deploy", and "GOV.UK Production".
+
+We will set up the staging and production deploy Jenkins instances to allow people in the GOV.UK Deploy GitHub Team to trigger builds.
+
+The bottleneck of needing people with production access to merge and deploy code will be substantially widened, since it will be much easier to give people deploy access.
+
+Teams may be less willing to have new engineers doing 2ndline shifts, since the carrot of production access will be a bit less appetising if people can get deploy access without the shifts. We'll need to keep an eye on this, since our current position is that 2ndline shifts are independently valuable things for engineers to do.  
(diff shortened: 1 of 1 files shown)


benjamineskola:
Something that was mentioned in the past also was that requiring front-end developers to do second-line shifts may not actually be useful either for the organisation or the individuals involved. If frontend skills don't really map to 2nd line tasks, frontenders may never feel confident doing 2nd line shifts, gain anything from doing 2nd line, etc. — second line ends up being just a sort of artificial barrier to getting deploy access, which is necessary.

On the other hand I don't really want to reduce the pool of developers available for 2nd line, and arguably a lot of 2nd line work isn't especially relevant to backend developer skills either. But this change would at least raise the possibility of reconsidering this.

rfc-146-production-deploy-access.md
```diff
@@ -0,0 +1,82 @@
...
+
+We will update the [Rules for Getting Production Access](https://docs.publishing.service.gov.uk/manual/rules-for-getting-production-access.html) to match the outcome of this RFC.
+
+We will create a new GitHub Team for the middle level of access, so we'll have "GOV.UK", "GOV.UK Deploy", and "GOV.UK Production".
+
+We will set up the staging and production deploy Jenkins instances to allow people in the GOV.UK Deploy GitHub Team to trigger builds.
+
+The bottleneck of needing people with production access to merge and deploy code will be substantially widened, since it will be much easier to give people deploy access.
+
+Teams may be less willing to have new engineers doing 2ndline shifts, since the carrot of production access will be a bit less appetising if people can get deploy access without the shifts. We'll need to keep an eye on this, since our current position is that 2ndline shifts are independently valuable things for engineers to do.
```
maxgds, line 75: If this theory were correct we'd be seeing a drop off of engineer availability after they'd done the minimum for access. Are we? I think that typically teams are reluctant to have engineers do 2nd line shifts because it's disruptive but reluctantly support the model because there's no real alternative.

richardTowers: Yeah, I don't think we do see that, so perhaps this is an unhelpful hypothetical. I'd be happy to remove this paragraph (but let's leave it for a few days incase anyone else has thoughts on this).

fredericfran-gds:
> Something that was mentioned in the past also was that requiring front-end developers to do second-line shifts may not actually be useful either for the organisation or the individuals involved. If frontend skills don't really map to 2nd line tasks, frontenders may never feel confident doing 2nd line shifts, gain anything from doing 2nd line, etc. — second line ends up being just a sort of artificial barrier to getting deploy access, which is necessary.
> 
> On the other hand I don't really want to reduce the pool of developers available for 2nd line, and arguably a lot of 2nd line work isn't especially relevant to backend developer skills either. But this change would at least raise the possibility of reconsidering this.

With the new model of responsibility and apps becoming now decoupled on the new platform, maybe each team can maintain their own service themselves. Might mean more people being on call but no scratching head to figure out how a service works.

rfc-146-production-deploy-access.md
```diff
@@ -0,0 +1,82 @@
...
+
+This level of access should not include:
+
+- ssh permission to production or staging
+- admin access to the Deploy Jenkins in production or staging
+- admin or poweruser access to AWS
+- admin access to other systems (e.g. Fastly etc.)
+- superadmin access to GOV.UK Signon (create and edit all user types and edit applications)
+
+Engineers should be granted this level of access once they've got the required level of security clearance and have been working on GOV.UK for at least one full month. Access should be granted at the discretion of the engineer's tech lead. Before approving access, tech leads should ensure that the engineer:
```
maxgds, line 54: Could/should this be revoked if the tech lead makes the call too early?Should it be a consideration? If so do we need to define the rules for it? Is there anything like that for current existing prod access?

bilbof (approved):
Seems reasonable to me. Delegating decision making to the tech lead on teams is going to make things easier. I remember waiting for a 2nd line shift so I could deploy took a long time and made it harder for me to contribute.

* The admin access to GOV.UK signon proposal needs a bit more  explanation. Would normal user access to production Signon suffice?
* The one month rule seems a little arbitrary, essentially replacing a three month wait with a one month one. Could we delegate this decision entirely to tech leads, or to a formal training process, and remove the one month rule?
* Could you add a note on how you expect this work to get done?

maxgds:
> 

There are plans to revisit how support works, but if we switched to the model you propose immediately it would probably be disastrous for the Find and View team of few developers and wide app ownership. It's hard to even know our apps well, let alone step up to be the sole support team, and we'd be on call as often as off.

rfc-146-production-deploy-access.md
```diff
@@ -0,0 +1,82 @@
...
+Note that by this logic, the 2ndline requirements are defending against accidental mistakes, not against malicious access. We have separate defences against malicious employees (security clearance, probation etc.).
+
+## Proposal
+
+We should implement a new level of access - "Production Deploy Access". This level of access should include:
+
+- Permission to deploy apps in Jenkins (ideally without Jenkins admin permission)
+- Permission to merge continuously deployed applications
+- Readonly access to AWS, Fastly, logging systems, etc.
+- admin access to GOV.UK Signon (create and edit normal users)
```
sihugh, line 44: Is this mainly to grant yourself permission to apps to debug etc? If so, could that be a different permission too?

rfc-146-production-deploy-access.md
```diff
@@ -0,0 +1,82 @@
...
+
+This level of access should not include:
+
+- ssh permission to production or staging
```
brucebolt, line 48: Should there also be a restriction on deploying Puppet?

Otherwise a deploy-only user could give themselves higher levels of access (e.g. SSH into prod machines) or make changes to Jenkins tasks by deploying a branch of puppet.

benjamineskola: >a deploy-only user could give themselves higher levels of access

This kind of depends on whether we're thinking of this as a guard against *malicious* users or simply careless/inexperienced ones.

And of course even with production deploy access they'd need someone else to review; it's not an absolute protection but should make it harder to make questionable changes.

richardTowers: Yeah, this isn't intended to guard against malicious users - doing 2ndline shifts doesn't make you any more trustworthy.

There are plenty of ways a malicious user could use "deploy" access to do escalate to higher levels (I won't go in to them here because it's a public forum, but if it's worthwhile we could discuss them in a google doc addendum to this RFC).

The point about "they'd need someone else to review" is interesting - at the moment, the Deploy App job takes a commit-ish, so there's no protection (that I'm aware of) against someone asking Jenkins to deploy a branch (which doesn't have to be reviewed).

rfc-146-production-deploy-access.md
ChrisBAshton: I echo Bill's point about the 1 month being arbitrary. Suggest we make this a guideline instead, e.g.:

```suggestion
Access should be granted at the discretion of the engineer's tech lead, once the engineer has the required level of security clearance. Before approving access, tech leads should ensure that the engineer:
```

rfc-146-production-deploy-access.md
maxgds: We have on occasion wanted to deploy a branch to staging. I know this for sure but can't recall why. Production I'm with you, though.

rfc-146-production-deploy-access.md
```diff
@@ -0,0 +1,82 @@
...
+
+These rules should be the same for contractors and civil servants at all levels. Junior and apprentice developers should be given production deploy access once their tech lead feels it's appropriate.
+
+The rules for gaining Production Admin Access should remain the same - engineers will still be required to complete two 2ndline shifts.
+
+## Consequences
+
+We will update the [Rules for Getting Production Access](https://docs.publishing.service.gov.uk/manual/rules-for-getting-production-access.html) to match the outcome of this RFC.
+
+We will create a new GitHub Team for the middle level of access, so we'll have "GOV.UK", "GOV.UK Deploy", and "GOV.UK Production".
```
ChrisBAshton, line 69: I like the team names, but wonder if these should be consistent with the terminology used earlier, i.e.

```suggestion
We will create a new GitHub Team for the middle level of access, so we'll have "GOV.UK", "GOV.UK Production Deploy", and "GOV.UK Production Admin".
```

(Or use "GOV.UK Deploy" and "GOV.UK Production" throughout this RFC).

rfc-146-production-deploy-access.md
ChrisBAshton: Agree this paragraph should be removed. Our 2nd line slots aren't volunteer-only, they're mandatory and allocated, so how a team feels about doing shifts shouldn't be relevant.

ChrisBAshton (commented):
LGTM 👍 thanks for putting this together. Just a couple of suggestions below.

nicholsj (commented):
Specifically on Signon permissions - I wonder if that should be kept out of this proposal? That has a different set of stakeholders (eg Content teams) and existing processes around granting access  (eg to a publishing app). 
It feels like the changes to deployment permissions and AWS/Logit/etc access will solve a lot of the biggest issues?

maxgds:
> Specifically on Signon permissions - I wonder if that should be kept out of this proposal? That has a different set of stakeholders (eg Content teams) and existing processes around granting access (eg to a publishing app). It feels like the changes to deployment permissions and AWS/Logit/etc access will solve a lot of the biggest issues?

It has to be covered as part of this proposal since it's currently tied to "production access". If we were state clearly that this is entirely for techies and other users/stakeholders may also have rules around Signon access at different levels would that be good enough?

nicholsj:
> > Specifically on Signon permissions - I wonder if that should be kept out of this proposal? That has a different set of stakeholders (eg Content teams) and existing processes around granting access (eg to a publishing app). It feels like the changes to deployment permissions and AWS/Logit/etc access will solve a lot of the biggest issues?
> 
> It has to be covered as part of this proposal since it's currently tied to "production access". If we were state clearly that this is entirely for techies and other users/stakeholders may also have rules around Signon access at different levels would that be good enough?

Yes that sounds like a useful distinction, ta

kevindew:
I've bumped the deadline on this until 9th March for when @richardTowers returns from leave.

richardTowers:
Thanks Kevin! I'm back now and will try to address the comments. Sounds like there's general approval, just some minor cleanup to do 🧹

rfc-146-production-deploy-access.md
richardTowers: I don't think we've got a process for revoking access if it's granted too early - I think it should be rare enough that we can handle it exception by exception.

rfc-146-production-deploy-access.md
richardTowers: I'm going to resolve this thread without putting any special restrictions on puppet.

I do think there's a wider piece about tightening up our access controls, but it's very tied up with the replatforming project and the move to Argo. I'm sure we can improve on what we've got now.

rfc-146-production-deploy-access.md
```diff
@@ -0,0 +1,82 @@
...
+- superadmin access to GOV.UK Signon (create and edit all user types and edit applications)
+
+Engineers should be granted this level of access once they've got the required level of security clearance and have been working on GOV.UK for at least one full month. Access should be granted at the discretion of the engineer's tech lead. Before approving access, tech leads should ensure that the engineer:
+
+- is aware of our processes and standards around code review
+- understands the responsibilities that releasing code brings with it
+- knows how to roll back to an older release if there are any issues
+- knows how to get help from someone with more access if they need it
+
+These rules should be the same for contractors and civil servants at all levels. Junior and apprentice developers should be given production deploy access once their tech lead feels it's appropriate.
```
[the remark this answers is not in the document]
richardTowers: I've already removed the requirement for 1 month's tenure in 1dcc5972, so I'll make this:

```suggestion
These rules should be the same for contractors and civil servants at all levels, including juniors and apprentices.
```

rfc-146-production-deploy-access.md
richardTowers: I sort of agree, but I'm not sure what I might break by renaming the GitHub teams. At least [the Jenkins Access controls](https://github.com/alphagov/govuk-puppet/blob/e5f47f2d4330ff0713145f93747364921ad3e4b9/hieradata_aws/class/production/jenkins.yaml#L28) would have to be updated to include the new team names.

Can you think of anything else? I can shoot for the clearer names, and if it becomes too difficult I guess I can rename them back.

richardTowers:
@nicholsj - I've scoped down the signon permissions to just "Normal" permissions in 40d20d8, instead of admin permissions which do seem a little more broad than most people would need. We can see how that goes.

I'm not sure that completely covers the concerns about whether signon access should be in scope for this. My view is if someone's got permission to deploy signon, we may as well give them permission to use it.

rfc-146-production-deploy-access.md
```diff
@@ -0,0 +1,79 @@
...
+
+Note that by this logic, the 2ndline requirements are defending against accidental mistakes, not against malicious access. We have separate defences against malicious employees (security clearance, probation etc.).
+
+## Proposal
+
+We should implement a new level of access - "Production Deploy Access". This level of access should include:
+
+- Permission to deploy apps in Jenkins (ideally without Jenkins admin permission)
+- Permission to merge continuously deployed applications
+- Readonly access to AWS, Fastly, logging systems, etc.
```
huwd, line 43: I fully 👍🏻 this RFC, but particularly for these points.
I like the idea of getting more read only access to things like AWS early on.

I would have liked to have been able to follow along more when pairing with others, earlier. Or had a poke around to familliarise myself, before I was able to make changes.

huwd (approved):

leenagupte:
All of the suggestions sound good 👍. One thing I've noticed in the spreadsheet is that people would still need full admin access to access CI. Could the new permission be used to re-trigger builds?

PR #148, opened 2022-05-03
RFC-147 - Allow SpeedCurve to capture a user's HTTP Protocol data

Nooshu:
Allow SpeedCurve to capture a user's HTTP Protocol for use as a comparison dimension in the GUI.

✨ [Rendered](https://github.com/alphagov/govuk-rfcs/blob/103acd22a74b32d4ea9dd321bc0ab05ac664aa82/rfc-147-enable-speedcurve-http-protocol-capture.md) ✨ 

**Deadline**: 2022-05-10

### DIFF
--- a/rfc-147-enable-speedcurve-http-protocol-capture.md
+++ b/rfc-147-enable-speedcurve-http-protocol-capture.md
@@ -12,8 +12,6 @@ Data capture to quantify the change from HTTP/2 + TCP to HTTP/3 + QUIC.
 
 ## Proposal
 
-Describe your proposal, with a focus on clarity of meaning. You MAY use [RFC2119-style](https://www.ietf.org/rfc/rfc2119.txt) MUST, SHOULD and MAY language to help clarify your intentions.
-
 I'd like to include a small piece of JavaScript in the current SpeedCurve RUM implimentation that will only fire when a user accepts the cookie banner. This JavaScript will look at the current HTTP protocol the user is using and push this anonymous data into SpeedCurve using their API. 
 
 This will then give us an additional dimension in the SpeedCurve GUI with which to compare before / after the change is made.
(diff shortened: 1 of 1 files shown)


rfc-147-enable-speedcurve-http-protocol-capture.md
```diff
@@ -0,0 +1,35 @@
...
+
+Data capture to quantify the change from HTTP/2 + TCP to HTTP/3 + QUIC.
+
+## Proposal
+
+Describe your proposal, with a focus on clarity of meaning. You MAY use [RFC2119-style](https://www.ietf.org/rfc/rfc2119.txt) MUST, SHOULD and MAY language to help clarify your intentions.
+
+I'd like to include a small piece of JavaScript in the current SpeedCurve RUM implimentation that will only fire when a user accepts the cookie banner. This JavaScript will look at the current HTTP protocol the user is using and push this anonymous data into SpeedCurve using their API. 
+
+This will then give us an additional dimention in the SpeedCurve GUI with which to compare before / after the change is made.
```
injms, line 19:
```suggestion
This will then give us an additional dimension in the SpeedCurve GUI with which to compare before / after the change is made.
```

rfc-147-enable-speedcurve-http-protocol-capture.md
```diff
@@ -0,0 +1,35 @@
...
+
+The actual JavaScript is small and will have 0 effect of page performance as it is simply reading a string value from an object in a supporting browsers [Navigation Timing Level 2 API](https://www.w3.org/TR/navigation-timing-2/).
```
injms, line 21: Slightly picky - nothing has zero effect, so suggest negligible effect?
```suggestion
The actual JavaScript is small and will have a negligible effect of page performance as it is simply reading a string value from an object in a supporting browsers [Navigation Timing Level 2 API](https://www.w3.org/TR/navigation-timing-2/).
```

rfc-147-enable-speedcurve-http-protocol-capture.md
```diff
@@ -0,0 +1,35 @@
...
+
+```js
+// check see if browser supports navigation timing API and has the HTTP protocol information.
+if(typeof window.performance.timing === "object" && performance.getEntriesByType('navigation')[0].nextHopProtocol === "String"){
+	// returns "h1", "h2", "h3"
+	var http-protocol = performance.getEntriesByType('navigation')[0].nextHopProtocol;
+	// push the users protocol data into SpeedCurve RUM
+	LUX.addData("http-protocol", http-protocol)
+}
+```
```
injms, line 33: I think that this example should either be safer - for example, more checks to see if `window.performance` exists before checking for `window.performance.timing`, since the latter will throw an error if `window.performance` isn't present.

Or it should just say that `performance.getEntriesByType('navigation')[0].nextHopProtocol` will return h1, h2, or h3 and is the thing that should be reported to the `LUX.addData()` function.

Nooshu: I'm open to structure the JS in the safest possible and most efficient way, so whichever works best. It could be as simple as this? Is this what you are suggesting in the 2nd paragraph?

```js
LUX.addData("http-protocol", performance.getEntriesByType('navigation')[0].nextHopProtocol);
```

injms: Maybe something like:

> We should report the contents of `performance.getEntriesByType('navigation')[0].nextHopProtocol` using the `LUX.addData()` function - SpeedCurve have some good example of this [in their recipes][1].


[1]: https://support.speedcurve.com/recipes/track-size-for-a-single-resource

Nooshu: Agreed, now updated the text and the code example.

injms: Thank you! Sorry to be picky - I'm overly wary of being prescriptive in examples in case it's taken as "this is the only way" rather than "this could be a good way, try it and see".

injms (approved):
I think this would be an interesting investigation of how supported HTTP/3 is and also a good test of how we can better use LUX.

kevindew:
@Nooshu doesn't look like we've reached the deadline for this so it seems premature to have merged this.

To be honest I feel this type of decision doesn't really need an RFC - as it doesn't look like we're seeking programme wide developer consensus here. I'd imagine this would have been ok with a PR in publishing components to implement it.

PR #154, opened 2022-10-11
RFC-154 - Integrating a CMS with GOV.UK to power custom pages

kevindew:
Trello: https://trello.com/c/tpDtfZ7f/324-put-together-rfc-request-for-comments-on-introducing-a-3rd-party-cms-to-govuk

[👉 Rendered version 👈 ](https://github.com/alphagov/govuk-rfcs/blob/ots-cms/rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md)

Deadline for comments: November 3rd

This explains the proposed approach GOV.UK Publishing are taking considering which would enable the integration of an off-the-shelf CMS with GOV.UK for an experiment.

### DIFF
--- a/rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
+++ b/rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
@@ -10,22 +10,22 @@ Examples of these types of pages, and their content management solutions, are:
 - [Travel Advice index](https://www.gov.uk/foreign-travel-advice), where content is [hardcoded](https://github.com/alphagov/frontend/blob/f2af829b14a9e6af21deb49c9bd0484e9ea49baa/app/views/travel_advice/index.html.erb) in HTML and needs developer support to amend.
 - [Coronavirus landing page](https://www.gov.uk/coronavirus), where [bespoke CMS tooling](https://github.com/alphagov/collections-publisher/tree/d68bf4cf910b62d7827f488ed2f20126bc9b25a4/app/controllers/coronavirus) was built for a single page at a high implementation cost.
 
-In 2022 GOV.UK again has the need for a new bespoke page, one to provide [Cost of living support](https://www.gov.uk/cost-of-living). It has been decided that GOV.UK will experiment with integrating an off-the-shelf [headless CMS](https://en.wikipedia.org/wiki/Headless_content_management_system) to meet the CMS needs for this page. This is to explore a hypothesis that with flexible CMS tooling a bespoke page could be built more simply, with the CMS functionality  managed entirely by the team developing the frontend code. The purpose of this RFC is to propose how a headless CMS product can integrate with GOV.UK infrastructure.
+In 2022 GOV.UK again has the need for a new bespoke page, one to provide [Cost of living support](https://www.gov.uk/cost-of-living). It has been decided that GOV.UK will experiment with integrating an off-the-shelf [headless CMS](https://en.wikipedia.org/wiki/Headless_content_management_system) to meet the CMS needs for this page. This is to explore a hypothesis that with flexible CMS tooling a bespoke page could be built more simply, with the CMS functionality managed entirely by the team developing the frontend code. The purpose of this RFC is to propose how a headless CMS product can integrate with GOV.UK infrastructure.
 
-_Note, to facilitate focused discussion it's worth clarifying the scope of this RFC, which is limited to how an off-the-shelf CMS can be integrated with GOV.UK. This RFC only consider an off-the-shelf CMS for the purpose of meeting existing unmet CMS needs on GOV.UK and is intended to supplement existing GOV.UK CMS tools, this is not an indication that these will be replaced by an off-the-shelf CMS tool. It is outside the scope of this RFC to discuss whether GOV.UK should be building their own CMS for a custom page - a GOV.UK Senior Management Team decision was made to explore a non-bespoke solution for this situation. It is also outside the scope of this RFC to discuss which CMS will be used and the operational concerns - for the initial implementation it was decided that [Contentful](https://contentful.com) will be used, and should the experiment be a success a wider procurement exercise will consider CMS products._
+_Note, to facilitate focused discussion it's worth clarifying the scope of this RFC, which is limited to how an off-the-shelf CMS can be integrated with GOV.UK. This RFC only considers an off-the-shelf CMS for the purpose of meeting existing unmet CMS needs on GOV.UK and is intended to supplement existing GOV.UK CMS tools. This is not an indication that these will be replaced by an off-the-shelf CMS tool. It is outside the scope of this RFC to discuss whether GOV.UK should be building their own CMS for a custom page; a GOV.UK Senior Management Team decision was made to explore a non-bespoke solution for this situation. It is also outside the scope of this RFC to discuss which CMS will be used and the operational concerns; for the initial implementation, it was decided that [Contentful](https://contentful.com) will be used, and should the experiment be a success, a wider procurement exercise will consider CMS products._
 
 ## Problem
 
-While it is relatively straight forward to model GOV.UK content in a headless CMS tool, it is less clear how this data will integrate with the GOV.UK stack.
+While it is relatively straightforward to model GOV.UK content in a headless CMS tool, it is less clear how this data will integrate with the GOV.UK stack.
 
 There are key considerations to make such as:
 
 - Will data from a CMS be input into the publishing pipeline (Publishing API, Content Store, etc) or bypass it?
-- How might pages powered by a CMS result in a routes registered with Router?
+- How might pages powered by a CMS result in a route registered with Router?
 - Could a page powered by a non-GOV.UK CMS have a presence in Search API?
-- Will GOV.UK rendering applications need to consider another content source than Content Store and Search API?
+- Will GOV.UK rendering applications need to consider a content source other than Content Store and Search API?
 - Should the CMS be offline would we accept unavailability of a public GOV.UK page or need a fallback system?
-- Whether modelling changes in a CMS would result in needing to make other changes in the GOV.UK publishing stack?
+- Will modelling changes in a CMS result in needing to make other changes in the GOV.UK publishing stack?
 
 ## Proposal
 
@@ -54,7 +54,7 @@ The [`special_route`](https://github.com/alphagov/govuk-content-schemas/blob/mai
 ### Further considerations
 
 - If the CMS was offline and we needed to make edits to content this would be done by a short term hardcoding in the frontend application to set content.
-- It is unlikely, due to complexity reasons, that we would integrate [Asset Manager](https://github.com/alphagov/asset-manager) with the CMS tooling and instead rely on the CMS to host assets such as images. If the CMS was offline these would be unavailable, but as the page would still be available this would not be a critical issue to resolve.
+- The initial integration will not integrate with [Contentful Media](https://www.contentful.com/help/working-with-media/) as we have yet to decide whether these would need to be hosted on a GOV.UK domain or not.
 - Should we change CMS we would retire contentful-listener-api in favour of an appropriate one, or rename contentful-listener-api to be cms-listener-api to be a generic tool that serviced multiple CMS tools.
-- Should the result of the experiment be that we will not continue with an off-the-shelf CMS the Publishing team will revert all changes made to the stack.
+- Should the result of the experiment be that we will not continue with an off-the-shelf CMS the Publishing team will revert all changes made to the stack, any pages powered by Contentful would be migrated to using hardcoded content.
 - Should the experiment with the Cost of living support page be successful we may experiment with supporting other custom pages such as GOV.UK homepage or GOV.UK Roadmap or potentially the content of a Smart Answer.
--- a/rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
+++ b/rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
@@ -1,0 +1,60 @@
+# Integrating an off-the-shelf CMS with GOV.UK infrastructure to power custom pages
+
+## Summary
+
+GOV.UK has built a number of bespoke content management systems (CMS) to enable the publishing of a variety of pre-defined document types. However, these tools are of limited assistance when a need emerges to create a one-off page on GOV.UK (hereafter referred to as custom page) that is not one of these pre-defined document types.
+
+Examples of these types of pages, and their content management solutions, are:
+
+- [GOV.UK Homepage](https://www.gov.uk/) and [GOV.UK Roadmap](https://www.gov.uk/roadmap), where Rails I18n has been used as a basic [YAML CMS system](https://github.com/alphagov/frontend/blob/f2af829b14a9e6af21deb49c9bd0484e9ea49baa/config/locales/en.yml#L454-L576) which require developer support and complicates actual internationalisation management.
+- [Travel Advice index](https://www.gov.uk/foreign-travel-advice), where content is [hardcoded](https://github.com/alphagov/frontend/blob/f2af829b14a9e6af21deb49c9bd0484e9ea49baa/app/views/travel_advice/index.html.erb) in HTML and needs developer support to amend.
+- [Coronavirus landing page](https://www.gov.uk/coronavirus), where [bespoke CMS tooling](https://github.com/alphagov/collections-publisher/tree/d68bf4cf910b62d7827f488ed2f20126bc9b25a4/app/controllers/coronavirus) was built for a single page at a high implementation cost.
+
+In 2022 GOV.UK again has the need for a new bespoke page, one to provide [Cost of living support](https://www.gov.uk/cost-of-living). It has been decided that GOV.UK will experiment with integrating an off-the-shelf [headless CMS](https://en.wikipedia.org/wiki/Headless_content_management_system) to meet the CMS needs for this page. This is to explore a hypothesis that with flexible CMS tooling a bespoke page could be built more simply, with the CMS functionality  managed entirely by the team developing the frontend code. The purpose of this RFC is to propose how a headless CMS product can integrate with GOV.UK infrastructure.
+
+_Note, to facilitate focused discussion it's worth clarifying the scope of this RFC, which is limited to how an off-the-shelf CMS can be integrated with GOV.UK. This RFC only consider an off-the-shelf CMS for the purpose of meeting existing unmet CMS needs on GOV.UK and is intended to supplement existing GOV.UK CMS tools, this is not an indication that these will be replaced by an off-the-shelf CMS tool. It is outside the scope of this RFC to discuss whether GOV.UK should be building their own CMS for a custom page - a GOV.UK Senior Management Team decision was made to explore a non-bespoke solution for this situation. It is also outside the scope of this RFC to discuss which CMS will be used and the operational concerns - for the initial implementation it was decided that [Contentful](https://contentful.com) will be used, and should the experiment be a success a wider procurement exercise will consider CMS products._
+
+## Problem
+
+While it is relatively straight forward to model GOV.UK content in a headless CMS tool, it is less clear how this data will integrate with the GOV.UK stack.
+
+There are key considerations to make such as:
+
+- Will data from a CMS be input into the publishing pipeline (Publishing API, Content Store, etc) or bypass it?
+- How might pages powered by a CMS result in a routes registered with Router?
+- Could a page powered by a non-GOV.UK CMS have a presence in Search API?
+- Will GOV.UK rendering applications need to consider another content source than Content Store and Search API?
+- Should the CMS be offline would we accept unavailability of a public GOV.UK page or need a fallback system?
+- Whether modelling changes in a CMS would result in needing to make other changes in the GOV.UK publishing stack?
+
+## Proposal
+
+We propose that integrating an off-the-shelf headless CMS into GOV.UK should involve the CMS syncing data with the Publishing API. This would allow:
+
+- no increase in dependency for GOV.UK rendering applications, they continue to receive page content from Content Store and do not need to consider CMS unavailability
+- ability to integrate a page with Search API using existing Publishing API message queue approach
+- implicit updating of Router API
+
+However in order to facilitate this there would need to be new software built, to bridge communication between a CMS and GOV.UK, and changes to existing software to facilitate this bridging.
+
+### Bridging software
+
+It is common that headless CMS tools make use of [webhooks](https://en.wikipedia.org/wiki/Webhook) as a means to notify subscribing software to events that occur in the CMS. We propose the introduction of a piece of software that will listen for a CMS's webhooks and use this information as an initiator of a process to update the content.
+
+This software will identify whether a webhook event signifies a live or draft event and update the Publishing API accordingly by the content data in the CMS (typically a JSON blob).
+
+For integration with Contentful, we will deploy a new GOV.UK application: Contentful Listener API. This was [developed](https://github.com/alphagov/contentful-listener-api) whilst spiking approaches to this problem. There is documentation written to the describe the [workflow](https://github.com/alphagov/contentful-listener-api/blob/main/docs/how-this-application-works.md) of this tool and [integration limitations](https://github.com/alphagov/contentful-listener-api/blob/main/docs/integration-limitations.md). This application will be owned by the Publishing Platform team.
+
+### Changes to existing software
+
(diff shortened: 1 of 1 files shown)


rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
```diff
@@ -0,0 +1,60 @@
+# Integrating an off-the-shelf CMS with GOV.UK infrastructure to power custom pages
+
+## Summary
+
+GOV.UK has built a number of bespoke content management systems (CMS) to enable the publishing of a variety of pre-defined document types. However, these tools are of limited assistance when a need emerges to create a one-off page on GOV.UK (hereafter referred to as custom page) that is not one of these pre-defined document types.
+
+Examples of these types of pages, and their content management solutions, are:
+
+- [GOV.UK Homepage](https://www.gov.uk/) and [GOV.UK Roadmap](https://www.gov.uk/roadmap), where Rails I18n has been used (and arguabably misused) as a basic [YAML CMS system](https://github.com/alphagov/frontend/blob/f2af829b14a9e6af21deb49c9bd0484e9ea49baa/config/locales/en.yml#L454-L576) which require developer support.
```
injms, line 9: I'd prefer that this part of the sentence be removed as it's kind of unhelpful, it violates the the [retrospective prime directive](https://www.funretrospectives.com/the-retrospective-prime-directive/), and we _know_ this anyway as otherwise this RFC wouldn't be being raised.

```suggestion
- [GOV.UK Homepage](https://www.gov.uk/) and [GOV.UK Roadmap](https://www.gov.uk/roadmap), where Rails I18n has been used as a basic [YAML CMS system](https://github.com/alphagov/frontend/blob/f2af829b14a9e6af21deb49c9bd0484e9ea49baa/config/locales/en.yml#L454-L576) which require developer support.
```

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
```diff
@@ -0,0 +1,60 @@
...
+
+The [`special_route`](https://github.com/alphagov/govuk-content-schemas/blob/main/formats/special_route.jsonnet) schema in [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas) will be updated to accept a details field. This details field will not have restrictions on how the data is structured, this will allow changes in the modelling of content without needing a schema to be updated (this was an [approach explored](https://github.com/alphagov/govuk-content-schemas/blob/170a941b42bbc53d529b4c50698ad5ba50776df5/formats/coronavirus_landing_page.jsonnet#L4-L8) for the Coronavirus landing page). It is theorised that having a schema for a custom page is more of a hindrance than benefit given the data is only used in a single place.
```
BeckaL, line 52: Is it right that we modify the existing special route schema, rather than add a new, generic one for externally generated content? I take your point about it only being used in a single place (for now), but I guess there's a reason that existing special routes explicitly cannot have details sections, so modifying them maybe feels a bit wrong.

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Further considerations
+
+- If the CMS was offline and we needed to make edits to content this would be done by a short term hardcoding in the frontend application to set content.
+- It is unlikely, due to complexity reasons, that we would integrate [Asset Manager](https://github.com/alphagov/asset-manager) with the CMS tooling and instead rely on the CMS to host assets such as images. If the CMS was offline these would be unavailable, but as the page would still be available this would not be a critical issue to resolve.
```
BeckaL, line 57: I feel a bit unsure about this, although obviously I haven't looked into the complexity. Feels a bit of a shame that we're using the common backend for the content, but not for the assets. I guess I'm thinking of 
* Adding complexity to the stack
* Having to migrate if we change tack
Obviously not a big deal if it's only ends up being one page, but thinking of this being used more widely.

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
kevindew: Thanks for the feedback, I’ve edited the text. It wasn’t intended as a comment on individuals choices so it’s good to resolve that.

kevindew:
Thanks for the comments Max, sorry for taking a while to get back to you. I’ve had a look at where the text can be amended to answer you from the RFC and made a couple of tweaks but feel it’s better to take most of the points directly.
 
> What would the CMS publish? json? html fragments? html pages? I'm assuming not pages since you referred to entities like accordion sections.

As a headless CMS the data would be available as JSON typically based on the structure of the content models. The models themselves are likely to be built up for the similar sort of data structures you find in GOV.UK schemas: string fields, number fields, date fields, booleans, arrays, objects. There’s unlikely to be HTML fragments directly but likely to be strings of markdown/govspeak. 

> What is the proposition? An externally hosted thing like wordpress.org, or an internally run piece of software hosted somewhere on our stack? If the latter what thought has been given to handling dependencies if any, and software updates eg security patches?

The initial software tried is Contenful which is a externally hosted SAAS, so we’d not be updating it.

> In the scenario where the CMS is unavailable and we work around it by publishing direct, what is the recovery strategy when it comes back up? How do we avoid clashing parallel versions that are out of sync?

The suggestion wouldn’t be to publish direct in that scenario it would be to modify the frontend rendering app to override a field. So something hacky like:

```
<% if accordion[“id”] == “123” %>
  <h1>Temporary name</h1>
<% else %>
  <h1><%= accordion[“name”] %></h1>
<% endif %>
```

So recovering would be a case of updating the CMS once it’s back online and then removing the hacky code. 

Do bear in mind this is a low likehood scenario, as I know the above looks a bit grim - but it wouldn’t be expected to be remotely part of day-to-day coding.

> You mentioned IDs in the publishing system linked to content published in the CMS, eg accordion sections. This is great for scenarios where the content areas/CMS elements are fixed with direct mappings and the existing content just needs a refresh, but how does it handle adding or removing new elements - say the accordion needs to expand to 12 instead of 10 sections, or a smart answer wants to lose a question from the middle?

So if you need to add or remove items you will have to edit their parent item for that action. Thus you will still be notified of the change.

I.e if we had cost of living page with ID `e1`, that had two accordions (`e2` and `e3`).

If we edited `e2` we’d be able to determine that we need to update the cost of living page as we’d find `e2` as an ID in the Publishing API for the cost of living page.

If we added a new accordion (`e4`) to the page, we’d have to edit `e1` to add the accordion and would be notified that `e1` had changed - so it wouldn’t matter that we weren’t tracking `e4` at that stage.

If we removed an accordion `e3` from the page, there’s two ways that could happen 1) we delete the entity `e3` from the CMS or 2) we edit the page (`e1`) to remove it from the collection. Both of those would produce different webhooks but as we track both `e3` and `e1` we still get the page updated.

> In our publishing tools we have the concept of editions baked into at least a proportion of output. Is this or some other form of version control something needed in the CMS?

Yes, it’s a standard feature in CMS to track versions of content and have a draft and published state. There are other more complex features that are more bespoke to GOV.UK for CMS content management that are flagged in the [limitations doc](https://github.com/alphagov/contentful-listener-api/blob/main/docs/integration-limitations.md#publishing-life-cycle-metadata).

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
kevindew: The existing ones don't have a details section as they are typically hardcoded content. 

We’ve got around this in the past with creating new schemas such for handling freeform data like [coronavirus_landing_page](https://github.com/alphagov/govuk-content-schemas/blob/cdc8bfe8b13a24285b7dcae99248adb1dd31de29/formats/coronavirus_landing_page.jsonnet).

I’d argue that these pages however are very much in common with most special routes as they are indeed special, and unlikely other schemas, don’t have a fixed data structure. Most special routes are used for custom GOV.UK pages and these will also be custom GOV.UK pages.

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
kevindew: As the Publishing API doesn’t have a relationship with asset manager the only thing we’d do if we had an asset manager file is replace the CMS URL in the content item with the asset manager one.

We’d then lose the ability to utilise any extra asset APIs, for example with Contentful we can use their [image API](https://www.contentful.com/developers/docs/references/images-api/) for dynamic resizing.

The one thing that I feel is a loss with using the CMS assets is the hostname (I like vanity URLs :laughing:) as with Contenful they’d be on cfassets.com. However I feel like, if we had to resolve that, I’d prefer to proxy to the CMS’ asset server off one of our hostnames.

kevindew: Thanks Ollie, some really interesting points.

> 1. Avoid breaking links to assets

This situation seems a concern if non-image files are uploaded for downloading, but doesn't seem to be too much of an issue for an image file. In the discussed use cases for this images seemed the only likely use case. I'm not too concerned if they break, as they feel like sub-resources.

> 2. Incomplete archives on The National Archives

This is a good point. It'd be worth finding out their policies. I wasn't aware this would be a restriction

> 3. Lack of control

Do agree it seems a bit annoying to lose control. This seems like something we'd only have limited control of with a proxy too.

---

I did ponder that setting up a proxy might be more delicate than I considered. We'd need to only allow proxying particular URLs of a CMS otherwise it could look like GOV.UK was hosting assets for other clients. I guess taking control of any rogue URLs would be a bit of pain as it'd be done through nginx config - but I imagine it's so rare it's not too big a deal.

---

Hmm tricky. I'm imagining the easiest resolution if there's not a clear route would be to defer a decision on what to do with assets until we learn more about if the experiment is successful, as it's not a clear need yet and thus likely won't get a decent amount of attention.

I'd think that considering Contentful, a proxying approach would be a good option for keeping things like image resizing, which I imagine could be frustrating to not have. For the issues you've outlined though an Asset Manager integration would be better. Though I imagine doing it would be a bit of a headache in trying to determine whether an asset needs updating or not.

ChrisBAshton: This does feel like the most contentious part of the proposal. A few points:

1) Agreed that proxying to the CMS asset URL opens up a bit of a security risk in terms of bad actors hosting offensive content and being able to craft GOV.UK URLs that link to said content.

2) Linking directly to `cfassets.com` also not ideal, I'm thinking of people who may have script blockers where they've allow-listed GOV.UK but not third party providers. A small point, but worth considering.

3) The most important point IMO is the migration strategy - it seems highly likely we'd eventually move away from whichever CMS is picked for this experiment, and then be left with a migration job of having to copy all the images to GOV.UK's asset manager and rewrite all the affected content in Publishing API to refer to the new image names.
We also may not actually _want_ to utilise the dynamic image sizing functionality in the Contentful Images API, as it would likely increase the complexity of the integration with GOV.UK's stack.

One thought that came to mind was building some sort of simple asset uploader CMS, where publishers can upload assets directly to Asset Manager, and then copy the resulting asset URL into the headless CMS. A bit more clunky, perhaps, but less of a headache when it comes to migration.
Would need to think about how to handle alt text - the asset uploader CMS feels like a reasonable fit for that, but it would be another thing to have to copy and paste, so probably best handled within the headless CMS.

ChrisBAshton (commented):
Interesting proposal! A few typos and a comment or two below.

The only thing the RFC doesn't cover in much detail IMO is how we handle functionality changes in the upstream CMS. Should we make it clear that the chosen CMS must have an opt-in (not automatic) version upgrade system, so that changes to how the CMS works doesn't then break the publishing pipeline?

(WordPress, for example, can be configured to auto-update, or be manually updated through the UI. I assume we'd prefer the latter, so that we're able to test any changes required in advance).

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
kevindew: Thanks for your points, I'm thinking the simplest thing is perhaps to defer asset handling for now and to have it as a limitation - perhaps just not include assets in JSON payloads so they can't be snuck in.

> Linking directly to cfassets.com also not ideal, I'm thinking of people who may have script blockers where they've allow-listed GOV.UK but not third party providers. A small point, but worth considering.

Would we expect that for images / assets? Do we have evidence of these existing?

> The most important point IMO is the migration strategy - it seems highly likely we'd eventually move away from whichever CMS is picked for this experiment, and then be left with a migration job of having to copy all the images to GOV.UK's asset manager and rewrite all the affected content in Publishing API to refer to the new image names.
We also may not actually want to utilise the dynamic image sizing functionality in the Contentful Images API, as it would likely increase the complexity of the integration with GOV.UK's stack.

This doesn't concern me too much as I'm expecting the initial footprint to be pretty low, so relatively solvable should we make the decision to drop/change.

> One thought that came to mind was building some sort of simple asset uploader CMS, where publishers can upload assets directly to Asset Manager, and then copy the resulting asset URL into the headless CMS. A bit more clunky, perhaps, but less of a headache when it comes to migration.
Would need to think about how to handle alt text - the asset uploader CMS feels like a reasonable fit for that, but it would be another thing to have to copy and paste, so probably best handled within the headless CMS.

This is a good idea, but I imagine it'd be less work to set-up a system that syncs a CMS assets with Asset Manager so would probably go that way. I imagine it'd be less confusing too than having to forgo a CMS assets features.

Manually uploading assets does seem a reasonable workaround though in the early days if this feature is needed.

kevindew:
> Interesting proposal! A few typos and a comment or two below.

Thanks so much for the comments and fix suggestions ⭐ 

> The only thing the RFC doesn't cover in much detail IMO is how we handle functionality changes in the upstream CMS. Should we make it clear that the chosen CMS must have an opt-in (not automatic) version upgrade system, so that changes to how the CMS works doesn't then break the publishing pipeline?

I see what you mean, though not sure it's universally applicable - as it's not a SAAS concern, we expect them to be updated constantly. If there was a point to add it'd be probably something like: "By integrating with a CMS, particularly a SAAS one, we will have an obligation to adapt to changes made to their APIs." - however that seems a little left of what you were aiming.

I think your point sounds more applicable if/when we consider the consequences of self-hosting a CMS.

maxf:
I'm writing this general comment on behalf of the CPTO Data Services teams.

We currently run several tools that search and analyse GOV.UK content, and are currently designing more. Some of those tools are important to the work of content designers, data analysts, and data scientists. 

Since those tools mostly run on GCP, we replicate databases like the Content Store and the Publishing API from backups copied from S3. Since we want to be able to copy as much of the GOV.UK content as we can (to run tools like GovGraph Search or Content Data, which lets content designers search for keywords throughout all of gov.uk), we need to have access to as much content as we can. Currently we don't have access to all the content (eg smart answers) and that's sometimes proving to be a problem for our users.

With this context in mind, we are fairly neutral regarding the introduction of a CMS as a new rendering app, as long as we have access to any GOV.UK content it would store. Obviously if the data from the CMS was put in the publishing pipeline, that would be ideal for us. Failing that, it would be ok for us if the data wasn't in the pipeline but elsewhere, as long as we are able to access it in whatever shape or form to include to our own data stores.

Hope this helps.

kevindew:
Thanks for all the comments. I've just pushed a commit with the suggested amends. I've modified the further considerations to state that we won't initially integrate with [Contentful Media](https://www.contentful.com/help/working-with-media/) to defer decisions around Asset Manager integration (there is not an expectation of media on the initial page, [cost-of-living](https://www.gov.uk/cost-of-living))

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Further considerations
+
+- If the CMS was offline and we needed to make edits to content this would be done by a short term hardcoding in the frontend application to set content.
+- The initial integration will not integrate with [Contentful Media](https://www.contentful.com/help/working-with-media/) as we have yet to decide whether these would need to be hosted on a GOV.UK domain or not.
+- Should we change CMS we would retire contentful-listener-api in favour of an appropriate one, or rename contentful-listener-api to be cms-listener-api to be a generic tool that serviced multiple CMS tools.
+- Should the result of the experiment be that we will not continue with an off-the-shelf CMS the Publishing team will revert all changes made to the stack.
```
brucebolt, line 59: I wonder if we also need to detail the exit strategy of this experiment (particularly if it isn't successful), as all of the options will involve some considerable work for a publishing team?

From what I understand, one of these four options will occur at the end of the experiment:
1. We switch off all the content that was created by this tool, in which case we just revert all the changes and have to implement redirects (I see this as being unlikely).
2. The page remains live but it is decided it should be edited within an existing GOV.UK Publishing tool, which will require some type of migration process that will have to be built.
3. We have to maintain the Contentful system forevermore, so the content it has previously published can continue to be edited, increasing the app ownership and support burden for one of the publishing teams.
4. We go through a procurement process and select a tool other than Contentful, in which case the intermediary code will need updating and all content (and change history?) migrated to the alternative tool.

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
leenagupte: I'm not sure if the coronavirus_landing_page is a good example. I think we used [Specialist Route Publisher](https://github.com/alphagov/special-route-publisher/pull/84/files) to publish the landing page, but its schema was never `special_route` but `coronavirus_landing_page`. 

Also there's a niggle in the back of my brain that says special routes by default aren't indexed by search. I haven't been able to find concrete evidence of this though, just [this entry](https://github.com/alphagov/search-api/blob/babb620cd2918ff3fc918fce5c21a07ce9b3ecd8/config/govuk_index/migrated_formats.yaml#L320), which seems to suggest only particular pages are not indexed. But perhaps the indexable-ness of special routes should be confirmed?

rfc-154-integrating-an-ots-cms-with-govuk-for-custom-pages.md
kevindew: Thanks for the comment @brucebolt 👍 

I think the most likely scenario, should the experiment not be successful, is not one you flagged above. Which would be hardcoding the content in the rendering application. For example, for the initial test page - [cost of living](https://www.gov.uk/cost-of-living) we'd revert the use of the CMS back to managing the content through [a config file](https://github.com/alphagov/collections/blob/main/config/cost_of_living_landing_page/content_item.yml)

Should the experiment be successful I expect option 3 or 4 would occur. In the case of option 3, I agree it increases support burden I'm somewhat hoping it'd be relatively low (think HMRC Manuals API) but an increase none the less. Should we go for option 4, adapting the listener system for a different API would indeed be extra work. I'd expect the quantity of content (given the limited trial) to be low though and wouldn't expect any history migration.

I'll add a note about the exit back to hardcoding.

brucebolt: That sounds good to me, thanks for confirming 👍

leenagupte:
>So if you need to add or remove items you will have to edit their parent item for that action. Thus you will still be notified of the change.
>
>I.e if we had cost of living page with ID e1, that had two accordions (e2 and e3).
>
>If we edited e2 we’d be able to determine that we need to update the cost of living page as we’d find e2 as an ID in the Publishing API for the cost of living page.
>
>If we added a new accordion (e4) to the page, we’d have to edit e1 to add the accordion and would be notified that e1 had changed - so it wouldn’t matter that we weren’t tracking e4 at that stage.
>
>If we removed an accordion e3 from the page, there’s two ways that could happen 1) we delete the entity e3 from the CMS or 2) we edit the page (e1) to remove it from the collection. Both of those would produce different webhooks but as we track both e3 and e1 we still get the page updated.

This is a really good explanation. It would be great if it could be documented somewhere.

Sorry if this has been asked already, but in the example above would the new accordion (e4) need be added to the configuration in Contentful and the listener first? Or would say a content designer be able to added as many accordions to e1 as they please?

Will the content item need to configured in two places (Contentful and the listener) or would a webhook be used in the listener to get config changes from Contenful? I'm trying to understand what the support burden would be and if it would be as fiddly as adding/updating specialist documents in Specialist Publisher.

That being said, I really like this plan, especially the idea of using a listener to integrate with publishing-api so we get the benefits of integrating with search-api and how frontend renders content items won't need to change much.

kevindew:
> This is a really good explanation. It would be great if it could be documented somewhere.

Yup, there's most of it documented in https://github.com/alphagov/contentful-listener-api/blob/main/docs/how-this-application-works.md but some more detail could be added.

> Sorry if this has been asked already, but in the example above would the new accordion (e4) need be added to the configuration in Contentful and the listener first? Or would say a content designer be able to added as many accordions to e1 as they please?

They would be able to add as many accordions as they wanted.

The listener would only need to be aware of e1 and would automatically pull through any data that is flagged as a child of e1.

> Will the content item need to configured in two places (Contentful and the listener) or would a webhook be used in the listener to get config changes from Contenful? I'm trying to understand what the support burden would be and if it would be as fiddly as adding/updating specialist documents in Specialist Publisher.

It's probably easiest to link to a code example to demonstrate this. This is [the config](https://github.com/alphagov/contentful-listener-api/blob/4cbbf7c0b12a2b8b040058809113e91a0bb28fdc/config/content_items.yaml#L10-L16) that the listener would have 

```
# Cost of living hub page
- contentful_space_id: q15b851vpa2y
  contentful_entry_id: Jid87gJxnMjLQlww1a4WF
  content_id: b8c8daed-f783-46d4-b6a6-9f22c38defda
  publishing_api_attributes:
    base_path: /cost-of-living-contentful
    rendering_app: collections
```

Which defines data about the resulting GOV.UK page that Contentful wouldn't have. What this config means is that if the Contentful entry `Jid87gJxnMjLQlww1a4WF` in [space](https://www.contentful.com/help/spaces-and-organizations/) `q15b851vpa2y` changes a GOV.UK Content Item with content id `b8c8daed-f783-46d4-b6a6-9f22c38defda` will be created or updated. If you referenced another Contentful entry in `Jid87gJxnMjLQlww1a4WF` this would be pulled into the data for the GOV.UK Content Item as the references of `Jid87gJxnMjLQlww1a4WF` are traversed.

The support burden of configuration would be low for a number of single pages (probably similar in quantity to what we have defined as special routes). However this strategy wouldn't be ideal if we wanted to be able to automatically create new pages through Contentful (if it acted as a place to create a particular format) and would need to think of a new strategy to glue the concepts together.

leenagupte:
> > Will the content item need to configured in two places (Contentful and the listener) or would a webhook be used in the listener to get config changes from Contenful? I'm trying to understand what the support burden would be and if it would be as fiddly as adding/updating specialist documents in Specialist Publisher.
> 
> It's probably easiest to link to a code example to demonstrate this. This is [the config](https://github.com/alphagov/contentful-listener-api/blob/4cbbf7c0b12a2b8b040058809113e91a0bb28fdc/config/content_items.yaml#L10-L16) that the listener would have
> 
> ```
> # Cost of living hub page
> - contentful_space_id: q15b851vpa2y
>   contentful_entry_id: Jid87gJxnMjLQlww1a4WF
>   content_id: b8c8daed-f783-46d4-b6a6-9f22c38defda
>   publishing_api_attributes:
>     base_path: /cost-of-living-contentful
>     rendering_app: collections
> ```
> 
> Which defines data about the resulting GOV.UK page that Contentful wouldn't have. What this config means is that if the Contentful entry `Jid87gJxnMjLQlww1a4WF` in [space](https://www.contentful.com/help/spaces-and-organizations/) `q15b851vpa2y` changes a GOV.UK Content Item with content id `b8c8daed-f783-46d4-b6a6-9f22c38defda` will be created or updated. If you referenced another Contentful entry in `Jid87gJxnMjLQlww1a4WF` this would be pulled into the data for the GOV.UK Content Item as the references of `Jid87gJxnMjLQlww1a4WF` are traversed.
> 
> The support burden of configuration would be low for a number of single pages (probably similar in quantity to what we have defined as special routes). However this strategy wouldn't be ideal if we wanted to be able to automatically create new pages through Contentful (if it acted as a place to create a particular format) and would need to think of a new strategy to glue the concepts together.

I looked at the config in the listener app and completely missed the `contentful_space_id` and `contentful_entry_id`.  This makes more sense now, thanks!

kevindew:
Thanks for the comments everyone, the deadline for them as now past. 

As all issues raised have been resolved I'm considering this accepted and going to seek an approval from another member of senior tech.

richardTowers (approved):

PR #156, opened 2023-01-26
RFC-156: Allow auto-merging of 'internal' Dependabot PRs

ChrisBAshton:
Proposal to auto-merge Dependabot PRs for internal dependencies (under strict conditions) to ease burden of keeping code up to date.

[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/auto-merge-internal-prs/rfc-156-auto-merge-internal-prs.md) 💅 

Trello: https://trello.com/c/rMDuLxbf/3073-evaluate-auto-merge-spike-and-write-rfc

### DIFF
--- a/rfc-156-auto-merge-internal-prs.md
+++ b/rfc-156-auto-merge-internal-prs.md
@@ -89,26 +89,23 @@ Note that this does put extra onus on getting our [semantic versioning][semver] 
 
 We want to limit the times at which auto-merging can take place, to typical office hours only. If a gem with a bug were to be released at 5pm on a Friday, this wouldn't be deployed into the apps until after most people have stopped working for the week, potentially creating work for those on-call if there is a user-facing impact.
 
-We propose a broad auto-merge window of between 9am and 4pm, Monday to Friday. Auto-merging must not happen outside of those times. In practice, the auto-merging will actually happen in a much narrower time period, around 9:30am, for reasons outlined below.
+We propose that auto-merging should happen around 9:30am, Monday to Friday, for reasons outlined below.
 
 We considered a [number of different ways][different-ways-to-defer-merging] of limiting the window during which auto-merging should happen. After [investigating][schedule-merge-investigation], there doesn't seem to be a reliable way auto-approving a Dependabot PR and then scheduling the merging of that PR to happen during office hours.
 
-Comparatively, it's trivial to [configure Dependabot to run at certain times][dependabot-configure-timing]. Currently, we set no such preference, so Dependabot often [raises PRs outside of office hours][example-dependabot-pr-out-of-hours]. We therefore propose configuring Dependabot to check for new versions at 9:30am every weekday. The proposed auto-merge implementation relies on running a number of GitHub Actions whenever a Dependabot PR is opened, so this should comfortably restrict auto-merging to happen only in office hours.
+Comparatively, it's trivial to [configure Dependabot to run at certain times][dependabot-configure-timing]. Currently, we set no such preference, so Dependabot often [raises PRs outside of office hours][example-dependabot-pr-out-of-hours]. We therefore propose configuring Dependabot to check for new versions at 9:30am, Monday to Friday. The proposed auto-merge implementation relies on running a number of GitHub Actions whenever a Dependabot PR is opened, so this should comfortably restrict auto-merging to happen only in office hours.
 
 There is one last consideration: Bank Holidays. We want to avoid auto-merging on Bank Holidays as these would be outside of office hours.
 
 Bank Holidays are typically Mondays or Fridays, so we could restrict Dependabot's configuration further, to only raise Dependabot PRs between Tuesday and Thursday. However, this would introduce unnecessary delays for sometimes very important (security) fixes, and also glosses over the fact that some Bank Holidays around Christmas might fall on a different weekday.
 
-Therefore, we propose building a [feature flag][] (for simplicity, a file with contents of either `0` or `1`) to make it easy to turn off auto-merging. The intention is that a developer would disable auto-merging on the last working day before a bank holiday, and then re-enable auto-merging upon their return. The feature has been successfully [spiked][spike-feature-flag] on a temporary repo, but for GOV.UK, the [flag file][] should live in [govuk-saas-config][].
+Therefore, we propose writing an additional `validate_not_a_bank_holiday` step into the auto-merge GitHub Action, which will use [GOV.UK's Bank Holidays API][bank-holidays-api] to ensure that the current date is not a bank holiday. If, for example, Dependabot runs on a Monday that happens to also be a Bank Holiday, the auto-merge would not happen and a developer would need to merge the PR manually at a later date.
 
+[bank-holidays-api]: https://www.gov.uk/bank-holidays.json
 [dependabot-configure-timing]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduletime
 [different-ways-to-defer-merging]: https://github.com/alphagov/govuk-rfcs/pull/156#issuecomment-1427552572
 [example-dependabot-pr-out-of-hours]: https://github.com/alphagov/content-data-admin/pull/1176
-[feature flag]: https://martinfowler.com/articles/feature-toggles.html
-[flag file]: https://github.com/ChrisBAshton/test-auto-merge/blob/9d6ed1518c39ef8b0e004c0e389a11898fffea02/feature-flags/auto-merge-dependabot
-[govuk-saas-config]: https://github.com/alphagov/govuk-saas-config
 [schedule-merge-investigation]: https://github.com/alphagov/govuk-rfcs/pull/156#issuecomment-1431282282
-[spike-feature-flag]: https://github.com/ChrisBAshton/test-auto-merge/blob/9d6ed1518c39ef8b0e004c0e389a11898fffea02/.github/workflows/dependabot-auto-merge.yml#L32-L44
 
 ## Actions
 
--- a/rfc-156-auto-merge-internal-prs.md
+++ b/rfc-156-auto-merge-internal-prs.md
@@ -93,7 +93,7 @@ We propose that auto-merging should happen around 9:30am, Monday to Friday, for 
 
 We considered a [number of different ways][different-ways-to-defer-merging] of limiting the window during which auto-merging should happen. After [investigating][schedule-merge-investigation], there doesn't seem to be a reliable way auto-approving a Dependabot PR and then scheduling the merging of that PR to happen during office hours.
 
-Comparatively, it's trivial to [configure Dependabot to run at certain times][dependabot-configure-timing]. Currently, we set no such preference, so Dependabot often [raises PRs outside of office hours][example-dependabot-pr-out-of-hours]. We therefore propose configuring Dependabot to check for new versions at 9:30am, Monday to Friday. The proposed auto-merge implementation relies on running a number of GitHub Actions whenever a Dependabot PR is opened, so this should comfortably restrict auto-merging to happen only in office hours.
+Comparatively, it's trivial to [configure Dependabot to run at certain times][dependabot-configure-timing]. Currently, we set no such preference, so Dependabot often [raises PRs outside of office hours][example-dependabot-pr-out-of-hours]. We therefore propose configuring Dependabot to check for new versions at 9:30am, Monday to Friday. The proposed auto-merge implementation relies on triggering a webhook whenever a Dependabot PR is opened, so this should comfortably restrict auto-merging to happen only in office hours.
 
 As an aside: setting the `schedule.day` and/or `schedule.time` configs [does not prevent security PRs from being raised][schedule-security] outside of those times. This is a good thing, and regardless, an external dependency update would not be auto-merged, even if it were a security one (that's perhaps a discussion for another RFC!).
 
@@ -101,7 +101,7 @@ There is one last consideration: Bank Holidays. We want to avoid auto-merging on
 
 Bank Holidays are typically Mondays or Fridays, so we could restrict Dependabot's configuration further, to only raise Dependabot PRs between Tuesday and Thursday. However, this would introduce unnecessary delays for sometimes very important (security) fixes, and also glosses over the fact that some Bank Holidays around Christmas might fall on a different weekday.
 
-Therefore, we propose writing an additional `validate_not_a_bank_holiday` step into the auto-merge GitHub Action, which will use [GOV.UK's Bank Holidays API][bank-holidays-api] to ensure that the current date is not a bank holiday. If, for example, Dependabot runs on a Monday that happens to also be a Bank Holiday, the auto-merge would not happen and a developer would need to merge the PR manually at a later date.
+Therefore, we propose writing an additional "validate this is not a bank holiday" step into the auto-merge process, which will use [GOV.UK's Bank Holidays API][bank-holidays-api] to ensure that the current date is not a bank holiday. If, for example, Dependabot runs on a Monday that happens to also be a Bank Holiday, the auto-merge would not happen and a developer would need to merge the PR manually at a later date.
 
 [bank-holidays-api]: https://www.gov.uk/bank-holidays.json
 [dependabot-configure-timing]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduletime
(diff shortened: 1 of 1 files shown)


brucebolt (commented):
This RFC makes me very happy, since the time wasted manually merging internal dependencies has always seemed to be excessively unnecessary.

The limited scope is reasonable too.

I wonder if we need to limit the times at which the auto-merging can take place to typical office hours only?  E.g. if a gem with a bug were to be released at 5pm on a Friday, this wouldn't be deployed into the apps until after most people have stopped working for the week, potentially creating work for those on-call if there is a user-facing impact.

BeckaL:
Echoing Bruce's comments - I think this will be good in reducing developer toil 🥳  but would want to limit the hours these can be merged in. It's rare, but I've previously seen a patch version bump cause an incident in a previous job, so I'd only want these to happen in work hours.

1pretz1:
Sounds good to me @ChrisBAshton, thanks for putting this together! +1 to the comments that this should only be in-hours.

AgaDufrat:
Great idea, which would save us time. All the conditions are making it feel safe and +1 for the in-hour only merging.

This would certainly increase the importance of correct versioning of our gems, which you mentioned in the description. We have documentation in the [dev docs](https://docs.publishing.service.gov.uk/manual/publishing-a-ruby-gem.html#versioning) and https://github.com/alphagov/govuk_publishing_components/blob/main/docs/publishing-to-rubygems.md but not all gem repos have this info.  Incorrectly versioning could be a point of failure.

Maybe we could expand on it the docs or have a refresher session/discussion (no, I’m not volunteering myself 🙈) including what would be considered as a breaking change for our internal gems. For a public facing gem a removal of something or variable type change would be classified as a breaking change, however we don't seem to treat it as such (for good reasons).

ChrisBAshton:
Thanks all! Completely agree about only merging in-hours - good catch. Just thinking through the implications...

If we look at [one example PR](https://github.com/alphagov/content-data-admin/pull/1176), it was raised at 9:56pm last Thursday, so we currently can't rely on Dependabot raising PRs in-hours. Based on the comments above, we'd want to defer the auto-merge until probably 9:30am the next day. To do this, I can think of a few options:

1) [Configure Dependabot to raise PRs at specific times](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduletime). But there's a risk that we'll have no protection for repos missing that config, or where devs have manually scanned for new Dependabot PRs.

It may also interfere with the Dependapanda, which IIRC caches the open PRs at a certain time in the night and presents these in the morning, so the results may always be a day behind. This is an issue with the architecture of the [Seal](https://github.com/alphagov/seal) though, more than anything - and its impact will be reduced by the effect of this RFC anyway.

2) Change the action to [run on a schedule](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule) instead of on a PR. I'm not sure that it would be compatible with the auto-approve-action - see [this fix](https://github.com/alphagov/govuk-developer-docs/commit/7d6e75cb124b3e6bc760a1b1637db8df8ea19c99) that was needed to run the action in the correct context. I'm also not sure what metadata will be available when run from a schedule as opposed to on PR, and how noisy it might be to have a scheduled action run for _every_ PR, Dependabot or not (albeit the [non-Dependabot ones would fail fast](https://github.com/alphagov/govuk-developer-docs/blob/316c6bbbe8ce2534772cdcbe0c4cee2d34166ad3/.github/workflows/dependabot-auto-merge.yml#L9-L13)).

3) Run action on `pull_request_target`, but find a way of deferring the 'merge' step of the action until in-hours. It [looks possible](https://dev.to/cardinalby/scheduling-delayed-github-action-12a6), but brittle.

In all cases, we'll also have to have a think about what to do in the edge cases, e.g. Bank Holidays.

I'll remove the deadline on this RFC to allow time to test out an approach, then push up a commit revising the RFC to accommodate the consideration for in-hours.

ChrisBAshton:
I've experimented with scheduling the merge time.

[Changing](https://github.com/ChrisBAshton/test-auto-merge/commit/5d793f394d0ea4ce63c2bec356d33abf539cb712) `on: pull_request` to `on: schedule: - cron: "30 9 * * 2-4"` didn't work, as this just [runs the auto-merge job on the `main` branch](https://github.com/ChrisBAshton/test-auto-merge/actions/runs/4182480078) (effectively a noop). You can [specify a branch](https://stackoverflow.com/questions/63612155/how-do-i-trigger-a-scheduled-action-on-a-specific-branch) instead of `main`, but it has to be hardcoded in the workflow file, so this didn't seem a viable approach.

I then tried utilising an action called [gr2m/merge-schedule-action](https://github.com/gr2m/merge-schedule-action) (via [this blog post](https://www.jasongaylord.com/blog/2020/07/31/schedule-merging-pull-requests-in-github)). Because its `on` property needs a `schedule`, I had to split out the auto-approve and auto-merge steps into [two separate workflows](https://github.com/ChrisBAshton/test-auto-merge/compare/cd6fa5c820171a5d14dd43dcad8c75b13dbb29be...d2ca3493aefa4ed949946e6a8bef148585fc93db). It's important that there is no room for auto-merging something that has not been auto-approved, so I [added a step](https://github.com/ChrisBAshton/test-auto-merge/blob/97a36aa413636f525ff77eee71e0c9c192554682/.github/workflows/dependabot-auto-merge.yml#L17-L27) that validates that the [auto-approve step](https://github.com/ChrisBAshton/test-auto-merge/blob/97a36aa413636f525ff77eee71e0c9c192554682/.github/workflows/dependabot-auto-approve.yml#L45-L52) succeeded. But the [lewagon/wait-on-check-action](https://github.com/lewagon/wait-on-check-action) action [doesn't seem to be compatible with being run via schedule](https://github.com/ChrisBAshton/test-auto-merge/actions/runs/4183492149/jobs/7247912836#step:10:174).

I'm sure with further hacks and tinkering, I could probably get something working, but it would be complicated and brittle. I'm therefore minded to look at [option 1 above](https://github.com/alphagov/govuk-rfcs/pull/156#issuecomment-1427552572) instead.

BeckaL:
Sounds sensible to configure dependabot times. Is there any way we can centrally configure this (govuk app config?? I say without knowing whether this is feasible or the right place)

rfc-156-auto-merge-internal-prs.md
```diff
@@ -0,0 +1,210 @@
...
+
+There is one last consideration: Bank Holidays. We want to avoid auto-merging on Bank Holidays as these would be outside of office hours.
+
+Bank Holidays are typically Mondays or Fridays, so we could restrict Dependabot's configuration further, to only raise Dependabot PRs between Tuesday and Thursday. However, this would introduce unnecessary delays for sometimes very important (security) fixes, and also glosses over the fact that some Bank Holidays around Christmas might fall on a different weekday.
+
+Therefore, we propose building a [feature flag][] (for simplicity, a file with contents of either `0` or `1`) to make it easy to turn off auto-merging. The intention is that a developer would disable auto-merging on the last working day before a bank holiday, and then re-enable auto-merging upon their return. The feature has been successfully [spiked][spike-feature-flag] on a temporary repo, but for GOV.UK, the [flag file][] should live in [govuk-saas-config][].
```
richardTowers, line 102: It feels risky to rely on developers manually enabling / disabling this on a per-bank holiday basis.

Could we have the github action check whether a day is a bank holiday using this cool bank holiday API I found online?

https://gov.uk/bank-holidays.json

rfc-156-auto-merge-internal-prs.md
```diff
@@ -0,0 +1,210 @@
...
+
+[govuk-semver-guidance]: https://docs.publishing.service.gov.uk/manual/publishing-a-ruby-gem.html#versioning
+[rails-major-upgrade]: https://docs.publishing.service.gov.uk/manual/keeping-software-current.html#rails
+[semver]: https://semver.org/
+
+#### 5. During office hours
+
+We want to limit the times at which auto-merging can take place, to typical office hours only. If a gem with a bug were to be released at 5pm on a Friday, this wouldn't be deployed into the apps until after most people have stopped working for the week, potentially creating work for those on-call if there is a user-facing impact.
+
+We propose a broad auto-merge window of between 9am and 4pm, Monday to Friday. Auto-merging must not happen outside of those times. In practice, the auto-merging will actually happen in a much narrower time period, around 9:30am, for reasons outlined below.
```
richardTowers, line 92: I wonder whether an earlier time on Friday might be sensible... or just leaving Friday out entirely because of its proximity to the weekend...

rfc-156-auto-merge-internal-prs.md
leenagupte: Not everyone has seen the api: https://github.com/jfi/bankholidays 😱

leenagupte: I had similar thoughts about the riskiness of a manual flag.

rfc-156-auto-merge-internal-prs.md
leenagupte: +1 to leaving out Fridays.

ChrisBAshton: @richardTowers the RFC currently proposes that Dependabot would run around 9:30am every weekday, i.e. auto merging should happen in the morning (assuming Dependabot runs at the specified time and that the GitHub Actions runs aren't delayed). I appreciate the wording of 9am-4pm doesn't help, here (I didn't see the need to constrain developers to only running Dependabot at 9:30am when we really only want to ensure it happens within office hours, but this does give unnecessary freedom and there's no obvious reason why they might want to schedule Dependabot to run later).

I've clarified the language so that it's explicitly proposing Dependabot running 9:30am Monday-Friday.

Both of you, let me know if you still think Fridays should be left out. Happy to apply that change to the RFC, but my two cents are that it's not particularly riskier than any other weekday and a bit arbitrary to drop (e.g. why not drop Thursday too?). We should have enough confidence in the other safeguards and the several hours of rollback time (if needed), IMO.

rfc-156-auto-merge-internal-prs.md
ChrisBAshton: Good idea. I've pushed up a commit to propose using the Bank Holidays API instead 👍

brucebolt (approved):
This looks good to me, thank you for investigating the time limiting @ChrisBAshton and also considering bank holidays.

Just adding it here as I'll otherwise forget: once this RFC has been implemented, we should  update the time Dependapanda posts to Slack so it is after Dependabot has run for the day and after CI builds would've completed (so the dependencies that can be auto-merged have already been processed), maybe 10:30am?

rfc-156-auto-merge-internal-prs.md
```diff
@@ -0,0 +1,210 @@
...
+
+Note that this does put extra onus on getting our [semantic versioning][semver] right. Before, it was just semantics, but if this RFC is approved, a developer's choice of versioning will determine whether or not a dependency will be automatically rolled out. That said, we do have existing [guidance on versioning our ruby gems][govuk-semver-guidance].
+
+[govuk-semver-guidance]: https://docs.publishing.service.gov.uk/manual/publishing-a-ruby-gem.html#versioning
+[rails-major-upgrade]: https://docs.publishing.service.gov.uk/manual/keeping-software-current.html#rails
+[semver]: https://semver.org/
+
+#### 5. During office hours
```
robinjam, line 88: Should this be further restricted to [core hours](https://intranet.cabinetoffice.gov.uk/task/hours-of-attendance-policy/) (10:00-16:00)?

sihugh (approved):
I think this is a good change, and could contribute to earlier deployment of changes and a decrease in the size of changes that are deployed. We'll have to advocate for the change to Alphagov's GitHub Actions policy as we don't own that outright.

heathd:
> There are some nuances to step 2. There is an [org-wide setting](https://github.blog/changelog/2022-01-14-github-actions-prevent-github-actions-from-approving-pull-requests/) to "Allow GitHub Actions reviews to count towards required approval", so it should be possible for the GitHub Actions bot to approve these PRs. However, we currently have that setting disabled, after a brief discussion in a [GDS-wide RFC](https://docs.google.com/document/d/1IFz7E4DcWJ09giNB38fxfccU6fdW4TrQi722zztmSxs/edit#heading=h.au70tiw6t1oo). In that RFC, it was acknowledged that enabling the feature would be "low risk", but as there were no identifiable use cases for it at the time, it was left disabled.

I think you've misread the original doc here.. the `[Low risk]` was referring to the "Possible unintended side effects" of the change proposed by the RFC. The change proposed was to _disable_ that setting. Ie. it's saying that there's a low risk of unintended side effects from disabling this setting. It's not saying that _having it enabled_ is low risk. 

The setting underlying issue is described in this [blog post](https://medium.com/cider-sec/bypassing-required-reviews-using-github-actions-6e1b29135cc7). I think the risk associated is not "low", partly because the attack vector is so surprising. It creates the possibility for an insider to self merge to main without further scrutiny.

heathd:
Just want to add, this is a really great initiative and I don't want to be a blocker here. Really excellent concrete proposal to remove toil that can be automated. To get it over the line I think we need a little more elaboration of a possible alternate option and a little more analysis of the risk/benefit tradeoff.

rfc-156-auto-merge-internal-prs.md
```diff
@@ -0,0 +1,210 @@
...
+- Who: Platform Reliability
+- When: After enabling the org-wide setting
+
+Platform Reliability intend to reference the homemade GitHub Action from [Packaging as a GitHub Action](#packaging-a-github-action), raising an initial config PR across all GOV.UK repos that represent continuously deployed apps. Whilst Platform Reliability would create the initial PRs, it would be up to the owning teams to decide whether or not to merge, and to what extent they'd want to reconfigure the repo and version scopes.
+
+Platform Reliability also intend to raise the necessary PRs to restrict Dependabot 'schedule times' as per [5. During office hours](#5-during-office-hours). As above, it will be up to the owning teams to decide whether or not to merge, though we will stipulate that they must not enable auto-merging without also restricting the Dependabot schedule times.
+
+The PR will be expanded from the initial spike, to include both patch and minor version bumps, of the following internal dependencies (but only where those dependencies are present in the repo):
+
+- `govuk_publishing_components` (18 versions - not including 3 major increases - since October 2022, across 32 apps)
```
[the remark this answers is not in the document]
ChrisBAshton: To clarify, do you mean changing the [standard Continuous Deployment pull_request_template](https://github.com/alphagov/content-publisher/commit/921150c0bb7629b2fcf85fc696f4292686035920) to something like this?

```diff
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

+ Note that some internal dependencies raised by Dependabot are automatically merged. See [workflows/dependabot-auto-merge.yml](./workflows/dependabot-auto-merge.yml) for details.
```

That seems fine to me, though it's a bit of a weird one, as developers will only ever see it out of context (i.e. PRs that auto-merge may not get looked at, PRs that don't auto-merge will all have this additional message that doesn't apply to them).

heathd (approved):
I'm happy with the latest proposal described in 4d55aac9237e54e557bdda85633d4b9eddc9f5f7

rfc-156-auto-merge-internal-prs.md
```diff
@@ -110,93 +110,81 @@ Therefore, we propose writing an additional `validate_not_a_bank_holiday` step i
...
+- `govuk_publishing_components` (18 versions - not including 3 major increases - between October 2022 and January 2023, across 32 apps)
+- `govuk_app_config` (6 versions - not including 1 major release - between October 2022 and January 2023)
+- `gds-api-adapters` (3 major versions between October 2022 and January 2023. 0 minor/patch ones. However, between July and August 2022 there were 4 qualifying patch increases we could have automated.)
+- `rubocop-govuk` (2 versions - 0 major ones - between October 2022 and January 2023, across a large number of repositories)
 
-The PR will be expanded from the initial spike, to include both patch and minor version bumps, of the following internal dependencies (but only where those dependencies are present in the repo):
+For `govuk_publishing_components` alone, we've had 576 PRs between October 2022 and January 2023, all of which could have been auto mergeable. If we estimate a minute of due diligence per PR, that amounts to almost 10 hours of wasted dev time.
```
sihugh, line 161: Sometimes this will be considerably more than a minute (speaking as someone who has, on occasion, been waiting for PRs to appear). Removal of the context switching penalty is also useful. We possibly need to be a little careful of ["is it worth the time"](https://xkcd.com/1205/), but I think the reduction in drudgery also has value.

I wonder also about whether we should look into a bulk "invoke dependabot check" button for deploys that might be less routine than waiting for the following day.

rfc-156-auto-merge-internal-prs.md
```diff
@@ -110,93 +110,81 @@ Therefore, we propose writing an additional `validate_not_a_bank_holiday` step i
...
+    - patch
+    - minor
+```
 
-### Review in 3 months
+### Raise initial config PRs
+
+Following a successful trial, Platform Reliability will raise PRs for all applicable GOV.UK repos. Each PR will configure Dependabot to run during office hours, and create the auto-merge configuration file.
+
+Whilst Platform Reliability would create the initial PRs, it would be up to the owning teams to decide whether or not to merge, and to what extent they'd want to reconfigure the repo and version scopes. We will stipulate that teams must not include external dependencies on the auto-merge list, and must not enable auto-merging without also restricting the Dependabot schedule times.
```
sihugh, line 181: It would be good to monitor configuration drift to indicate problems/improvements with this process.

rfc-156-auto-merge-internal-prs.md
```diff
@@ -205,6 +193,4 @@ If the trial is deemed successful, we will keep the GitHub Action configurations
 [Manage Ruby Dependencies]: https://docs.publishing.service.gov.uk/manual/manage-ruby-dependencies.html
 [The GDS Way]: https://gds-way.cloudapps.digital/standards/tracking-dependencies.html#update-dependencies-frequently
 
-If the trial is unsuccessful, we will remove all the GitHub Action configurations, disable the org-wide setting, and add an updated note to the top of this RFC.
-
-[spike]: https://github.com/alphagov/govuk-developer-docs/blob/68146cbddadb6adbf96fe3caaf15a3210ca66a37/.github/workflows/dependabot-auto-merge.yml
+If the trial is unsuccessful, we will remove all the configurations, terminate the auto-merge service, and add an updated note to the top of this RFC.
```
sihugh, line 196: 👍

sihugh (approved):
The less radical change incurs a bit more cost to us, but limits risk.

PR #157, opened 2023-02-28
RFC-157: Migrate from VCL-based CDN config to edge compute on Fastly & AWS

robinjam:
Proposal to move from VCL-based CDN config to edge compute on both Fastly ([Compute@Edge](https://www.fastly.com/products/edge-compute)) and AWS ([Lambda@Edge](https://aws.amazon.com/lambda/edge/)).

[✨ Rendered version](https://github.com/alphagov/govuk-rfcs/blob/edge-compute/rfc-157-edge-compute.md)

[Trello card](https://trello.com/c/29wRpINj/3091-write-rfc-proposing-at-edge-functionality-to-configure-our-cdns)


ChrisBAshton (commented):
Amazing work, @robinjam ⭐ Very thorough and exciting proposal.

Some suggestions and nitpicks below before we seek comments more widely.

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,211 @@
...
+
+Platform reliability would own most of this work.
+
+This project can be divided into a number of stages:
+
+1. Development
+    - We will develop & deploy the new Fastly services and the new CloudFront distribution separately from our existing CDN service.
+    - Both the new Compute@Edge service and the Lambda@Edge handlers should be built together, so that we don't end up painting ourselves into a corner with an architecture that's only suitable for one service.
+    - Both services must have access controls during development (I propose an IP allowlist restricting access to the office & VPN).
+    - The new Compute@Edge service (or rather, the VCL service behind it in the "sandwich"), as well as the new CloudFront distribution, must both be added to origin's IP allowlist, so that they are able to make requests to origin.
```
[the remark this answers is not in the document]
robinjam: Good point, Fastly's whole IP range is already covered, we'd only need to add the new CF distro

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,211 @@
...
+  ```
+
+On the Fastly side, we will also need:
+
+- A VCL service _before_ our Compute@Edge one so that we can continue to use shielding on the `/alerts` route (and potentially roll out shielding for all of GOV.UK), and
+- Another VCL service _after_ our Compute@Edge one so that we can manipulate backend responses before they hit the cache and/or control which backend responses are cacheable.
+
+In other words, we will need a VCL -> Compute@Edge -> VCL "sandwich". Such an approach is explicitly mentioned in the [Fastly docs](https://developer.fastly.com/learning/concepts/service-chaining/#chaining-more-than-two-services) as the only use case where a chain of more than 2 services is recommended.
+
+On the Lambda@Edge side, manipulation of the backend response before it hits the cache could be achieved by building an origin response handler. This would create code duplication between Lambda@Edge and the second VCL service in the "sandwich", but that can't be helped right now.
```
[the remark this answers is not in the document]
robinjam: I think I need to clarify this paragraph, but what I'm proposing is that the `GOVUK-Account-Session` and `GOVUK-Account-End-Session` handling should happen in a VCL service on the Fastly side, and an origin response handler on the AWS side. This creates duplication because we're doing essentially the same thing in both VCL and TypeScript. When Compute@Edge gains support for controlling caching around backend responses then we can refactor this to remove the duplication.

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,211 @@
...
+The primary benefit TypeScript provides over JavaScript is type safety, including enforcement of `readonly` properties. This will be particularly useful on the Lambda@Edge side because parts of the payload we receive from CloudFront are intended to be read-only. TypeScript enforces these invariants (potentially preventing us from introducing bugs), whereas JavaScript doesn't know how to.
+
+If folks have a strong aversion to TypeScript, JavaScript would also be a good option.
+
+## Architecture
+
+I propose that we build a three-layer architecture, with:
+
+- A platform-agnostic wrapper API to abstract away differences between the platform-specific types (e.g. `Request` and `Response`)
+- Platform-agnostic request and response handlers (corresponding to Lambda@Edge's viewer request and viewer response handlers concepts respectively), where the majority of the code lives, e.g.
```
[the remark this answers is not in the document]
robinjam: That's not quite right, I think I need to clarify a little. The C@E response handler might indeed contain logic, but it will be logic that corresponds to the viewer response handler on L@E (not the origin response handler).

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,211 @@
...
+
+I propose that:
+
+- Both the Compute@Edge and Lambda@Edge code should live in `govuk-cdn-config`.
+- We should use the [Fastly Terraform provider](https://registry.terraform.io/providers/fastly/fastly/latest/docs) (or possibly the Fastly CLI, if it supports all the features we need) to:
+  - Build the new Compute@Edge service into a WASI bundle and deploy it to Fastly
+  - Configure the two VCL services in the "sandwich"
+- A new Terraform project should be created in `govuk-aws` which:
+  - Pulls, bundles (using e.g. Webpack or esbuild), and packages (e.g. as a series of ZIP archives) the Lambda@Edge code
+    - It's important therefore that `govuk-cdn-config` have a regression test suite to ensure that the package is always in a buildable state, as otherwise we would be unable to deploy the corresponding Terraform project in `govuk-aws`
```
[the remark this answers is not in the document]
robinjam: Good point! The zips could actually still live in `govuk-aws` following the established precident; we could add a GitHub Actions workflow to `govuk-cdn-config` that builds the Lambda@Edge packages and then commits them to `govuk-aws` on each push to `main`. I think I'll propose that and see if folks agree/have a better suggestion.

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,236 @@
...
+This is the "traditional" way to configure Fastly services (see [high-level overview](https://developer.fastly.com/learning/vcl/using/)).
+
+[Our VCL](https://github.com/alphagov/govuk-cdn-config/blob/main/vcl_templates/www.vcl.erb) currently overrides the `RECV`, `FETCH`, `DELIVER` and `ERROR` subroutines, and handles a lot of things including (non-exhaustive):
+
+- IP allowlisting
+- Requiring HTTP basic auth on integration
+- A/B testing, including Replatforming's traffic split experiment logic
+- [JA3](https://github.com/salesforce/ja3) fingerprinting and denylisting
+- Enforcing SSL
+- Setting feature flag headers (e.g. `Govuk-Use-Recommended-Related-Links`, which enables recommended links in Whitehall)
```
sihugh, line 45: (Slightly off topic) I think we should look at removing/replacing this particular header. It's never been used as a toggle (after launch) and leaks into response headers that get sent to the client. If we want to use feature flags there are probably better ways of doing it now.

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,236 @@
...
+- Serving from static mirrors on HTTP 5xx
+- Requiring authentication for cache purge requests
+- Redirect for `/.well-known/security`
```
sihugh, line 48: (Slightly off topic) I think we should move this upstream and configure it as a short URL redirect instead. I can't remember why we didn't do this in the first place. The VCL caters for a few variants of this path, but that's fine.

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,236 @@
...
+
+We also make use of some features in the Fastly UI, which are not currently available within VCL (or indeed Compute@Edge code - see [Known limitations](#known-limitations)):
+
+- Enabling [shielding](https://docs.fastly.com/en/guides/shielding) for paths beginning with `/alerts`
+- Setting up logging to S3 and Splunk
+
+### Compute@Edge (Fastly)
+
+[Compute@Edge](https://www.fastly.com/products/edge-compute) (C@E) is Fastly's serverless edge compute product.
+
```
sihugh, line 65: Can be deployed using Terraform(?) - if that's desirable

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,236 @@
...
+Platform Reliability would own most of this work.
+
+This project can be divided into a number of stages:
+
+1. Development
+    - We will develop & deploy the new Fastly services and the new CloudFront distribution separately from our existing CDN service.
+    - Both the new Compute@Edge service and the Lambda@Edge handlers should be built together, so that we don't end up painting ourselves into a corner with an architecture that's only suitable for one service.
+    - Both services must have access controls during development (I propose an IP allowlist restricting access to the office & VPN).
+    - The new CloudFront distribution must be added to origin's IP allowlist, so that it is able to make requests to origin. The new Compute@Edge service (or rather, the VCL service behind it in the "sandwich") will [already be covered](https://docs.publishing.service.gov.uk/manual/cdn.html#fastly39s-ip-ranges-and-our-access-controls-on-origin-servers) by our existing allowlist.
```
sihugh, line 227: Would prohibit others from making a cloudfront distribution and pointing it at our origin?

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,236 @@
...
+2. A/B testing with live traffic by adding the new Compute@Edge service and CloudFront distribution as backends to the existing VCL-based service, and passing a percentage of traffic directly to each
```
sihugh, line 228: This will be slightly different from what would happen at failover. Is that OK?

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,236 @@
...
+  - This lack of feature parity presents us with several issues:
+    - Lack of clarity or confidence around what works in Cloudfront (e.g. developers have to be aware of this when building things, SMT need to be reminded of impact failover will have, we may be more hesitant in triggering the failover, etc.)
+    - Complexities also introduced downstream (e.g. if we failover and A/B testing stops working, data analysts now need to scrub out the failover period in their analysis)
+    - We're less likely to drill the failover in Production to ensure it works (as it will cause a somewhat degraded experience)
+    - It restricts our ability to consider a multi CDN strategy where traffic is split between two CDNs simultaneously 
+
+## Proposal
```
sihugh, line 22: Can we factor in some web perf testing?

rfc-157-edge-compute.md
```diff
@@ -0,0 +1,236 @@
...
+
+  // Lambda@Edge viewer response handler entrypoint:
+  function viewer_response_handler(response: Response): Response {
+    return handle_response(response)
+  }
+  ```
+
+On the Fastly side, we will also need:
+
+- A VCL service _before_ our Compute@Edge one so that we can continue to use shielding on the `/alerts` route (and potentially roll out shielding for all of GOV.UK), and
```
sihugh, line 176: [Cloudfront has a similar shielding concept](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/origin-shield.html), do we think this would work OK?

robinjam:
We (Platform Security & Reliability) are going to take this work in a different direction.

Much of the functionality that we currently handle at edge is probably better handled in other parts of the stack, e.g. at the application layer, our WAF (which is already CDN-agnostic). For (non-exhaustive) example:

- IP/JA3 allow/denylisting probably belongs in our WAF
- A/B testing, feature flags, cache control and stripping cookies probably belong in the application layer

Platform Security & Reliability are going to raise a separate RFC proposing these changes in more detail.

If we’re going to be moving the vast majority of this logic out of the CDN, then the work proposed within this RFC becomes significantly less valuable. We may instead want to consider maintaining equivalent but different CDN configurations for Fastly and Cloudfront, both of which are deployed using Terraform, and drop the use of VCL on the Fastly side. This work will also be covered in the aforementioned RFC.

We may wish to revisit edge compute at a later date, after the above work is completed, if we still have a significant amount of logic remaining at the CDN level.

PR #160, opened 2023-05-30
RFC-160 Use ES6 in govuk_publishing_components

patrickpatrickpatrick:
RFC-160 Use ES6 Modules in govuk_publishing_components

[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/aa16e2b5c0f32784dc83117dc0c90d1d4fbe73c8/rfc-160-using-es6-modules-in-govuk_publishing_components.md)

[Relevant Trello Card](https://trello.com/b/iACGoJN3/govuk-navigation-doing)


rfc-160-using-es6-modules-in-govuk_publishing_components.md
```diff
@@ -0,0 +1,428 @@
...
+    this.govukFrontendComponent.init();  // initialise the instance of the component
+    // specific init code for the gem component
+  }
+}
+
+export default GemComponent;
+```
+
+```
+window.GOVUK.GemComponent.init(exampleElement);
```
colinrotherham, line 121: Thanks for the detailed talk this morning @patrickpatrickpatrick 😊

As this is similar to our `window.GOVUKFrontend` global I'll add some of our findings here

We realised a while back that it was the Rollup UMD bundle that added globals for:

```mjs
window.GOVUKFrontend.Accordion
window.GOVUKFrontend.CharacterCount
window.GOVUKFrontend.Checkboxes
window.GOVUKFrontend.Radios
// etc
```

Whilst we're still going to publish UMD bundles in `govuk-frontend@5` we'd prefer if teams used ES6 module `import`/`export` syntax too, rather than put UMD bundles in a `<script type="module">` wrapper

### Before 
Rollup UMD bundles automatically add `window` globals

```html
<script src="/govuk-frontend.min.js"></script>
<script>
  window.GOVUKFrontend.initAll()
</script>
```

### After
Rollup ES bundles use ES6 `import`/`export` syntax

```html
<script type="module">
  import { initAll } from '/govuk-frontend.min.js'
  initAll()
</script>
```

Might be a while before you can move away from `window` globals

But by transpiling ES6 to ES5 (using [Bublé](https://github.com/bublejs/buble#readme)) you'd lose access to the good stuff™ like exports 😊

colinrotherham: I should add that ES module downloads will be deferred by default

So we'd recommend they still be "preloaded" with a script tag before import

```html
<script type="module" src="/govuk-frontend.min.js"></script>
```

Or using [`modulepreload` hints where supported](https://caniuse.com/link-rel-modulepreload)

```html
<link rel="modulepreload" href="/govuk-frontend.min.js">
```

patrickpatrickpatrick: With regards to transpiling at all, this is because frontend rendering applications don't have support for using ES6 modules as they are using sprockets and asset pipeline. If we were to make[ the significant change](https://github.com/alphagov/govuk-rfcs/blob/aa16e2b5c0f32784dc83117dc0c90d1d4fbe73c8/rfc-160-using-es6-modules-in-govuk_publishing_components.md#significant-change-move-all-javascript-to-a-separate-npm-module) (moving the Javascript outside of the gem entirely) then this would require all the frontend rendering applications to be updated to using bundlers so we could actually make use of modules and actual ES6 features in Rails applications. My thinking of this RFC was the least disruptive way to our frontend applications that would make govuk_publishing_components compatible with ES6.

rfc-160-using-es6-modules-in-govuk_publishing_components.md
```diff
@@ -0,0 +1,428 @@
...
+  input,
+  output: {
+    file: `app/assets/javascripts/govuk_publishing_components/components/${input.match(/(\w|-)+(?=\.js)/g)[0]}.js`,
+    format: "umd",
+    name: `GOVUK.Modules.${NameOfTheComponent}`,
+    sourcemap: true,
+  },
+  plugins: [
+    resolve(),
+    buble(),
```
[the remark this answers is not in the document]
patrickpatrickpatrick: This sounds like a more sensible way of doing things than using Bublé! In practice I will definitely use this!

rfc-160-using-es6-modules-in-govuk_publishing_components.md
colinrotherham: Definitely. Mainly commenting to show what's coming in future

>Might be a while before you can move away from `window` globals

But hopefully can avoid transpiling to ES5 anything inside ES2015 `<script type="module">`

rfc-160-using-es6-modules-in-govuk_publishing_components.md
```diff
@@ -0,0 +1,368 @@
...
+
+### Significant change: Move all Javascript to a seperate npm module
+
+The intermediary change to govuk_publishing_components will solve the upcoming problem we have of needing to be able to use the ES6 modules of the latest release of govuk-frontend. However it doesn't solve the issue of the developers of other applications being forced to use AssetPipeline and Sprockets to deliver Javascript. A standard practice in other gems is to seperate the Javascript into an npm package. This package then gets installed alongside the gem. If we made this separation, it would mean that the modules of the Javascript could be imported into projects and gives more flexibility to how developers can use govuk_publishing_components. If the changes above are made to how Javascript is built then it would not be much more work to move it to a seperate npm module (in theory) as we could use the same configuration. However maintaining a seperate package, rethinking how development would work with that package and then updating the workflow of all our frontend applications would take some work (and is perhaps best addressed in another RFC?).
```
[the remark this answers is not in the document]
colinrotherham: Have you considered [npm workspaces](https://docs.npmjs.com/cli/v9/using-npm/workspaces?v=true) to keep the new package locally, but symbolically linked into `node_modules` as if it was already published?

rfc-160-using-es6-modules-in-govuk_publishing_components.md
patrickpatrickpatrick: That would make sense actually, since it wouldn't run without polyfill anyway we could just strip out the transpiling step for components. Whether to include transpiling at all depends on if we just leave analytics as they are (which I think we could, unless we want to transpire them but that could start to introducing polyfill for them which I assume would want to be avoided). 

This would mean rethinking minification as Uglify is only compatible with ES5. There are alternatives out there (https://github.com/ahorek/terser-ruby) we could use so it's not a big issue. Would just mean integration would be more complex as we'd have to update the frontend rendering applications simultaneously with the changes to govuk_publishing_components.

rfc-160-using-es6-modules-in-govuk_publishing_components.md
```diff
@@ -0,0 +1,428 @@
...
+
+which would add
+
+`<script type="module" src="path/to/component"/>`
+
+similar to the format of `add_component_stylesheet("component-name")`. As the latest version of govuk-frontend is targeting browsers that support `type="module"` for components we use a script tag with the type module. Although analytics could also be individually loaded, [I would argue that for the intermediary change this is not required](#loading-analytics-code-individually).
+
+[As well as the ability to test the support of a script, there are also performance benefits to individual loading of Javascript. In the performance tests in the appendix, I examine the differences in page load time and asset size if we were to switch to individually loaded Javascript](#performance-tests-of-Individual-javascript-loading).
+
+### Significant change: Move all Javascript to a separate npm module
```
andysellick, line 181: I'm concerned about separating the JS out into a separate NPM package. You've highlighted some of the problems with this already (maintaining a separate package, how would local development work) but I think it's worth clarifying further.

Local development for frontend devs is already pretty complicated at times. For example (and I need to do this frequently) running a local application pointed at a local static, both of them pointing at a local version of `govuk_publishing_components`. This can be further complicated by any of those local apps randomly caching asset files, which leads to a lot of time wasted hunting through to delete them, or rebuild or restart applications. Adding another layer into this (a local version of all the JS for `govuk_publishing_components`) is going to make things more complicated.

It's also worth thinking about the process for making and deploying a change.

- make change to a component in `govuk_publishing_components` (1 PR)
- make change to component JS in JS package (+1 PR)
- make change in application and static to use component change (+2 PRs)
- make new version of JS package and new version of `govuk_publishing_components` to include it (+2 PRs)
- install new version of `govuk_publishing_components` in static and applications (+2 PRs, possibly more)

We're looking at potentially 8 pull requests for a single change, and even more if that change turns out to have an unforeseen bug.

We also have a problem (currently unsolved) in `govuk_publishing_components` of JS test files having to contain their own copy of the HTML that they operate on. If the JS and the component template are modified but the HTML in the test is not, it is possible to introduce errors into the code but still have the tests pass (again this has happened several times). Separating the JS and JS tests from the components they operate on still further by moving them out of `govuk_publishing_components` removes any chance of solving this problem.

I think all of this might come down to the question of whether separating out the JS into a separate package is something we _should_ do, or something we _have_ to do. 

If we have to do it then we have to, but this seems to be based on an assumption about the future of Rails that hasn't been confirmed yet. I'd also be interested to investigate if there are other approaches that could work for us even if that's what Rails decides.

As you mentioned, this issue might be worth a separate RFC.

colinrotherham: Can you keep the npm package in the same repo?

We use [npm workspaces](https://docs.npmjs.com/cli/v9/using-npm/workspaces) on GOV.UK Frontend with all our tools and [review app](https://govuk-frontend-review.herokuapp.com) as separate packages:

**package.json**
```json
"workspaces": [
  ".github/workflows/scripts",
  "packages/govuk-frontend",
  "packages/govuk-frontend-review",
  "shared/*",
  "docs/examples/*"
],
```

1. [**Publishable** packages in `./packages`](https://github.com/alphagov/govuk-frontend/tree/main/packages)
2. [**Internal-only** packages in `./shared`](https://github.com/alphagov/govuk-frontend/tree/main/shared)

Makes sure things like `import 'govuk-frontend'` hit the local package, not the registry, so no publishing needed. You'll notice when running `npm install` they'll be symbolically linked into `node_modules` too

But you get other [npm workspaces](https://docs.npmjs.com/cli/v9/using-npm/workspaces) features like:

### Build all workspaces
```console
npm run build --workspaces
```

### Build single workspce
```console
npm run build --workspace govuk-frontend
```

rfc-160-using-es6-modules-in-govuk_publishing_components.md
```diff
@@ -0,0 +1,428 @@
...
+- [This gem is coupled to sprockets/asset pipeline](https://github.com/alphagov/govuk_publishing_components/issues/505)
+
+(Thanks to Kevin for providing lots of insight with Rails, govuk_publishing_components and being very helpful throughout the research process!!)
+
+## Problem
+
+### The future of IE11 support on GOV.UK
+
+To support Internet Explorer 11 (IE11) all Javascript in govuk_publishing_components is written in the ES5 standard. This is because IE11 does not support ES6. However lately in the GOV.UK Frontend Community there have been calls to drop JS support for IE11. As of June 2022, [Microsoft no longer supports IE11](https://learn.microsoft.com/en-us/lifecycle/products/internet-explorer-11). From 24th of April to the 21st of May, the number of users that accessed GOV.UK using IE11 was 26,205. This is 0.067% of all users (38,821,598) who accessed GOV.UK during that time. The concensus in the community is that we should eventually drop support for IE11 but the timeline for this has yet to be established. Part of the complexity is that we would still want analytics to work for the not insigificant number of IE11 users.
```
maxgds, line 17: Data point: Between 10th August and 10th September '23 there were 18,841 users of IE, 18,704 of those on IE11. The next major release of the design system will drop js support for IE and is expected before the end of the year. The only pragmatic approach for GOV.UK to take is to fall in line for components inherited from the design system at which point supporting js for IE elsewhere becomes questionable in terms of effort:benefit ratio. The exceptions to this would be cookie acceptance and tracking.

rfc-160-using-es6-modules-in-govuk_publishing_components.md
kevindew: Just catching up on this RFC (super late) and thought I should add a comment here as I imagine this suggestion was a misunderstanding based on a suggestion I made.

The idea of using a separate repo was the misunderstanding, that should not have been a considered option. It should have been exactly what @colinrotherham described where the same repo publishes both a gem and an npm module. This is what Rails core gems do (for example: https://github.com/rails/rails/tree/main/activestorage).

Not that doing this is without headaches, but vastly less headaches than an idea of two separate codebases.

kevindew:
As I don't think we've seen any further activity on this in over 6 months, it seems prudent to close this now and resurrect it should it be continued further.

PR #162, opened 2023-07-26
Add line break in rfc-161-technical-direction-in-publishing-2023.md

xander-gds:


kevindew:
Thank you @xander-gds - I've applied that fix in https://github.com/alphagov/govuk-rfcs/commit/75a520efa36f5caf802c6e061efbca0eb164ddcd

PR #163, opened 2023-07-31
RFC-163: Simplify our CDN config, and standardise on Terraform for deployment

robinjam:
[Trello card](https://trello.com/c/3zDgJzZq/3218-write-rfc-proposing-retiring-fastly-vcl)

[✨ Rendered version](https://github.com/alphagov/govuk-rfcs/blob/cdn-rationalisation/rfc-163-cdn-rationalisation.md)

Deadline for comments: **Friday 25th of August**


rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,120 @@
+# Appendix: Proposal for relocating logic from the CDN layer to other parts of the stack
+
+Things that could be moved to WAF:
+
+- IP allowlisting on staging and production EKS[^ip-allowlist]
+- Requiring HTTP Basic auth on integration[^http-basic-1][^http-basic-2] (unless the user's IP is in the allowlist[^http-basic-allowlist])
+- IP denylisting[^ip-denylist]
+  - This functionality is currently unused (the dictionary that the denylist is read from is empty), but it exists in case we ever need to quickly block IP addresses (for example, during an incident), so we need to preserve it
```
ChrisBAshton, line 8: I would question whether we _need_ to preserve it. But definitely something to consider (post-RFC):

```suggestion
  - This functionality is currently unused (the dictionary that the denylist is read from is empty), but it exists in case we ever need to quickly block IP addresses (for example, during an incident).
```

rfc-163-cdn-rationalisation.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Proposal
+
+### Where possible, relocate logic from the CDN layer to other parts of the stack
+
+Please see the [appendix](rfc-163/relocating-logic-from-cdn.md) for specific examples of what functionality we are proposing to move, and where we are proposing to move it to. Note however that this is not an exhaustive list, and is subject to change; the main point is that we want to move as much functionality out of the CDN as possible, to make it easier to maintain.
+
+### Migrate our Fastly VCL to Compute@Edge
+
+Compute@Edge is Fastly's edge compute product, and their recommended approach for developing services moving forwards.
```
ChrisBAshton, line 38: I seem to remember reading something along these lines, but I can’t find a web page to back that up - can you provide a link if possible?

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,120 @@
...
+  - We could set up an HTTP 301 redirect using Router instead; CDNs and browsers should cache HTTP 301 responses
+
+[^redirect-security-txt-1]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L231-L233
+[^redirect-security-txt-2]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L606-L612
+
+Things that could be implemented in our applications:
+
+- Setting feature flags, such as showing recommended related links for Whitehall content[^whitehall-recommended-links]
+  - On the old platform it was necessary to implement these flags in the CDN layer to prevent the need to deploy a new release every time we wanted to enable/disable a feature. A header is set on the backend request to indicate to the application whether the feature is enabled or disabled.
+  - In the replatformed world, this becomes a lot easier: we can enable/disable features through environment variables (as opposed to headers), and then enabling or disabling becomes a matter of pushing a new commit to `govuk-helm-charts` and waiting a couple of minutes for Argo to pick it up.
```
ChrisBAshton, line 37: Worth linking to https://govuk-kubernetes-cluster-user-docs.publishing.service.gov.uk/manage-app/set-env-var/#update-an-ordinary-non-secret-environment-variable here?

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,120 @@
...
+[^whitehall-recommended-links]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L251
+
+Things that need to remain in our CDN (but become easier to implement/maintain if we migrate to Compute@Edge):
+
+- Require authentication for Fastly `PURGE` requests[^purge-auth]
+  - This doesn't need parity on Cloudfront
+- Sorting query string params[^sort-query] and removing Google Analytics campaign params[^remove-utm] to improve cache hit rate
+- Stripping query string params only for the homepage and `/alerts`[^remove-query]
+  - This appears to be a DDoS prevention measure(?) - should we expand this protection to other routes?
+- Automatic failover to static S3/GCS mirror if origin is unhealthy (only in staging and production)[^mirror-failover]
```
ChrisBAshton, line 49: Why only in staging and production?

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,120 @@
...
+
+Known issues with our current config that could be addressed more easily in Compute@Edge:
+
+- Currently if origin returns an HTTP 500, and we failover to the S3 mirror, but the requested path is not present in the mirror, the user receives an HTTP 403 and a very ugly XML-based error page
+  - This is expected behaviour: S3 returns a 403 if the file is missing and the access key that was used to make the request does not have the `s3:ListBucket` permission
+  - The fix is to intercept HTTP 403 responses _only from the S3 backend_, and replace them with a hardcoded error page - much easier to implement in Compute@Edge than in VCL
+
+Undecided/needs input from other developers:
+
+- Setting TLSversion request header for requests to Licensify for some reason[^licensify-tlsversion]
```
ChrisBAshton, line 105: This was removed in https://github.com/alphagov/govuk-cdn-config/pull/464.

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,120 @@
+# Appendix: Proposal for relocating logic from the CDN layer to other parts of the stack
```
ChrisBAshton, line 1: Really good job summarising each VCL feature here - thanks! ⭐

ChrisBAshton (commented):
Great work, @robinjam! Some comments below, and I'll leave an aside comment on the Trello card too.

rfc-163-cdn-rationalisation.md
```diff
@@ -0,0 +1,85 @@
...
+
+---
+
+> Currently, Compute@Edge does not support shielding, making service chaining a useful mechanism for adding shielding to C@E services
+>
+> -- <cite>https://developer.fastly.com/learning/concepts/service-chaining/#chaining-with-shielding-in-computeedge</cite>
+
+We're currently using shielding on `gov.uk/alerts`, so we would need to make use of service chaining to allow this to continue working when we migrate the rest of our logic to Compute@Edge.
+
+### Standardise on Terraform for CDN configuration
```
robinjam, line 79: It looks like @samsimpson1 from Platform Engineering is already looking into this, but is currently on leave - I will check in when he returns and update this section accordingly

rfc-163-cdn-rationalisation.md
robinjam: I suppose "their recommended approach" may be an exaggeration - their dev docs homepage points you towards Compute@Edge as opposed to VCL now, but I think both approaches are still supported. Will rephrase this!

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Redirecting `/security.txt` and `/.well-known/security.txt` to `https://vdp.cabinetoffice.gov.uk/.well-known/security.txt`[^redirect-security-txt-1][^redirect-security-txt-2]
+  - We could set up an HTTP 301 redirect using Router instead; CDNs and browsers should cache HTTP 301 responses
+
+[^redirect-security-txt-1]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L231-L233
+[^redirect-security-txt-2]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L606-L612
+
+Things that could be implemented in our applications:
+
+- Setting feature flags, such as showing recommended related links for Whitehall content[^whitehall-recommended-links]
```
sihugh, line 35: I think this should be removed. It's certainly not a pattern we want to repeat!

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,114 @@
...
+
+[^whitehall-recommended-links]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L251
+
+Things that need to remain in our CDN (but become easier to implement/maintain if we later migrate to Compute@Edge):
+
+- Require authentication for Fastly `PURGE` requests[^purge-auth]
+  - This doesn't need parity on Cloudfront
+- Sorting query string params[^sort-query] and removing Google Analytics campaign params[^remove-utm] to improve cache hit rate
+- Stripping query string params only for the homepage and `/alerts`[^remove-query]
+  - This appears to be a DDoS prevention measure(?) - should we expand this protection to other routes?
```
sihugh, line 48: Query string allow lists could be a feature in future. Not something to worry about immediately though.

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,114 @@
...
+[^http-basic-allowlist]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L154-L165
+[^ip-denylist]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L189-L192
+[^drop-requests-1]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L223
+[^drop-requests-2]: https://github.com/alphagov/govuk-cdn-config-secrets/blob/536de2171d17297c08a0a328df53a6b65002e2c4/fastly/fastly.yaml#L30-L39
+[^autodiscover-template]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L579-L603
+[^autodiscover-matcher]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L226-L228
+
+Things that could be implemented via Router:
+
+- Redirecting `/security.txt` and `/.well-known/security.txt` to `https://vdp.cabinetoffice.gov.uk/.well-known/security.txt`[^redirect-security-txt-1][^redirect-security-txt-2]
```
sihugh, line 27: I'm conscious that I've recently added this to some of our other CDN services (Bouncer, DGU etc.) Do we propose to move this in all of the services, or just GOV.UK?

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,114 @@
...
+[^redirect-security-txt-2]: https://github.com/alphagov/govuk-cdn-config/blob/55e587b238338caea1c7187c1f5d70cac8e5b104/vcl_templates/www.vcl.erb#L606-L612
+
+Things that could be implemented in our applications:
+
+- Setting feature flags, such as showing recommended related links for Whitehall content[^whitehall-recommended-links]
+  - On the old platform it was necessary to implement these flags in the CDN layer to prevent the need to deploy a new release every time we wanted to enable/disable a feature. A header is set on the backend request to indicate to the application whether the feature is enabled or disabled.
+  - In the replatformed world, this becomes a lot easier: we can implement feature flags through environment variables (as opposed to headers), and then enabling or disabling the feature becomes a matter of [updating the environment variables in `govuk-helm-charts`](https://govuk-kubernetes-cluster-user-docs.publishing.service.gov.uk/manage-app/set-env-var/#update-an-ordinary-non-secret-environment-variable) and waiting a couple of minutes for Argo to pick up the changes.
```
sihugh, line 37: 🙌

rfc-163-cdn-rationalisation.md
```diff
@@ -0,0 +1,39 @@
+# Simplify our CDN config, and standardise on Terraform for deployment
+
+## Summary
+
+- Where possible, relocate logic from the CDN layer to other parts of the stack
```
sihugh, line 5: Is this scoped just for the GOV.UK service in Fastly, or does it indicate where some of the others should go as well?

rfc-163-cdn-rationalisation.md
```diff
@@ -0,0 +1,39 @@
...
+- Standardise on Terraform for CDN configuration
```
sihugh, line 6: I think this is applicable across all of the services.

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,114 @@
...
+
+- IP allowlisting on staging and production EKS[^ip-allowlist]
+- Requiring HTTP Basic auth on integration[^http-basic-1][^http-basic-2] (unless the user's IP is in the allowlist[^http-basic-allowlist])
+- IP denylisting[^ip-denylist]
+  - This functionality is currently unused (the dictionary that the denylist is read from is empty), but it exists in case we ever need to quickly block IP addresses (for example, during an incident).
+- Silently ignore certain requests[^drop-requests-1][^drop-requests-2]
+  - This was the outcome of an [incident report](https://docs.google.com/document/d/12DzQsDeu7zUcICy9zVporjprX4qZFIrpOOWtYYRx-nk/edit) - details cannot be provided here, as this is a public repo
+- Serving an HTTP 404 response with a hardcoded template[^autodiscover-template] if the request URL matches `/autodiscover/autodiscover.xml`[^autodiscover-matcher]
+  - Context: https://github.com/alphagov/govuk-cdn-config/pull/86
+  - Blocking this from our WAF would remove the hardcoded page template, but no users are likely to visit this URL anyway
```
sihugh, line 13: We can use custom response bodies in AWS WAF rules 👍 : https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl.html#custom_response_body-block
They're hard coded too, unfortunately. It would be nice to read it from a file on a mirror or something

sihugh (commented):
Looks good! Just a few comments

rfc-163/relocating-logic-from-cdn.md
robinjam: Good shout, I haven't looked into the other services (should do so). It might make sense to leave this logic in the VCL then.

Alternatively, we could implement this via WAF for services which have one (www, bouncer, any others?), and leave it in the VCL for the ones that don't.

Not sure which option I prefer, to be honest 🤔

rfc-163-cdn-rationalisation.md
robinjam: Good point! So far I've only looked at the GOV.UK service, but I should probably expand the scope of this RFC to include all of them...

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,107 @@
+# Appendix: Proposal for relocating logic from the CDN layer to other parts of the stack
+
+Things that could be moved to WAF:
+
+- IP allowlisting on staging and production EKS[^ip-allowlist]
+- Requiring HTTP Basic auth on integration[^http-basic-1][^http-basic-2] (unless the user's IP is in the allowlist[^http-basic-allowlist])
+- IP denylisting[^ip-denylist]
+  - This functionality is currently unused (the dictionary that the denylist is read from is empty), but it exists in case we ever need to quickly block IP addresses (for example, during an incident).
```
richardTowers, line 8: I think there's a problem with these three around caching.

If the CDN has a page in its cache, then it won't make a request to AWS (where the WAF lives), so the IP allowlisting decision / basic auth / denylisting decision happens to late. For example, if a user with a banned IP requests a page that's cached, how is the CDN to know it shouldn't serve that to them?

We could work around this by disabling caching (probably not acceptable, even in non-prod environments as it would be a big disparity between the environments), or maybe by [doing something clever with cookies and the Vary header](https://www.fastly.com/blog/best-practices-using-vary-header#vary-cookie) (but that might end up being more complex / confusing than the current situation).

I think this is worth a bit of a discussion, because it's possibly a showstopper. It's quite possible I've got the wrong end of the stick though.

robinjam: @ChrisBAshton and I had a chat about this earlier and we actually think it's probably ok for denylisting to happen in the WAF.

If a denylisted user makes a request of any kind, they will be blocked by the WAF and get an HTTP 403 in response. Origin won't be hit, and client errors aren't (or at least, shouldn't be) cached by the CDN.

If a non-denylisted user makes a cacheable request (e.g. a `GET` to the homepage), they will get a 200 and this will be cached. If the denylisted user then makes the same request, they will receive the cached version (and again, origin won't be hit).

We think that's OK, because to our knowledge denylisting is intended to protect origin from attacks, rather than for access control.

Basic auth will still have to live in the CDN for the reason you mentioned (otherwise we'd have to add `Vary: Authorization` to the response, which would pretty much defeat caching).

What do you think @richardTowers ?

richardTowers: > denylisting is intended to protect origin from attacks, rather than for access control

It's true that it's not used for access control, but I'm not sure if that means it's just there to protect origin. It also gives us some useful protection at the CDN level. For example, if someone's downloading terrabytes of assets from one IP, Fastly might ask us if we could make them stop doing that (or if Fastly ever move us to usage based billing, we might want to make them stop), and I think IP denylisting is the only way we could do that at the moment.

> client errors aren't (or at least, shouldn't be) cached by the CDN

Definitely [citation needed] for this - I seem to remember 404s being cached (which is a reasonable choice for a CDN to make - there wasn't a resource here last time it was requested, there won't be a resource here this time). I'm not sure whether 403s get different treatment. We could fix that in the VCL for sure, if we haven't already (but then we're adding more code to the VCL to enable us to remove some code from the VCL, so we'd have to consider the cost / benefit of that).

> otherwise we'd have to add Vary: Authorization to the response, which would pretty much defeat caching

🤔  I wonder how bad this would be actually. Browsers don't do any kind of nonce / salting of Authorization headers, so they're always the same for the same username / password. Our basic auth use cases don't have a massive variety of username / password combinations, so I think we'd get a reasonable cache hit rate for them. Maybe that one could move to the WAF.

richardTowers: https://www.rfc-editor.org/rfc/rfc7231#section-6.5.4 lists a few 4xx responses as cacheable by default, but notably not 403s (404s are cacheable, as are 405s, 410s, and a few others). Whether Varnish implements this is another question...

richardTowers: https://github.com/varnish/Varnish-Book/blob/master/varnish_book.rst#the-initial-value-of-beresp-ttl

> Only the following status codes will be cached by default:

> 200: OK
> 203: Non-Authoritative Information
> 300: Multiple Choices
> 301: Moved Permanently
> 302: Moved Temporarily
> 304: Not modified
> 307: Temporary Redirect
> 410: Gone
> 404: Not Found

So not quite the same as RFC7231, but it won't cache 403s so you can ignore that argument :)

robinjam: > For example, if someone's downloading terrabytes of assets from one IP, Fastly might ask us if we could make them stop doing that (or if Fastly ever move us to usage based billing, we might want to make them stop), and I think IP denylisting is the only way we could do that at the moment.

I think we'd be more likely to use JA3 denylisting for this(?)

I'm actually not sure how valuable IP denylisting is as a general concept, given how easy it is for an attacker to get access to a bunch of IPv4 addresses, and given that IPv6 is a thing, but I'm open to being persuaded 😅

> > otherwise we'd have to add Vary: Authorization to the response, which would pretty much defeat caching
> 
> 🤔 I wonder how bad this would be actually. Browsers don't do any kind of nonce / salting of Authorization headers, so they're always the same for the same username / password. Our basic auth use cases don't have a massive variety of username / password combinations, so I think we'd get a reasonable cache hit rate for them. Maybe that one could move to the WAF.

Unfortunately I think it'd have to be more complicated than that, because we allow access to integration if _either_ the user provided the password _or_ their IP is in the allowlist. We'd have to do something like either:

- Set `Vary: Fastly-Client-IP, Authorization` on the response (which would essentially defeat caching), or
- Set `Vary: Authorization` always, and `Vary: Fastly-Client-IP` only if the `Authorization` header is missing (which actually might be a good shout, depending on how complex it'd be to implement in WAF 🤔)

I think I might tentatively move that back to the "move to WAF" section, with a caveat that it might prove to be too complicated.

robinjam: Thanks, that's useful context! I think we'll need to leave the denylist in the CDN, in that case (and see what we can do about ensuring parity between Fastly and Cloudfront)

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,114 @@
+# Appendix: Proposal for relocating logic from the CDN layer to other parts of the stack
+
+Things that could be moved to WAF:
+
+- Silently ignore certain requests[^drop-requests-1][^drop-requests-2]
+  - This was the outcome of an [incident report](https://docs.google.com/document/d/12DzQsDeu7zUcICy9zVporjprX4qZFIrpOOWtYYRx-nk/edit) - details cannot be provided here, as this is a public repo
+- Serving an HTTP 404 response with a hardcoded template[^autodiscover-template] if the request URL matches `/autodiscover/autodiscover.xml`[^autodiscover-matcher]
+  - Context: https://github.com/alphagov/govuk-cdn-config/pull/86
+- Redirecting `/security.txt` and `/.well-known/security.txt` to `https://vdp.cabinetoffice.gov.uk/.well-known/security.txt`[^redirect-security-txt-1][^redirect-security-txt-2]
```
theseanything, line 9: Could this even be done at the application level? I think there's potential we create ourselves another problem of being locked into using AWS WAF. Which would be an issue if we wanted to look at an alt cloud provider - (for example Disaster Recovery)

robinjam: That's a good point actually, I hadn't considered that we'd be moving essentially from one form of vendor lock-in to another.

My concern with e.g. the `security.txt` redirects is that we have the same functionality set up across several different domains (including gov.uk, assets.gov.uk, service.gov.uk and DGU). If we want to move this logic to the application layer then it'll presumably need to be implemented in a different way for each domain, whereas if we move it to the WAF then the implementation is consistent for each.

rfc-163-cdn-rationalisation.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Standardise on Terraform for CDN configuration
+
+The [custom Ruby scripts](https://github.com/alphagov/govuk-cdn-config/tree/main/lib) in [govuk-cdn-config](https://github.com/alphagov/govuk-cdn-config) should be replaced with Terraform projects.
+
+Platform Engineering have already begun work on migrating the WWW and Assets services to Terraform, with the new code living in [`govuk-fastly`](https://github.com/alphagov/govuk-fastly) and [`govuk-fastly-secrets`](https://github.com/alphagov/govuk-fastly-secrets). Once this work is complete, the Bouncer service, the service domain redirect service (which handles redirecting from https://service.gov.uk to https://www.gov.uk), and the TLD redirect service (which handles redirecting from https://gov.uk to https://www.gov.uk) should also be migrated to this new repo.
+
+The data.gov.uk Fastly services are [already deployed with Terraform](https://github.com/alphagov/govuk-aws/tree/main/terraform/projects/fastly-datagovuk) - we may want to consider migrating this code from `govuk-aws` to `govuk-fastly`, to keep all of our Fastly configuration in one place.
```
ChrisBAshton, line 41: Good suggestion 👍

rfc-163/relocating-logic-from-cdn.md
ChrisBAshton: I'd be inclined to do the security.txt redirect in Router, and consider the redirect on service.gov.uk and DGU to be out of scope for this RFC, as they're already not working:

- https://service.gov.uk/security.txt ->

rfc-163-cdn-rationalisation.md
```diff
@@ -0,0 +1,45 @@
...
+
+Across GDS we are standardising on infrastructure as code, and the use of Terraform to describe this infrastructure. Our Fastly service configuration has so far avoided this treatment, and is still deployed using custom Ruby code. We should address this by replacing our [handwritten Ruby scripts](https://github.com/alphagov/govuk-cdn-config/tree/main/lib) in [govuk-cdn-config](https://github.com/alphagov/govuk-cdn-config) with Terraform.
+
+## Proposal
+
+### Where possible, relocate logic from the CDN layer to other parts of the stack
+
+Please see the [appendix](rfc-163/relocating-logic-from-cdn.md) for specific examples of what functionality we are proposing to move, and where we are proposing to move it to. Note however that this is not an exhaustive list, and is subject to change; the main point is that we want to move as much functionality out of our CDN services as possible, to make them easier to maintain.
+
```
ChrisBAshton, line 34: I think it would be good to hit home that we don't want more VCL being written:

```suggestion

### Avoid adding new VCL unless absolutely necessary

See section above. Wherever possible, we should consider where else our logic might live, whether that's at the application layer, the Router layer or the WAF layer. Now that we've started using Terraform to manage our Fastly configuration, some features may also be able to be represented in Terraform rather than VCL.
```

ChrisBAshton (commented):
Hi @robinjam, apologies this took me longer than I'd have liked to get to!

I think we should get this merged in soon, but it seems a shame to merge it when a fair chunk of it is now somewhat outdated (the nature of longstanding RFCs!). I've added some comments below on areas I think would be good to update before merge. Thank you ⭐

sihugh (approved):

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,114 @@
...
+  - If we remove this functionality then we will also need to [update the docs](https://docs.publishing.service.gov.uk/manual/related-links.html#suspending-all-suggested-related-links).
+  - Regarding feature flags as a general concept:
+    - On the old platform it was necessary to implement these flags in the CDN layer to prevent the need to deploy a new release every time we wanted to enable/disable a feature. A header is set on the backend request to indicate to the application whether the feature is enabled or disabled.
+    - In the replatformed world, this becomes a lot easier: we can implement feature flags through environment variables (as opposed to headers), and then enabling or disabling the feature becomes a matter of [updating the environment variables in `govuk-helm-charts`](https://govuk-kubernetes-cluster-user-docs.publishing.service.gov.uk/manage-app/set-env-var/#update-an-ordinary-non-secret-environment-variable) and waiting a couple of minutes for Argo to pick up the changes.
```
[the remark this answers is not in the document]
robinjam: I think I'd rather leave it here, because that card was supposed to be an outcome of this RFC (they ended up getting delivered out of order because this RFC was held up waiting for approval 😅)

I also want to keep the context around feature flags in general

rfc-163/relocating-logic-from-cdn.md
```diff
@@ -0,0 +1,114 @@
...
+- Automatic failover to static S3/GCS mirror if origin is unhealthy or returns an HTTP 5xx (only in staging and production - in integration we want to be able to see errors as they happen)[^mirror-failover]
+- Stripping the `Accept-Encoding` header if the content is already compressed[^strip-accept-encoding]
+  - Context: https://github.com/alphagov/govuk-cdn-config/pull/379
+  - [Fastly already normalises the `Accept-Encoding` header](https://developer.fastly.com/reference/http/http-headers/Accept-Encoding/#normalization), but it doesn't automatically remove it if the requested content is already compressed
+- Controlling cache behaviour based on the `Cache-Control` header returned by origin[^cache-control]
+  - Manually set `Fastly-Cachetype` to `PRIVATE` if `Cache-Control: Private`[^cache-control-private]
+  - Explicit `pass` if `Cache-Control: max-age=0`[^cache-control-max-age]
+  - Explicitly `pass` if `Cache-Control: no-(store|cache)`[^cache-control-no-store]
+  - It is unclear which (if any) of these remain necessary if we decide to move to Compute@Edge (it's also unclear why Fastly doesn't respect them automatically 🤷‍♂️)
```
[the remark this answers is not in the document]
robinjam: I'll make a note to investigate this further (or write up a card for it) 👍

robinjam:
Thanks very much for the feedback everyone 🙏

rfc-163/relocating-logic-from-cdn.md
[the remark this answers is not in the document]
ChrisBAshton: Sounds good 👍

PR #164, opened 2023-09-28
RFC-164 Retire email-alert-monitoring as a separate app

KludgeKML:
Simplify monitoring of email alerts, move into email-alert-api, and retire email-alert-monitoring as a separate application.

[Rendered Version](https://github.com/alphagov/govuk-rfcs/blob/rfc-164/rfc-164-retire-email-alert-monitoring-as-a-separate-app.md)


rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Summary
+
+Replace the current email-alert-monitoring system with a task built into email-alert-api using the Notify API rather than the current Gmail/App based system, giving us a more stable and testable system and allowing us to retire a repo.
+
+## Problem
+
+email-alert-monitoring is a stand-alone app run as a cronjob (currently on Jenkins, although it's being moved over to EKS). It's designed to confirm that medical and travel alert emails have been sent out on time, and supports an SLA. It's clearly critical that we have some way of knowing that these alerts have gone out, but the current system has a number of problems:
+
+- It relies on a slightly Heath Robinson collection of google accounts
```
ChrisBAshton, line 11: This is a bit of an exaggeration, AFAICT. The only Google account the monitoring tool uses is `govuk_email_check@digital.cabinet-office.gov.uk`.

The reference to `gov.uk.email@notifications.service.gov.uk` in the README is unnecessary detail, it looks like that's just double-checked as the "FROM" address and no other checks are made.

I think we can delete this line as it doesn't really make the case for a problem.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+- Its method of matching alerts to emails is brittle
+- False alarms well outnumber valid alarms (which causes annoying on-call pages)
+- It's not entirely clear how actionable the valid alarms are
+
+## Proposal
+
+We should retire email-alert-monitoring as a separate application, and instead build a system into email-alert-api that can be called by a cron job (as is currently the case for the separate app).
+
+Currently it looks like we're not tracking the id for emails sent out via notify, but we do store email information for one day. By storing the notify ID, either for all emails (easier) or just for alert emails, we could then [query notify as to their current status]. Our cron job could query the current status for the relevant alert emails, and if _one_ email is in the "delivered" state, we have roughly as much proof that they're being delivered as we currently do.
```
ChrisBAshton, line 20: Let's be consistent with casing:

```suggestion
Currently it looks like we're not tracking the id for emails sent out via Notify, but we do store email information for one day. By storing the Notify ID, either for all emails (easier) or just for alert emails, we could then [query Notify as to their current status]. Our cron job could query the current status for the relevant alert emails, and if _one_ email is in the "delivered" state, we have roughly as much proof that they're being delivered as we currently do.
```

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+
+As a side-effect of adding this, we could also potentially query every alert email in the given period to determine how many of our alerts on average are getting delivered (and perhaps identify recipients who no longer exist).
+
+This isn't quite as end-to-end as the current system (if something was delivered, does that mean it's actually in a given person's inbox?), but since any record that Notify deems delivered is at the absolute boundaries of anything actionable by GOV.UK, there seems to be little to gain by tracking the alerts further than that.
```
ChrisBAshton, line 24: I agree with this thinking. But I think it's also worth keeping the existing `govuk_email_check@digital.cabinet-office.gov.uk` inbox and subscriptions. That way, if we get reports from departments/users that emails aren't being received, 2nd line can check if Notify is reporting that emails are being sent, and can _also_ verify that they've physically seen an email land in the inbox. We should just remove the monitoring around the account, as you're proposing 👍

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+
+[RSS feed]: https://www.gov.uk/drug-device-alerts.atom
+
+### How email-alert-monitoring currently works for Travel Advice Alerts
+
+The app is called as a cronjob once every 15 mintues. It opens a [healthcheck URL] on the travel-advice-publisher app, which contains a list of all travel alerts published in the last 2 days, but excluding the last 150 minutes. For these, it makes a list of the subject lines of the emails that should have been sent out, as well as the time (this is because more than one travel advice alert migt be issued for the same country in a two day period, so matching purely on the subject line isn't sufficient). It then uses the GMail API to read the inbox of govuk_email_check@digital.cabinet-office.gov.uk to check for the existence of those emails. If any are missing, an alert happens. Minimum time between publishing and alert is 2 and a half hours.
```
ChrisBAshton, line 38:
```suggestion
The app is called as a cronjob once every 15 minutes. It opens a [healthcheck URL] on the travel-advice-publisher app, which contains a list of all travel alerts published in the last 2 days, but excluding the last 150 minutes. For these, it makes a list of the subject lines of the emails that should have been sent out, as well as the time (this is because more than one travel advice alert might be issued for the same country in a two day period, so matching purely on the subject line isn't sufficient). It then uses the GMail API to read the inbox of govuk_email_check@digital.cabinet-office.gov.uk to check for the existence of those emails. If any are missing, an alert happens. Minimum time between publishing and alert is 2 and a half hours.
```

(Thanks for explaining the need for the time check!)

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+
+[healthcheck URL]: https://travel-advice-publisher.publishing.service.gov.uk/healthcheck/recently-published-editions
+
+### Problems with matching subject lines
+
+Problems can occur when trying to match by subject line if the title of the alert is altered after the email has been sent out. The [RSS feed] or [healthcheck URL] will contain a title that differs from the subject line of the email sent out. This will cause a failed match, and an alert will go off, even though for practical purposes the email has been sent. This causes false alarms, and needs someone to add a [hard-coded exception] into the matching code, increasing toil.
+
+[hard-coded execption]: https://github.com/alphagov/email-alert-monitoring/blob/main/lib/email_verifier.rb#L8-L36
+
```
ChrisBAshton, line 47: Might be worth adding a "Problems with matching timestamps" section too, since they add another vector for false positives into the mix - see https://github.com/alphagov/email-alert-monitoring/pull/106

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+We're rate limited in calls to Notify, so we'd need to make sure these calls were tied into the existing rate limiting service in email-alert-api
+
+We would need some way of matching the emails in the Email model to the emails expected to have been sent out. An MVP would be to use the exact same matching pattern (subject line and optionally time) that we use at the moment. This would get us the same level of reliability as now, but would retain the same problems (that if an alert's title is changed after it is sent out, the match will not work). To combat this, we could retain in Email records the content item that triggered them, and augment the [RSS feed] and [healthcheck URL] to add the content item into each displayed item. This is already publically available information and has no real security implication, but would require small changes to travel-advice-publisher and finder-frontend.
+
+[kept for one week, then deleted]: https://github.com/alphagov/email-alert-api/blob/main/app/workers/email_deletion_worker.rb#L5
```
ChrisBAshton, line 63:
```suggestion
[kept for one week, then deleted]: https://github.com/alphagov/email-alert-api/blob/032fe645e62b5d3b26fedba84a3e852a87575a8b/app/workers/email_deletion_worker.rb#L5
```

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+- A spam filter might redirect the message: Untestable, relies on idiosyncratic account setup, and out of our control
+- The recipient might ignore the message: Untestable, and out of our control.
+
+### How we could replace this with a job internal to email-alert-api
+
+We currently keep a record of every single email sent out via email-alert-api (the Email model). These records are [kept for one week, then deleted]. When emails are sent out, we could save the notify response ID in the Email record. This would allow us to get the [current delivery status] of the email using the Notify API. As long as one email sent in response to an alert has a delivery status of "delivered", we would have roughly the same assurance as we do now that the email alert has been sent successfully. It would also allow us the opportunity (if we wanted) to make a more stringent alert (say, that a certain percentage of the alert emails were delivered successfully) that we can't currently test.
+
+We're rate limited in calls to Notify, so we'd need to make sure these calls were tied into the existing rate limiting service in email-alert-api
```
ChrisBAshton, line 59: Can you link to this rate limiting service?

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+
+We would need some way of matching the emails in the Email model to the emails expected to have been sent out. An MVP would be to use the exact same matching pattern (subject line and optionally time) that we use at the moment. This would get us the same level of reliability as now, but would retain the same problems (that if an alert's title is changed after it is sent out, the match will not work). To combat this, we could retain in Email records the content item that triggered them, and augment the [RSS feed] and [healthcheck URL] to add the content item into each displayed item. This is already publically available information and has no real security implication, but would require small changes to travel-advice-publisher and finder-frontend.
```
ChrisBAshton, line 61: When you say "add the content item", do you mean just the content ID? 🙏

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+[hard-coded execption]: https://github.com/alphagov/email-alert-monitoring/blob/main/lib/email_verifier.rb#L8-L36
+
+### What Notify API can tell us about the status of an email
+
+Notify API gives us a [current delivery status] for an email, which begins as "sending" and is changed to "delivered" when Notify receives a callback from Amazon SES (the backing service Notify uses to delivery emails) to confirm the mailserver handling emails for the recipient address has accepted the email. This is a reasonably solid assurance that the email has got to the receipient, and is definitely at the point where further handling of the email is outside of our control - the only things that might stop the recipient seeing the email at that point are:
+- The mailserver might fail to deliver internally: Valid, but we can't currently test that unless the failure occurs in the exact GMail server handling our account anyway.
+- A spam filter might redirect the message: Untestable, relies on idiosyncratic account setup, and out of our control
+- The recipient might ignore the message: Untestable, and out of our control.
+
+### How we could replace this with a job internal to email-alert-api
```
ChrisBAshton, line 55: Probably worth explicitly calling out in this section what to do about 'batched' emails (daily, weekly).  People can still sign up to travel advice / medical advice on a daily/weekly cadence - do we bother alerting on whether or not those batched emails succeeded? (I'd assume "no", since, our current monitoring doesn't do that either, and to a lesser extent because the timeliness of the email is less important).

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+
```
ChrisBAshton, line 56: Also worth calling out in this section that travel advice and medical advice communicate directly with Email Alert API (as opposed to going through Email Alert Service), and that this is [recognised tech debt](https://trello.com/c/tWIZfxfc/25-travel-advice-publisher-and-specialist-publisher-talk-directly-to-email-alert-api).
The proposed changes to Email Alert API don't look mutually exclusive from a future where those publishing apps communicate with email alert service instead, so I don't think it's an issue, but worth being clear you've considered the implications of that tech debt being paid down.

ChrisBAshton (commented):
Thanks @KludgeKML, I love this idea. Some comments/tweaks below, but definitely in overall favour of the RFC.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
KludgeKML: Well, accounts is perhaps an oversimplified blanket term - it's not just the google account, it's the google cloud project that needs to be set up as well (and which there are no instructions for). I admit I've used prejudicial language here, so I'll revise, but trying to replicate this setup has cost a lot of dev and IT time, so I think it is definitely a selling point to get rid of it. (Also, see below for money-savings)

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
KludgeKML: I definitely think we should keep the existing subscriptions, but - we can move them from the current email address to an address backed by a google group rather than a full gmail account. We'd still be able to manually confirm that emails have got into the group, but:
- google groups are free, gmail accounts cost money, and
- google groups can be easily joined by any dev, and don't require an additional password hunt/login (reducing use of govuk-secrets)

I'll add this in.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+[healthcheck URL]: https://travel-advice-publisher.publishing.service.gov.uk/healthcheck/recently-published-editions
+
+### Problems with matching subject lines
+
+Problems can occur when trying to match by subject line if the title of the alert is altered after the email has been sent out. The [RSS feed] or [healthcheck URL] will contain a title that differs from the subject line of the email sent out. This will cause a failed match, and an alert will go off, even though for practical purposes the email has been sent. This causes false alarms, and needs someone to add a [hard-coded exception] into the matching code, increasing toil.
+
+[hard-coded execption]: https://github.com/alphagov/email-alert-monitoring/blob/main/lib/email_verifier.rb#L8-L36
```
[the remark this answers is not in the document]
KludgeKML: Ah, perma-links have tripped me up so many times I instinctively avoid them. But for RFCs, yes, definitely the right choice.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,76 @@
...
+We're rate limited in calls to Notify, so we'd need to make sure these calls were tied into the existing rate limiting service in email-alert-api
+
+We would need some way of matching the emails in the Email model to the emails expected to have been sent out. An MVP would be to use the exact same matching pattern (subject line and optionally time) that we use at the moment. This would get us the same level of reliability as now, but would retain the same problems (that if an alert's title is changed after it is sent out, the match will not work). To combat this, we could retain in Email records the content item that triggered them, and augment the [RSS feed] and [healthcheck URL] to add the content item into each displayed item. This is already publically available information and has no real security implication, but would require small changes to travel-advice-publisher and finder-frontend.
+
+[kept for one week, then deleted]: https://github.com/alphagov/email-alert-api/blob/main/app/workers/email_deletion_worker.rb#L5
+[current delivery status]: https://docs.notifications.service.gov.uk/ruby.html#get-the-status-of-one-message
+
+
+
+### Proposed Action Plan
```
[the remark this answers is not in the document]
KludgeKML: I'm assuming that it will actually be me doing the work, but for the sake of the RFC and flexibility, I'll mention that it's CIOP as a whole that owns it and would be responsible.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,94 @@
...
+
+As a side-effect of adding this, we could also potentially query every alert email in the given period to determine how many of our alerts on average are getting delivered (and perhaps identify recipients who no longer exist).
+
+This isn't quite as end-to-end as the current system (if something was delivered, does that mean it's actually in a given person's inbox?), but since any record that Notify deems delivered is at the absolute boundaries of anything actionable by GOV.UK, there seems to be little to gain by tracking the alerts further than that. We can keep the existing subscription to use in cases where we needed to check that emails have been physically seen in the inbox, but those checks can be done manually by second line (and the inboxes in question can be free google group inboxes rather than paid-for gmail accounts, since we won't need the API that gmail accounts have and google groups don't have. This would also simplify access, since we wouldn't need to store and retrieve secrets to access the gmail account when working on second line).
+
+[query notify as to their current status]: https://docs.notifications.service.gov.uk/ruby.html#get-the-status-of-one-message
```
ChrisBAshton, line 26:
```suggestion
[query Notify as to their current status]: https://docs.notifications.service.gov.uk/ruby.html#get-the-status-of-one-message
```

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Summary
+
+Replace the current email-alert-monitoring system with a task built into email-alert-api using the Notify API rather than the current Gmail/App based system, giving us a more stable and testable system and allowing us to retire a repo.
+
+## Problem
+
+email-alert-monitoring is a stand-alone app run as a cronjob (currently on Jenkins, although it's being moved over to EKS). It's designed to confirm that medical and travel alert emails have been sent out on time, and supports an SLA. It's clearly critical that we have some way of knowing that these alerts have gone out, but the current system has a number of problems:
+
+- It relies on a gmail account and a related google cloud console project that are hard to set up
```
ChrisBAshton, line 11: Nitpick: please use `Gmail` and `Google` throughout!

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,94 @@
...
+
+[healthcheck URL]: https://travel-advice-publisher.publishing.service.gov.uk/healthcheck/recently-published-editions
+
+### How the alerts actually happen (Jenkins)
+
+Where we use the phrase "an alert happens" above, the actual mechanism is that the email-alert-monitoring task exits with a fail status (2), and the Jenkins job running the task uses the Failure_Passive_check job to send a passive check to ICINGA with a failed status. Icinga handles escalation to Pagerduty.
```
ChrisBAshton, line 44:
```suggestion
Where we use the phrase "an alert happens" above, the actual mechanism is that the email-alert-monitoring task exits with a fail status (2), and the Jenkins job running the task uses the Failure_Passive_check job to send a passive check to Icinga with a failed status. Icinga handles escalation to PagerDuty.
```

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,94 @@
...
+
+[hard-coded execption]: https://github.com/alphagov/email-alert-monitoring/blob/79a865dcd8be07447735ae8ae99b78002241504a/lib/email_verifier.rb#L8-L36
+
+### What Notify API can tell us about the status of an email
+
+Notify API gives us a [current delivery status] for an email, which begins as "sending" and is changed to "delivered" when Notify receives a callback from Amazon SES (the backing service Notify uses to delivery emails) to confirm the mailserver handling emails for the recipient address has accepted the email. This is a reasonably solid assurance that the email has got to the receipient, and is definitely at the point where further handling of the email is outside of our control - the only things that might stop the recipient seeing the email at that point are:
```
ChrisBAshton, line 54:
```suggestion
Notify API gives us a [current delivery status] for an email, which begins as "sending" and is changed to "delivered" when Notify receives a callback from Amazon SES (the backing service Notify uses to delivery emails) to confirm the mailserver handling emails for the recipient address has accepted the email. This is a reasonably solid assurance that the email has got to the recipient, and is definitely at the point where further handling of the email is outside of our control - the only things that might stop the recipient seeing the email at that point are:

```

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,94 @@
...
+
+Currently travel-advice-publisher and specialist-publisher (the publishing tool for medical alerts) talk directly to email-alert-api. The current design philosophy for emails is that publisher apps should not talk directly to the alert API, instead publishing to a queue which email-alert-service consumes, and it then makes the calls. This is captured as technical debt, and we would have to ensure that when we pay down that debt the fix takes into account this mechanism. It's not anticipated that this will particularly complicate the debt, though.
+
+[kept for one week, then deleted]: https://github.com/alphagov/email-alert-api/blob/main/app/workers/email_deletion_worker.rb#L5
+[current delivery status]: https://docs.notifications.service.gov.uk/ruby.html#get-the-status-of-one-message
+[rate limiting service]: https://rubygems.org/gems/ratelimit/versions/1.0.2?locale=en
```
ChrisBAshton, line 83:
```suggestion
[rate limiting service]: https://github.com/alphagov/email-alert-api/blob/032fe645e62b5d3b26fedba84a3e852a87575a8b/lib/services.rb#L2C1-L4
```

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,94 @@
...
+
+We would need some way of matching the emails in the Email model to the emails expected to have been sent out. An MVP would be to use the exact same matching pattern (subject line and optionally time) that we use at the moment. This would get us the same level of reliability as now, but would retain the same problems (that if an alert's title is changed after it is sent out, the match will not work). To combat this, we could retain in Email records the content_id of the item that triggered them, and augment the [RSS feed] and [healthcheck URL] to add the content item into each displayed item. This is already publically available information and has no real security implication, but would require small changes to travel-advice-publisher and finder-frontend.
+
+To trigger an alert, we could either use a Prometheus collector on a reasonable metric, or use the Prometheus push gateway. A collector is probably preferable. The question of raising a PagerDuty alert from a Prometheus metric is already well understood by the platform teams.
```
ChrisBAshton, line 67: Would be good to expand on why a collector is preferable.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,94 @@
...
+
+### How this would be testable
+
+One of the problems with the current system is that it can only be tested in production, so it's important to consider how the new version could be tested. In integration and staging, emails are only sent to [specific email addresses]. Either we would need to add additional addresses to receive alerts, or sign up the existing ones. This would allow us to test the system working on the happy path (emails would be delivered to them, notify would mark them as delivered, and we could confirm that with the new monitoring code). To test the unhappy path (that alerts actually go off if the email isn't sent), temporarily removing the relevant subscription would cause there to be no detected receipts of any alert generated in that environment. We could possibly create rake tasks to disable/renable the subscriptions to make this easier for developers.
```
ChrisBAshton, line 71:
```suggestion
One of the problems with the current system is that it can only be tested in production, so it's important to consider how the new version could be tested. In integration and staging, emails are only sent to [specific email addresses]. Either we would need to add additional addresses to receive alerts, or sign up the existing ones. This would allow us to test the system working on the happy path (emails would be delivered to them, notify would mark them as delivered, and we could confirm that with the new monitoring code). To test the unhappy path (that alerts actually go off if the email isn't sent), temporarily removing the relevant subscription would cause there to be no detected receipts of any alert generated in that environment. We could possibly create rake tasks to disable/re-enable the subscriptions to make this easier for developers.
```

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Future Considerations
+
+email-alert-monitoring currently doesn't handle alerting on daily or weekly digests. We can probably safely ignore them for the moment. If we wanted to alert on them too we'd need some additional mechanism to tie the digests to a list of content_ids, and we would need to extend the [RSS Feed] and [healthcheck URL] endpoints in their respective apps to ensure data was available for the past week's worth of updates (currently not guaranteed in the [RSS Feed] and definitely not supplied in the [healthcheck URL])
+
+Currently travel-advice-publisher and specialist-publisher (the publishing tool for medical alerts) talk directly to email-alert-api. The current design philosophy for emails is that publisher apps should not talk directly to the alert API, instead publishing to a queue which email-alert-service consumes, and it then makes the calls. This is captured as technical debt, and we would have to ensure that when we pay down that debt the fix takes into account this mechanism. It's not anticipated that this will particularly complicate the debt, though.
```
ChrisBAshton, line 79: Worth linking to the tech debt?

ChrisBAshton (approved):
Thanks Keith, makes sense to me. Some minor nitpicks below, but happy to approve.

As you're a developer in the team that owns this application, I don't think you need to do much outreach beyond checking with your team TL/DM, as you should have agency to make improvements to your team's apps. But don't let that stop you getting one or two more ✅ approvals before merging, if you wish!

1pretz1:
In agreement with Chris that an improvement to the this alert is very welcome (and removing email-alert-monitoring) so great work on uncovering and kicking this off. Your plan seems reasonable to me, however as I was reviewing, I thought of a slightly different approach concerning the Notify confirmation part.

Instead of persisting Notify IDs and setting up our own cronjob, could we not tap in into the Notify status callback endpoint we [already have setup](https://github.com/alphagov/email-alert-api/blob/main/app/controllers/status_updates_controller.rb)? This currently only surfaces metrics (bar permanent failure that leads to subscriber being removed), but once upon a time it was very much integrated into the email system (see [ADR explaining the decision to remove](https://github.com/alphagov/email-alert-api/blob/032fe645e62b5d3b26fedba84a3e852a87575a8b/docs/adr/adr-006-email-delivery-responsibilities.md?plain=1#L1)). 

They send us a callback when the status of email is confirmed by Notify (delivered, permanent failure etc) and we already have an identifier setup in our system to match Notify records to Email Alert APIs: [email's id](https://github.com/alphagov/email-alert-api/blob/032fe645e62b5d3b26fedba84a3e852a87575a8b/app/services/send_email_service/send_notify_email.rb#L15) which is used in the response callback under ['reference'](https://github.com/alphagov/email-alert-api/blob/main/app/controllers/status_updates_controller.rb#L6).

I think this could have advantages against the current Notify ID / cron plan:
* Not relying on a cron job every x minutes and just letting Notify tell us when the thing is delivered
* Should be simpler to setup as some of the work is already done for us (don't need additional idenifier)
 
Potential disadvantages:
* If the callback system falls over (Notify's side) / isn't particularly loved then we may get alerted for emails that have actually been delivered, but we just haven't received a status from Notify. We could also argue this as potential risk with the Notify ID endpoint too, though. 

We'd still need to add `content_id` to the email / RSS / healthcheck for reliable results, though. Nice suggestion btw 👍 

I also agree that testing beyond Notify seems pointless as it out of GDS's remit (what could we do). We could take this one step further and state that once the email send request gets a successful response from Notify then that seems a reasonable end point, as Notify has their own alerting + retry logic and we wouldn't be able to do anything ourselves anyway, bar calling them up. This would also simplify the check + remove the risk of false positives, e.g not hearing back from Notify. However, we may deem this to be a step too far in terms of risk as we'd be less confident in the check 🤷 

Interested to hear your thoughts!

dgm34:
I'm late to the party but I've read the proposal (and the comments) and I'm in support of this RFC - it sounds like an excellent plan.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
maxgds: How does this help us? I had a discussion with Keith around some of this that led to us finding out more about how Notify satisfies itself that things have been delivered. If gmail/google groups is telling Notify "yep, got that" and we don't see the message, what lines of investigation does that actually open for us? If rando email server tells Notify that it got a message, rando user doesn't see it, and yet the gmail journey is fine and we've got one, again, what lines of enquiry are actually open? Is there definitely a value add in keeping a delivery chain open for this? Unless the users experiencing problems are also having their subs sent to a google group I'm not sure this adds much to the assurance.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,100 @@
...
+
+- The mailserver might fail to deliver internally: Valid, but we can't currently test that unless the failure occurs in the exact Gmail server handling our account anyway.
+- A spam filter might redirect the message: Untestable, relies on idiosyncratic account setup, and out of our control
+- The recipient might ignore the message: Untestable, and out of our control.
+
+### How we could replace this with a job internal to email-alert-api
+
+We currently keep a record of every single email sent out via email-alert-api (the Email model). These records are [kept for one week, then deleted]. This model could have a notify_status column added which begins nil and is then filled in by the callback. As long as one email sent in response to an alert has a notify_status of "delivered", we would have roughly the same assurance as we do now that the email alert has been sent successfully. It would also allow us the opportunity (if we wanted) to make a more stringent alert (say, that a certain percentage of the alert emails were delivered successfully) that we can't currently test.
```
1pretz1, line 68: A positive use of keeping the email box is an email should always be attempted for each travel / drug alert, even if there are no actual subscribers. We may think about a mechanism that prevents the alert being fired if there are no subscribers to hopefully prevent false positives (e.g if for some reason we the our Gmail box doesn't receive the email). Seems low risk but possibly something to factor in.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,100 @@
...
+
+[specific email addresses]: https://docs.publishing.service.gov.uk/repos/email-alert-api/receiving-emails-from-email-alert-api-in-integration-and-staging.html
+
+#### Future Considerations
+
+email-alert-monitoring currently doesn't handle alerting on daily or weekly digests. We can probably safely ignore them for the moment. If we wanted to alert on them too we'd need some additional mechanism to tie the digests to a list of content_ids, and we would need to extend the [RSS Feed] and [healthcheck URL] endpoints in their respective apps to ensure data was available for the past week's worth of updates (currently not guaranteed in the [RSS Feed] and definitely not supplied in the [healthcheck URL])
```
1pretz1, line 82: content_id -> content IDs

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,100 @@
...
+
+We would need some way of matching the emails in the Email model to the emails expected to have been sent out. An MVP would be to use the exact same matching pattern (subject line and optionally time) that we use at the moment. This would get us the same level of reliability as now, but would retain the same problems (that if an alert's title is changed after it is sent out, the match will not work). To combat this, we could retain in Email records the content_id of the item that triggered them, and augment the [RSS feed] and [healthcheck URL] to add the content item into each displayed item. This is already publically available information and has no real security implication, but would require small changes to travel-advice-publisher and finder-frontend.
```
1pretz1, line 70: content_id -> content ID

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,100 @@
...
+
+### How this would be testable
+
+One of the problems with the current system is that it can only be tested in production, so it's important to consider how the new version could be tested. In integration and staging, emails are only sent to [specific email addresses]. Either we would need to add additional addresses to receive alerts, or sign up the existing ones. This would allow us to test the system working on the happy path (emails would be delivered to them, notify would mark them as delivered, and we could confirm that with the new monitoring code). To test the unhappy path (that alerts actually go off if the email isn't sent), temporarily removing the relevant subscription would cause there to be no detected receipts of any alert generated in that environment. We could possibly create rake tasks to disable/re-enable the subscriptions to make this easier for developers.
```
1pretz1, line 76: Have we confirmed that status callbacks are received in staging / integration (I assume so)?

1pretz1: `notify` -> `Notify`

1pretz1: Thinking for testing the unhappy path - we may want to fake a Notify callback receipt vs testing with no subscriptions as we may want to guard against alerting when there are no subscriptions, but this could be a future consideration.

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
1pretz1: I agree that excluding digests is okay for now, as people signed up for digests don't intend to receive immediate updates 👍

rfc-164-retire-email-alert-monitoring-as-a-separate-app.md
```diff
@@ -0,0 +1,100 @@
...
+
+Currently travel-advice-publisher and specialist-publisher (the publishing tool for medical alerts) talk directly to email-alert-api. The current design philosophy for emails is that publisher apps should not talk directly to the alert API, instead publishing to a queue which email-alert-service consumes, and it then makes the calls. This is captured as [technical debt], and we would have to ensure that when we pay down that debt the fix takes into account this mechanism. It's not anticipated that this will particularly complicate the debt, though.
+
+[kept for one week, then deleted]: https://github.com/alphagov/email-alert-api/blob/main/app/workers/email_deletion_worker.rb#L5
+[current reciept status]: https://docs.notifications.service.gov.uk/ruby.html#delivery-receipts
+[technical debt]: https://trello.com/c/tWIZfxfc/25-travel-advice-publisher-and-specialist-publisher-talk-directly-to-email-alert-api
+
+### Proposed Action Plan
+
+- Update Email model to allow capturing notify callback status
```
1pretz1, line 92: `notify` -> `Notify`

1pretz1 (commented):
Overall looking great, just left a few comments but nothing major 👍 As discussed offline, the proposed implementation using status callbacks seems a good addition.

1pretz1 (approved):
Looks good to me - excellent work!

PR #165, opened 2023-10-17
RFC 165 - Add API endpoints to Transition for consumption by Bouncer

jkempster34:
[Trello](https://trello.com/c/Pr1sKak1) 

Deadline for comments: 10th November

[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/rfc-165-transition-bouncer-api/rfc-165-add-api-endpoints-to-transition.md)

### DIFF
--- a/rfc-165-add-api-endpoints-to-transition.md
+++ b/rfc-165-add-api-endpoints-to-transition.md
@@ -31,6 +31,10 @@ Bouncer should retrieve data from Transition via an API. This involves:
 - Adding adapters to GDS API Adapters.
 - Removing database dependencies from Bouncer.
 
+There has been a push recently to fully separate our frontend and backend apps.
+This approach would fit in well with this theme, essentially turning Transition
+into a publishing app and Bouncer into a platform concern.
+
 This approach is an alternative to merging to the Bouncer codebase into
 Transition entirely. Merging the applications is a more complex task that
 involves:
--- a/rfc-165-add-api-endpoints-to-transition.md
+++ b/rfc-165-add-api-endpoints-to-transition.md
@@ -25,8 +25,11 @@ ActiveRecord subclasses and dependency versions consistent.
 
 ## Proposal
 
-We should add API endpoints to Transition that will enable us to retrieve the
-data required for Bouncer to function.
+Bouncer should retrieve data from Transition via an API. This involves:
+- Adding API endpoints that allow us to retrieve the minimum amount of data for
+  Bouncer to function.
+- Adding adapters to GDS API Adapters.
+- Removing database dependencies from Bouncer.
 
 This approach is an alternative to merging to the Bouncer codebase into
 Transition entirely. Merging the applications is a more complex task that
--- a/rfc-165-add-api-endpoints-to-transition.md
+++ b/rfc-165-add-api-endpoints-to-transition.md
@@ -1,0 +1,40 @@
+# Add API endpoints to Transition for consumption by Bouncer
+
+## Summary
+
+This proposes that the Bouncer application should access transition data via
+API endpoints on Transition instead of via direct database access.
+
+## Problem
+
+The [Transition
+System](https://docs.publishing.service.gov.uk/manual/transition-architecture.html)
+is built to transition government websites to GOV.UK. It includes:
+- Transition - Ruby on Rails application used by admins to create mappings from
+  old URLs to pages on GOV.UK.
+- Bouncer - Rack-based application that uses the mappings created by Transition
+  and handles requests to those old domains.
+
+Currently, Bouncer retrieves the mappings written by Transition through direct
+read-only access to the database shared by the two applications.
+
+This is problematic as any code related to database access must be kept in sync
+between Transition and Bouncer.
+This includes the use of ActiveRecord; we must take the effort to keep
+ActiveRecord subclasses and dependency versions consistent.
+
+## Proposal
+
+We should add API endpoints to Transition that will enable us to retrieve the
+data required for Bouncer to function.
+
+This approach is an alternative to merging to the Bouncer codebase into
+Transition entirely. Merging the applications is a more complex task that
+involves:
+- Rewriting large parts of Bouncer as it's currently not a Rails app.
+- Handling of routes that are the same in Bouncer and Transition.
+- Dealing with infrastructure complexity - references to Bouncer are littered 
+  through the infrastructure.
+
+The documentation in the Transition and Bouncer repositories must be updated to
+reflect these changes.
--- a/rfc-165-add-api-endpoints-to-transition.md
+++ b/rfc-165-add-api-endpoints-to-transition.md
@@ -41,3 +41,12 @@ involves:
 
 The documentation in the Transition and Bouncer repositories must be updated to
 reflect these changes.
+
+## Consequences
+
+- Bouncer's dependency on the database shared with Transition is removed.
+- Bouncer's codebase will be less complex.
+- This approach will likely be less performant than the current approach. We
+  plan to take this into consideration during development by identifying and
+  measuring relevant performance metrics.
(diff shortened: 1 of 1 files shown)


rfc-165-add-api-endpoints-to-transition.md
craigsparling, line 1: Would the performance (principally speed) of Bouncer be negatively impacted by this change?

How much work would it take to make this change, compared to the other option of merging Transition and Bouncer? I don't need a very specific number. Maybe just a comparison. 90% of the work? 50% of the work? 10% of the work?

jkempster34: I think this will likely be less performant, though to what extent I'm not sure!

I think this will be 25% of the work compared to merging

mike29736: I agree about the response times, @jkempster34. We're comparing the current setup where Bouncer makes requests to a database with the proposed setup where Bouncer makes requests to Transition which makes requests to a database.

Unless there are existing impediments that we're unaware of, it doesn't sound like it can ever be as fast as it currently is.

Another thing that this brings to mind is the added risk in the proposed setup that comes with Bouncer depending on the availability of both the database and the Transition API.

Maybe those are both worth mentioning as, like, "consequences" in the doc?

rfc-165-add-api-endpoints-to-transition.md
```diff
@@ -0,0 +1,40 @@
...
+We should add API endpoints to Transition that will enable us to retrieve the
+data required for Bouncer to function.
+
+This approach is an alternative to merging to the Bouncer codebase into
+Transition entirely. Merging the applications is a more complex task that
+involves:
+- Rewriting large parts of Bouncer as it's currently not a Rails app.
+- Handling of routes that are the same in Bouncer and Transition.
+- Dealing with infrastructure complexity - references to Bouncer are littered 
+  through the infrastructure.
```
mike29736, line 37: You've included a breakdown of some of the implementation details of the alternative option and haven't gone into as much detail about the option we're proposing. If you left that out because it didn't seem noteworthy, I just thought I should put in a vote for: the contrast between the two lists could be interesting. And they might help illustrate the 25% estimate/help readers make their own estimates.

rfc-165-add-api-endpoints-to-transition.md
mike29736: By the way, if response times and/or other qualities of the current Bouncer setup are important to us, this might be one of those convenient cases where we can measure what we care about.

As part of the implementation we can take existing data and compare the new work against it. Or even set targets against it. Is that worth making part of the proposal somehow?

rfc-165-add-api-endpoints-to-transition.md
```diff
@@ -25,16 +25,27 @@ ActiveRecord subclasses and dependency versions consistent.
 
 ## Proposal
 
-We should add API endpoints to Transition that will enable us to retrieve the
-data required for Bouncer to function.
+Bouncer should retrieve data from Transition via an API. This involves:
+- Adding API endpoints that allow us to retrieve the minimum amount of data for
+  Bouncer to function.
+- Removing database dependencies from Bouncer.
```
mike29736, line 31: Is it worth mentioning adding support for the Transition API to GDS API Adapters as a distinct part of the work of introducing the endpoints?

rfc-165-add-api-endpoints-to-transition.md
floehopper: I'm coming at this from a place of significant ignorance about these apps, so please excuse me if what I say here doesn't make sense:

> - Rewriting large parts of Bouncer as it's currently not a Rails app.

Can you explain a bit more about why you'd need to rewrite the code from the Rack app? e.g. why couldn't you move the Rack app code into the Rails app and [set up Rails routes for the Rack app](https://guides.rubyonrails.org/routing.html#routing-to-rack-applications).
 
> - Handling of routes that are the same in Bouncer and Transition.

Could this be handled by moving the existing Transition routes under a namespace? I'm making a (possibly incorrect) assumption that there's a requirement to retain existing Bouncer routes, but existing Transition routes could more easily change...?

> - Dealing with infrastructure complexity - references to Bouncer are littered 
  through the infrastructure.

That's unfortunate. Are there any fewer references to Transition? i.e. could you merge Transition into Bouncer...?

jkempster34: Hi James, thanks for your comment. We are going to raise the idea of merging Transition into Bouncer in the next team dev chat. We can hopefully get a suitable response to you then. Can I ask whether you prefer the idea of merging the apps in general, or do you simply not find the justifications to not merge them convincing?

floehopper: > We are going to raise the idea of merging Transition into Bouncer in the next team dev chat.

I really don't know whether that's sensible or not (or even possible!) - it was just a thought!

> Can I ask whether you prefer the idea of merging the apps in general, or do you simply not find the justifications for not merging them convincing?

From what little I know (massive caveat!) and with what you've explained here the two apps sound very coupled to each other, so I'd lean towards merging them.

JonathanHallam: Hi James, We had a look at this idea and have decided against it. This is in part because of the added complexity but also because it means that Transition and Bouncer become Publishing and BE apps respectively. Figured this fits in with the work we've been doing to separate FE functionality from Whitehall. 

On the complexity front, I think it's worth noting that these aren't massive bits of GOV.UK and what we're proposing is an improvement. They can always be merged at a later date if we decide that the API approach isn't working.

richardTowers (approved):
Hello! Thanks for writing this up, and thanks everyone for the discussion.

To summarize the main points from the discussion:
- We feel like merging the two applications would require more engineering work than separating them with an API
- We're aware that there will be a performance impact to inserting an API between bouncer and the database, but we hypothesise that the extra latency is tolerable and we're going to keep an eye on it as we develop the solution.

I'm happy with this plan based on those points. Should hopefully be a relatively simple piece of work, with a reasonable architectural improvement as a payoff.

Let's do it :)

sengi:
I don't see a compelling case for this. The stated aim is to reduce the overall maintenance burden, but I doubt it would achieve that aim. I think it's more likely that it would end up increasing the overall maintenance burden while providing minimal value.

- The proposed work adds complexity by adding an abstraction layer between the two apps. The proposal simply assumes that this abstraction is worth the implementation cost and ongoing maintenance cost, without any attempt to compare that effort with the effort expected to be saved.
- While the proposed API would (if appropriately designed) decouple Bouncer from Transition's data schema, it would likely take decades to break even in terms of effort saved vs. effort spent up front. This is because:
    - The effort involved to introduce the API is significant (major changes to both apps), especially if the proposed performance analysis is done.
    - It's unlikely to take much effort to continue to keep the data model in sync between the two apps.
    - The data model changes very rarely.
    - The apps change very rarely. For example Bouncer has had 10 PRs this year (excluding docs and Dependabot), of which zero were changes to the data model.

sengi:
(fwiw I wrote this yesterday evening based on a discussion with Xander, before I'd seen Richard's post! Thought I'd still post anyway because I'd promised Xander I'd write up the case against.)

xander-gds:
We've talked about Chris' comments above, and agree that there's at least uncertainty about the effort, value and balance of the two. For now, we're planning to put this on hold while we work on some other things. We'll make sure that when/if we return to it, the first thing to do is a timeboxed spike to get more confidence in our estimate, followed by an evaluation of whether the work is compelling.

jkempster34:
After much discussion, we have decided that this is the right thing to do, but is not a priority right now. Therefore we are merging this as an intention to do it in the future.

brucebolt:
An example of the toil we're experiencing by the coupling of these two apps: https://github.com/alphagov/bouncer/pull/470

PR #168, opened 2024-03-04
Add RFC metadata as YAML frontmatter

richardTowers:
We reviewed the latest 42 RFCs with senior tech to check on their statuses (i.e. accepted / superseded etc.) and implementation status (i.e. done / in progress etc.).

This adds metadata to each RFC as YAML frontmatter.

Also add a script to the README to show how this can be used in a programmatic way. The results of running this script can be seen in this spreadsheet:

https://docs.google.com/spreadsheets/d/1bYF7r4IC7owbE7NrbFMZYJiuqsPZeQka1fzYD4apx9U/edit#gid=166323961

PR #169, opened 2024-03-06
Review status of RFC 8-30

kevindew:
This relates to https://github.com/alphagov/govuk-rfcs/pull/168

This adds in review dates for the RFCs 0 - 30. 

Broadly my changes have been: 

- Flag status of as many RFCs as I can
- Delete RFCs that are a complete formatting mess and don't seem to have any future value
- Improve/fix formatting (most of these were automatically imported from Confluence)


leenagupte (approved):
Wow, that was like stepping through the history pages of GOV.UK.

PR #172, opened 2024-05-13
RFC 172 - GraphQL for GOV.UK

richardTowers:
RFC 172 - GraphQL for GOV.UK

Deadline for comments: 31st May 2024

Please leave comments inline in the diff, as these can be threaded and resolved. General comments that might merit a reply should be made as comments on the heading on line 1.

✨  [Rendered version](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-172-graphql-for-govuk.md) ✨ 

image


rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
++--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+```
+
+Which can be [explored visually in PG Explain][pg-explain-query-plan]
+
+![PG Explain Visualizer showing the execution plan above](rfc-172/pgexplain-query-plan-visualisation.png)
+
+</details>
+
+Still, even with dataloader batching queries and individual queries making good use of indexes, we're asking the database to do at least around 10x more work than the current content-store database has to do, and that will have some performance or infrastructure cost implications. We will need to carefully consider how this solution performs at scale to evaluate these.
```
sengi, line 300: Our database instances are very underutilised right now, so I expect any cost implications would likely be below the noise floor, so to speak.

It's also important to set this valid concern in the wider context of enabling the rationalisation of GOV.UK's database architecture, which has the potential to reduce infrastructure costs and improve performance.

In other words, if the proposal is well executed and the mitigations you suggest are implemented, perceived performance from the end-user perspective is likely to improve while somewhat reducing running costs in the medium to long term.

sengi (commented):
I'm excited about this!

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+In terms of page load times as seen by an end user, there are additional factors that mitigate the impact of API performance. Mainly, the impact of caching - the ~90% of requests which we serve from Fastly's cache won't be affected at all by API performance, while the 10% of cache misses will. We could consider further tuning of our cache behaviour (for example using [stale-while-revalidate][stale-while-revalidate]) to further reduce performance impact.
+
+### Frontend applications will need to make two API calls instead of one
+
+Currently, frontend applications only need to make a single request to content store, and they can then switch on the document type in the response to determine which view to render.
+
+The value of the GraphQL approach is that the frontend can specify the fields it needs for the document type it's rendering. For this to work, the frontend needs to know which document type a particular page has before it can make the query to get the page data.
```
KludgeKML, line 310: Could this be mitigated by some kind of "standardised fields" option? I can't remember if GraphQL has a way of expressing this. I'm envisioning something like "Let me know what type of content this is, and also return the usual fields returned for this sort of content". I guess a little bit like a prepared statement in SQL. Then the usual fields are the ones required to render that type of page, and we only need to do one call. It does mean that there's a reduction in flexibility (if we want to add additional fields and keep the single call we have to manually tell the API that new fields are needed), but as an option if we really need the speed of just a single call, it's something.

richardTowers: Yes, I think you can achieve this [with interfaces / unions and fragments](https://graphql.com/learn/interfaces-and-unions/#fragments). So you could have one giant query like:

```graphql
query {
  edition(basePath: $basePath) {
    publicLastUpdatedAt
    # ... Other fields that are present on all editions
    
    ... on Guide {
      bodyHtml
      # ...
    }
    ... on Manual {
       parts
       # ...
    }
    ... on MinistersIndex {
       cabinetMinisters { }
       # ...
    }
    # ... ~180 other content types
  }
}
```

This would be a big request, but we could store it as a [persisted query](https://graphql-ruby.org/operation_store/overview.html).

DaveLee-GDS (approved):
It's a great concept and after various conversation with Richard and others about it, i'm convinced it will be a great enabler for the use-cases documented plus other emergent use-cases yet to be determined.
There are clearly challenges within, but i feel the benefits far outweigh the effort and investment to get it up and running. Top work!

dgm34:
Thanks @richardTowers! I'm really in support of this proposal (❤️ ), but I wouldn't be me if I didn't have a few comments.

* This RFC - probably by design - doesn't talk about what it would take to achieve this. At some point we'll have to work through the effort required to implement the various parts of this proposal, and get further buy-in from other Publishing stakeholders.

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+- Selecting only the subset of fields relevant to the client
+
+- Dynamically following links between content relevant to the client, for example on /government/ministers index page [listing the ministers, and for each minister all of the roles they currently hold](#query-to-get-the-content-on-the-ministers-index-page)
+
+- Aggregations of content, for example [listing governments](#query-to-list-governments) or [listing documents related to a topical event](#query-to-list-documents-related-to-a-topical-event)
+
+The implementation of this API would avoid many of the issues which make GOV.UK's current [publishing-api][publishing-api] / [content-store][content-store] / [search-api-v1][search-api-v1] architecture so complicated. In particular, it would avoid precomputing link expansion at the point when documents are published, and instead collect the links relevant to the client at query time. Similarly, with appropriate indexes in place, the API would cover many (or possibly all) of [the use cases we have for search-api-v1][search-api-v1-use-cases] and its associated Elasticsearch database.
```
kevindew, line 20: Is this assuming any relevancy based use cases are covered by Search API v2?

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+4. How to implement authentication for non-GOV.UK-microservice API clients (e.g. mobile apps), or whether to make the API fully public, as content-store is (although I think this is inadvisable)
+
+5. How to implement support for draft content
+
+6. How to authorise which clients are permitted to query draft content, and which are only permitted to query published content
+
+7. Whether pagination is required for nested resources, or whether we can only use pagination for top level queries
+
+8. Whether we need to adjust caching behaviour to lessen the impact of performance degradations (e.g. by using [stale-while-revalidate][stale-while-revalidate])
```
kevindew, line 232: A few things other things that'll probably need thinking about are:

### Message queue consumers

I imagine there might be an opportunity to send out much smaller notifications and then expect consumers to query Publishing API for more data, rather than the Content Store like object that is broadcast (and is presumably mostly superfluous).

However there does seem to be a problematic area, at the moment by having expanded links  in a message queue notification a consumer can make logic decisions based off data from the content item and the content in links. For example: in GOV.UK Chat we don't index content in [history mode](https://insidegovuk.blog.gov.uk/2015/03/23/what-happens-to-content-from-previous-governments-history-mode/) which involves a property of a piece of content (political) with a property of a link (government, with a boolean of current). Currently it's quite simple to know when this has changed as there is a message that the links of that piece of content have changed, whereas if there wasn't a replacement a consumer application may need to store the link graph itself to know when to act on a linked piece of content changing.

### Rendering/handling of Govspeak

Currently as part of the work to downstream content there is the rendering of Govspeak for content that uses this feature, which can be quite a slow task for a large piece of content (I have vauge, potentially inaccurate, memories of this potentially reaching seconds for a very large document). I imagine switching from a Ruby implementation to a C implementation of markdown rendering could speed this up drastically - but carries a lot of migration pain.

The other aspect of Govspeak is a rather quirky Publishing API feature that Content Store conceals. Which is that fields intended for HTML or Govspeak have a type of array which contains objects with a content type. For example a [mainstream Answer](https://github.com/alphagov/publishing-api/blob/021710b00af2977d3944d4180dbf0691f8520996/content_schemas/dist/formats/answer/publisher_v2/schema.json#L299-L317) sends a body field of:

```
"body": [
  { "content_type": "text/govspeak", "content" => "## Hello" }
]
```

Which gets broadcast to the message queue and content store as:

```
"body": [
  { "content_type": "text/govspeak", "content" => "## Hello" },
  { "content_type": "text/html", "content" => "<h2>Hello</h2>" }
]
```

And then, at runtime, [content store mutates it](https://github.com/alphagov/content-store/blob/4973198b4a6b185447ff489a5d84089a163f056a/app/presenters/content_item_presenter.rb#L45) into:

```
"body": "<h2>Hello</h2>"
```

I imagine we have to make a decision as to how this type of field would be queried and what data would be served (I imagine we could push markdown rendering down to clients)

### Schemas and data validity

I assume we need to keep parity between schemas and a GraphQL object definitions. Is this going to be achievable? I'm guessing in early cases we could even use the current jsonnet schema files to generate the GraphQL object definitions.

One of the long running Publishing API problems with schemas has been that they're enforced only at the time of insert and then if a schema changes later the data in the Publishing API no longer conforms to it. Would GraphQL be able to cope with this relatively gracefully (i.e. ignore superfluous fields and return nil to missing ones) or return errors (which I suppose would enforce validity) ? I guess if it did try to be graceful it'd main almost every field would have to be flagged as being potentially nil.

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+Unlike currently, more complicated pages (with lots of links) will be more expensive to render than simple pages. The most complicated page by far on GOV.UK currently is the ministers index page, and in the first prototype GraphQL API this takes between ~500ms and ~1000ms to generate an API response - real world performance may differ, but it will be important to evaluate the performance of the more complicated pages.
+
+In terms of page load times as seen by an end user, there are additional factors that mitigate the impact of API performance. Mainly, the impact of caching - the ~90% of requests which we serve from Fastly's cache won't be affected at all by API performance, while the 10% of cache misses will. We could consider further tuning of our cache behaviour (for example using [stale-while-revalidate][stale-while-revalidate]) to further reduce performance impact.
+
+### Frontend applications will need to make two API calls instead of one
+
+Currently, frontend applications only need to make a single request to content store, and they can then switch on the document type in the response to determine which view to render.
```
kevindew, line 308: With a bit of squinting you could describe our current architecture as already having this given we have the call to Router to first look up the content destination and then having the application look up the content. 

So we could, in someways, handle the first query as an aspect of routing which can also determine which code should be used to render an item. At the moment [we have frontend that loads full content items as a means of routing](https://github.com/alphagov/frontend/blob/50d0ffe4db54bbb34398aa4c90536e89e10fb0e5/config/routes.rb#L59-L97), [government frontend that has a single action for _every_ content item](https://github.com/alphagov/government-frontend/blob/30fe1de1026971d08c22a734bf3956dcbfbbb355/config/routes.rb#L26-L29) and [collections that does the routing based on URL structure](https://github.com/alphagov/collections/blob/main/config/routes.rb). If we do further frontend app consolidation we want to standardise on an approach.

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+- Draft Content Store (instead serving draft frontend requests from the GraphQL API, with appropriate authorization in place)
+
+- Search API V1
+
+- Router / Router API
+
+- Link Expansion
+
+- Dependency Resolution
```
kevindew, line 336: Would be glorious to see all these things go 🙏

kevindew (commented):
This sounds really cool, really well explained proposal and lots of cool ideas - I hope we have success with it. Does seem like there are some thorny areas - particularly performance (though I do wonder whether we could utilise a key value store to reduce need to hit postgresql for all the bits)

My biggest concern with this is, given this will take quite a long time to implement and migrate everything, how we minimise the risk of ending up stuck with an incomplete partial implementation if things don't pan out like we hope. We've seen this many times before on GOV.UK and I wonder what insurance steps we could take to de-risk it.

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+As mentioned above, there are two types of link.
+
+- "[Link set links][link-set-links]" link all versions (draft and published, in all locales) of a document to the latest version of another document.
+
+- "[Edition links][edition-links]" link a specific edition of a document (e.g. the latest draft, or the latest published version) to the latest version of another document.
+
+In many cases, edition links are preferable, because they allow users to draft adding / removing / reordering links on a specific edition. With link set links, you can't change the links on a draft without immediately changing the links on the published edition.
+
+The docs suggest that link set links "are typically used for taxonomy based links", but in practice they're used much more widely, including for link types like "documents" and "ordered_roles" which feel like they ought to be edition links. There are around 5 million link set links on GOV.UK, compared to only around 1 million edition links to published editions.
```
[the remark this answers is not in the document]
richardTowers: [Market research? If you build it, they will come. That's my market research.](https://www.youtube.com/watch?v=Qp4xQGyL80Q)

Yeah. If we'd fully implemented edition links and then done the leg work to migrate everything over, we'd be in a better situation. I think a big part of the reason they're not fully implemented / adopted is the general sense of fear uncertainty and doubt around the link expansion code. Resolving that is the great hope of this RFC. Of course there might also have been a degree of us not following through on technical improvement work, but I can't resolve that issue with an RFC.

I appreciate the concern that GraphQL could also be an idea that only gets half adopted. Maybe it would be healthy for us to plan to fully revert the proof of concept, once we've learned what we're going to learn - I definitely don't think we want to end up with some once hopeful PoC code lingering around used in 5% of places for the next 5 years.

I don't want to grow the scope of this RFC to a full "how are we going to adopt this everywhere" plan, because I think that's jumping the gun. If at the end of the PoC we're all convinced it's a good idea, and something we want to pursue then we should have the discussion about how to make sure it's fully adopted. Or alternatively, we write up what we learned and continue with the current architecture.

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+We propose adding a new [GraphQL][graphql] content delivery API to GOV.UK.
+
+Initially, we will start with a minimal proof of concept, which will be deployed only to non-production environments and not queried by production code. This will allow us to experiment with the concept of GraphQL, and learn about the right shape for the API before we commit significant engineering time.
+
+If we're happy to proceed based on what we learn from the proof of concept, we will consider various implementation details before enabling the API in production. We'll then make a gradual transition from using content-store and search-api-v1 in the website frontends to using the GraphQL API, while we continue to learn about its strengths and weaknesses.
+
+### Initial proof of concept
+
+As an initial proof of concept, we will develop a simple GraphQL API using [GraphQL Ruby][graphql-ruby] inside publishing-api. This API be disabled by a feature flag in production, but in other environments it will run in the same process as publishing-api, and query the primary publishing-api database.
```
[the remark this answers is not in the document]
richardTowers: Yeah. At the moment my [pre-proof-of-concept-proof-of-concept branch](https://github.com/alphagov/publishing-api/compare/main...graphql#diff-d35f07e99d6cc9240ca8e6dfcbe1f194d2ca6da8b133c5158d5cd8ffd6d68e30) keeps almost all the code in `app/graphql`, so it could be cleany deleted (so long as nothing ends up using it in production).

The only other thing we'd need to add outside of this directory is the new controller.

I'll add to the RFC that we should keep as much of the code as possible in `app/graphql`

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+We will only implement GraphQL queries (i.e. the API will be read-only). There's no plan to enable GraphQL mutations or subscriptions.
+
+The API will initially support published content only - it will not show any draft information.
+
+Authentication will use a signon bearer token, as we currently do with publishing-api itself. All users of the API will have full read access - there is no need for Authorization until we introduce draft content.
+
+To allow efficient loading of large numbers of editions, we'll use the [Dataloader][graphql-ruby-dataloader] pattern, which batches queries together to reduce the total number of queries needed.
```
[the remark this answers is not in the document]
richardTowers: Ooh, I hadn't heard about that. That'll be useful in Whitehall too 👀

rfc-172-graphql-for-govuk.md
richardTowers: It's handwaving a bit I think 😅  - could do with being more explicit. I think:

1) Use cases with no query, ordered by newest (e.g. latest pages in a topical event) are clear cut - should be graphql
2) Use cases with no query, ordered by popularity (e.g. /search/services landing page) are only doable if we move popularity statistics into publishing-api, because if we need to do `select ... order by ... limit ...` then the database needs to have the field we're ordering on. Personally I'd go for that, but probably a separate discussion.
3) Use cases with a query, where we're happy with keyword search (rather than semantic search) are only doable if we have some kind of free text search capability in a DB that the graphql API has access to (could still be publishing-api with postgres free text search, or it could be an OpenSearch database or something if we don't mind maintaining another DB)
4) Use cases that require semantic search / relevance based ordering should be Search API v2

rfc-172-graphql-for-govuk.md
richardTowers: ### Message queue consumers

Ah excellent, thanks for the example. I was aware that the eventual plan to remove dependecy resolution was likely to run into some problems like this, but I couldn't think of a clear example. I was most worried about the search indexes (which I guess GOV.UK chat is one of).

Without dependecy resolution happening, it's going to be tricky to keep any of our denormalised views of data synchronised with information in linked documents. This could be a big problem with content reuse too, for example if there's a page that references VAT rate like `The VAT rate is {{current_vat_rate}}`, and it ends up in search as `The VAT rate is 20%` and then later `current_vat_rate` changes, we need to know to update every document in search that references that value.

We'll need to think about how to architect this, because queries to the GraphQL API could potentially follow links that aren't in the expansion rules, so it won't be robust to do dependency resolution like we currently do.

It's tempting to think about [GraphQL Subscriptions](https://graphql-ruby.org/subscriptions/overview.html) for this, but I'm not sure they're really the right solution - it feels like the assumption is that clients will be showing a specific page, and they'll use a subscription to do live updates to that page. Whereas we'd want to subscribe to lots of things.

I'll definitely add this to the RFC as something to consider before a wider rollout. We'll need an answer to this question, or really the whole idea is holed below the waterline in my view.

richardTowers: ### Rendering/handling of Govspeak

This one I was aware of. Currently [my poc handles this](https://github.com/alphagov/publishing-api/blob/graphql/app/graphql/types/generic_details_type.rb#L29-L47) at request time by checking for `text/html`, then falling back to `text/govspeak` and rendering it on the fly.

I've definitely seen the conversion of govspeak to HTML be slow - for example on the ministers index page publishing-api spends about 8 seconds converting govspeak when it's presenting the content to content store. But that's rendering lots of little documents (which the frontend never needs).

I guess for massive govspeak documents, there's a good chance this might be problematically slow. I'll check some.

richardTowers: ### Rendering/handling of Govspeak (continued)

For realistically massive govspeak documents, we're talking about ~500ms. The biggest bit of govspeak on GOV.UK at the moment is on https://www.gov.uk/guidance/immigration-rules/immigration-rules-appendix-skilled-occupations. This is pretty complex too - tables with lists inside cells etc.

This takes ~500ms to render on my mac:

```ruby
edition = Edition.live.find_by(base_path: '/guidance/immigration-rules/immigration-rules-appendix-skilled-occupations')
big_govspeak = edition.details.dig(:body, 0, :content)
Benchmark.bm do |x|
  20.times do
    x.report { Govspeak::Document.new(big_govspeak).to_html; nil }
  end
end
```

```
       user     system      total        real
   0.543176   0.011939   0.555115 (  0.559625)
   0.529987   0.012697   0.542684 (  0.544910)
   0.516697   0.011469   0.528166 (  0.529885)
   0.524782   0.012387   0.537169 (  0.539990)
   0.513405   0.011126   0.524531 (  0.526543)
   0.511204   0.012277   0.523481 (  0.525216)
   0.513562   0.011301   0.524863 (  0.526484)
   0.504459   0.009990   0.514449 (  0.515474)
   0.519732   0.011064   0.530796 (  0.532878)
   0.498527   0.010224   0.508751 (  0.509766)
   0.515784   0.012898   0.528682 (  0.530133)
   0.503691   0.010312   0.514003 (  0.515056)
   0.516301   0.010775   0.527076 (  0.528383)
   0.522541   0.010417   0.532958 (  0.534817)
   0.528775   0.011795   0.540570 (  0.542422)
   0.529228   0.013451   0.542679 (  0.544661)
   0.518665   0.010731   0.529396 (  0.531591)
   0.513944   0.011284   0.525228 (  0.526737)
   0.523058   0.011911   0.534969 (  0.536822)
   0.517797   0.011215   0.529012 (  0.530628)
```

In this particular case, publishing_api has the HTML version as well, so we wouldn't have to do this rendering (my SQL fu isn't strong enough to find documents that have govspeak but not html).

Pages with very long bits of govspeak which aren't pre-rendered to HTML are going to be another worst-case performance situation that we'll have to keep an eye on though.

richardTowers: I did a bit of investigation into how often the body is an array with text/govspeak or text/html vs. a string.

```
+---------+-----------------+-------------+------+
|body_type|has_govspeak_body|has_html_body|count |
+---------+-----------------+-------------+------+
|string   |false            |false        |474986|
|array    |true             |false        |205506|
|null     |false            |false        |117316|
|array    |true             |true         |2863  |
|array    |false            |true         |1     |
+---------+-----------------+-------------+------+
```

... so there are >200k pages where we'd need to do this conversion (out of a total of ~800k). There aren't that many pages (<3k) where publishing-api has both types available, but there are _lots_ of pages where `body` is just a string (which will be HTML by the looks of it).

richardTowers: ### Schemas and data validity

> I assume we need to keep parity between schemas and a GraphQL object definitions. Is this going to be achievable? I'm guessing in early cases we could even use the current jsonnet schema files to generate the GraphQL object definitions.

Yes, that was my thought too. In the short term we'd keep the current json schemas, and the graphql APIs would have to at least mirror those (maybe with some slight transformations, e.g. always returning HTML from `body` instead of exposing the govspeak / html choice to clients). A lot of the graphql code could be generated from the schemas, as you say.

> One of the long running Publishing API problems with schemas has been that they're enforced only at the time of insert and then if a schema changes later the data in the Publishing API no longer conforms to it. Would GraphQL be able to cope with this relatively gracefully (i.e. ignore superfluous fields and return nil to missing ones) or return errors (which I suppose would enforce validity) ? I guess if it did try to be graceful it'd main almost every field would have to be flagged as being potentially nil.

I'm not planning on changing the way that content is stored, so we'll still have this problem. I'd probably start with making the GraphQL API quite permissive of bad data (i.e. if it finds something in publishing-api's database that's not valid, it should still do its best to render a response, maybe with some [errors reported using GraphQL's errors field](https://spec.graphql.org/draft/#sec-Errors).

If we wanted to really make sure that things in the database are valid against the schemas, we should probably do that as some kind of batch job / test / database constraint.

I wonder how much of a problem this has been in practice?

rfc-172-graphql-for-govuk.md
richardTowers: At one stage I had a whole section in this RFC about how we could have router make the first query, and then pass the schema / document_type through to the frontend as headers. But then I decided it was too much detail / growing the scope too much.

I totally agree that the first query feels like part of routing, and the second one part of rendering. If we're able to consolidate the frontends on a similar timescale to the graphql rollout, it would be great to just do this in a rails routing constraint. Even if we don't do that though, we could explore doing a "which doc type, schema and rendering app should I use for this page" query in router, and then passing these bits through to the frontend. At that point, router could probably be replaced by an off-the-shelf / open source routing component with a custom plugin to do the graphql query ✨

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
+---
+status: proposed
+implementation: proposed
+status_last_reviewed: 2024-05-13
+---
+
+# GraphQL for GOV.UK
```
richardTowers, line 7: @dgm34 - responding to your comments as an inline comment, so it can be threaded 🧵 

> - This RFC - probably by design - doesn't talk about what it would take to achieve this. At some point we'll have to work through the effort required to implement the various parts of this proposal, and get further buy-in from other Publishing stakeholders.

Yeah, that's mostly by design. I think the initial proof of concept work can probably be justified and a place found for it in the publishing-platform team's roadmap. Building a production ready solution which will work at GOV.UK scale is going to require more thought, and I don't have a good feel for how much work that's actually going to be (hopefully the PoC should resolve taht though). Also as Kevin points out, building the API is one thing, actually adopting it in all of the frontend code is probably the bigger bit of work (and if we don't complete that, we leave ourselves in a worse position than when we started).

> - I feel quite strongly that the GraphQL API should not be public, at least until we understand a whole lot more about how it performs and impacts our systems. I note that you also think this is inadvisable and I'm reassured by that.

Yeah, I'll add that to the RFC - I think allowing anyone external to have access to the API is at the very least a problem for after the PoC.

> - It would be interesting to see some 'worst case' render times. I wonder if we can do this using some traffic replay magic, to see how render times vary under real load (subject to trying to correct for things like different environments, etc)

I'll add some examples of pages which I expect to be worst cases to the RFC, and we can evaluate them once the PoC is in integration. The possibly pathological cases I'm aware of are (1) pages with a lot of links, like the minsiters index page (2) pages with a lot of govspeak which needs transforming into HTML (3) pages which conduct several searches (e.g. topical events, which list recently published content tagged to the event).

> - I'd really like to get the new Lead TA for Publishing involved in this when he joins. Perhaps he could even own / drive it, working with you of course, in a similar way to how Al took responsibility for the Mongo to PostgreSQL migrations.

Absolutely - it would be absolutely great to have a TA to work through this with.

> Thanks again for this great write up ❤️

🙇🏻  you're very welcome

nacnudus:
I'm unfortunately unable to comment inline (as a GOV.UK exile).

1. This is a great read. Thank you for putting the work into it. It would be wonderful to make the structured data more accessible.
2. The benefits are arguably about custom, on-demand link expansion. GraphQL is arguably of secondary importance. I'm sure GraphQL is the right choice, but it isn't the only one. 
3. Could Govspeak be rendered once when it's received by the Publishing API, rather than every time it is queried? I vaguely recall that this was once on a roadmap. Another benefit of doing it only in the Publishing API, rather than upstream or downstream, would be that exactly one version of Govspeak would be used.
4. I have an awful feeling that when `body` is a string, it's sometimes govspeak rather than HTML. I think GovGraph takes the cautious approach of running everything through govspeak rendering, just in case.
5. Public access to a GraphQL API would spare departments from having to scrape the Content API, which I keep hearing about them doing (or considering doing).
6. Querying inside the `details` column of the Publishing API is very expensive, because that's where the body text is, but it's also where a lot of other useful data is.  Separating those things might help with performance.

richardTowers:
Thanks @nacnudus ! Great to have some thoughts from the GOV.UK diaspora. I hadn't realised you needed write permission to comment inline - sorry about that.

> 1. This is a great read. Thank you for putting the work into it. It would be wonderful to make the structured data more accessible.

🙇🏻  thank you!

> 2. The benefits are arguably about custom, on-demand link expansion. GraphQL is arguably of secondary importance. I'm sure GraphQL is the right choice, but it isn't the only one.

Yes, that's true. The proposal is to use GraphQL, so I think it's right that this RFC is framed that way, but you're right that we could get the same benefits with any technique that moves link expansion from publish-time to request-time.

I considered a few alternatives, but not in enough depth to write them up in the RFC. I've thought about:

1) A custom REST API with one endpoint per content type (which does link expansion on demand). This would resolve the issues with link expansion, but it wouldn't allow for clients to be flexible in their requests - we'd still need to decide at API design time what would be in any given response.
2) Something similar, but with GRPC - I think this would be basically the same as a REST API, in that we'd need to determine the payloads the API could respond with at API design time. But like GraphQL it would bring in a new, unfamiliar technology. Feels like it's in "worst of both worlds" territory.
3) Giving clients direct low level query access using something like SQL or SPARQL. But I think these are too flexible, and expose too much of the way data is stored internally. Some of that could be mitigated with SQL views. It would be very hard to do this in a way that didn't allow clients to make very unperformant queries, or open up security issues though.

> 3. Could Govspeak be rendered once when it's received by the Publishing API, rather than every time it is queried? I vaguely recall that this was once on a roadmap. Another benefit of doing it only in the Publishing API, rather than upstream or downstream, would be that exactly one version of Govspeak would be used.

Yes, I think all options should be on the table there. At the moment (as you know) we're already pretty wildly inconsistent - some things are stored as HTML (having been rendered by the publishing app), some are stored as Govspeak but rendered when they're presented to Content Store, and (maybe) some things make it all the way through to the frontend as Govspeak.

There are advantages and disadvantages to each option. Doing the conversion at publishing time is nice for query-time performance, but it makes updates to the Govspeak annoying (e.g. we sometimes tweak the way it outputs HTML for accessibility reasons, and we then have to "republish" every document because we need to update its Govspeak). On the other hand, doing it at request / render time means we always get the latest Govspeak version rendering the content, but it adds a performance overhead.

In terms of the proof of concept suggested in this RFC, I would prefer not to have to change the way data is stored at all, so that means doing some rendering of Govspeak at request time. It will be interesting to see how much of a performance issue that is, and we can consider what the right approach is before we productionise this.

> 4. I have an awful feeling that when body is a string, it's sometimes govspeak rather than HTML. I think GovGraph takes the cautious approach of running everything through govspeak rendering, just in case.

From what I can see in the publishing-api database, if body is a string, it's always HTML... ... appart from government/history pages ...

<details><summary>Man-Made Horrors Beyond My Comprehension</summary>

... where's it's ... HTML escaped ERB ❗❓❗❓...

```
$ curl --silent 'https://www.gov.uk/api/content/government/history/king-charles-street' | jq -r '.details.body' | head


    &lt;%= render "govuk_publishing_components/components/breadcrumbs", {
      breadcrumbs: [
        {
          title: "Home",
          url: "/",
        },
        {
          title: "History of the UK Government",
...
```

</details>

 I'm less confident about other fields of details though... but hopefully they're at least consistent within a particular document type.

> 5. Public access to a GraphQL API would spare departments from having to scrape the Content API, which I keep hearing about them doing (or considering doing).

Ooh, that's interesting - I wasn't aware of anyone using the content store API other than us.

The fact that we don't know who's using the content store API is exactly why I'm inclined to (at least) restrict access to people who we've given an API key to - I feel quite strongly that we shouldn't let people use this API unless they've agreed to some terms and conditions (e.g. they accept that we might make breaking changes) and we have a way to revoke their access that's better than blocking their IP.

That said, I _do_ think we should make it as easy as possible for the public to get an API key and use the API. But for now I'll stick to what I said to Dan "allowing anyone external to have access to the API is at the very least a problem for after the PoC".

> 6. Querying inside the details column of the Publishing API is very expensive, because that's where the body text is, but it's also where a lot of other useful data is. Separating those things might help with performance.

Yeah, the details column will be [TOAST](https://www.postgresql.org/docs/current/storage-toast.html)ed, so querying for things inside it requires loading pages off disk and untoasting them. This could be significantly improved [by indexing](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING). For the more search-y use cases in the RFC, particularly those where we'd want to facet on things in the details column without strongly selective document type filter, we'd probably need to look at this.

You could well be right that we might have to consider pulling the really chunky bits (body, parts) out of details into their own thing, so we can efficiently search the _other_ things in details, which are more likely to be used as facets.

rfc-172-graphql-for-govuk.md
richardTowers: Addressed in 6a596a6b6b6275cd066395c1453900746bced575 and 8098970d5b4035a4feaa687ca41ca48d8600ce58

rfc-172-graphql-for-govuk.md
richardTowers: Removed the rather ambitious "(or possibly all)" parenthetical in 76181bc1cd9ef9c2f88b0ee65170a2c3c5c883e3, and quoted this thread in the commit message.

rfc-172-graphql-for-govuk.md
```diff
@@ -0,0 +1,509 @@
...
+
+The concepts of link expansion and dependency resolution are [acknowledged as complicated][link-expansion-is-complicated-comment], and the code which implements them is hard to follow, and difficult to modify to cover new requirements.
+
+The GraphQL API proposed in this RFC doesn't depend on link expansion or dependency resolution because it computes links at request time instead of publish time. It doesn't need a central list of link expansion rules, because the frontend can specify the links it needs expanding in the GraphQL query.
+
+Although the concepts of link expansion and dependency resolution will continue to be required until we're able to remove content-store (which is a very long term goal), the GraphQL API could allow us to do more with links because developers can add features without necessarily having to worry about link expansion rules (or link expansion / dependency resolution generally).
```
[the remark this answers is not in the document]
richardTowers: Added a note about this in a555dd485a55ea26bbbe93795c8cc23b9d538504

rfc-172-graphql-for-govuk.md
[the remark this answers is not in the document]
richardTowers: Noted in b48a6478682955deb197d7180f8dc0335b1cdac8

dgm34 (approved):

richardTowers:
Coincidentally, there's been a fair bit of commentary online about GraphQL in recent days, prompted by this post by Matt Bessey - [Why, after 6 years, I’m over GraphQL](https://bessey.dev/blog/2024/05/24/why-im-over-graphql/). Most commentators tend to agree with Matt on several drawbacks of GraphQL, and the prevailing sentiment seems to be negative.

However, even having considered these issues, my view is still that [it might work for us](https://www.youtube.com/watch?v=Po4adxJxqZk).

More specifically, I think the valid drawbacks are less relevant to our use case:

> 1. Attack surface (Authorisation, rate limiting, query parsing)

This is mostly not relevant to us, at least initially, because we're not planning on exposing the GraphQL API to untrusted clients. This will become more of a concern if we allow mobile clients direct access to the GraphQL API, which we _are_ planning on doing. Even then, because all of our content is public, and it's modelled in a simple and consistent way in the database (so query performance is predictable), I don't foresee this being a big issue for us.

> 2. Performance

I've mentioned in the RFC that "To allow efficient loading of large numbers of editions, we'll use the Dataloader pattern", which should avoid the N+1 problem. Matt's article mentions that having to defensively use the dataloader pattern everywhere increases boilerplate, which is true. However in our case I think this should be less problematic, because of the simplicity of publishing-api's database schema - we basically only need one dataloader-ified query - to go from a list of editions to a list of lists of linked editions.

Again, authorisation won't be a problem unless we decide to serve non-public (e.g. draft) content to some clients. We may want to do that at some point, but it's still relatively simple to avoid serving non-live content.

> 3. Coupling

We're not planning on touching the tricky areas Matt mentions here (authorisation and mutations). I don't expect that adopting GraphQL will force us to do extensive additional integration testing - I suppose we will have to think about the contract tests however, as we currently use those to give us some confidence we're not breaking interfaces between components.

> 4. Complexity

I think the RFC makes it clear that we've got a lot of complexity already, and that we'd be able to simplify a lot by adopting this proposal.

At the end of the day, if we find that the complexity that GraphQL brings to the party is too much for us, there should be a reasonably straightforward path to a REST API that delivers many of the same benefits (as I suggested in my response to @nacnudus). We could start by moving things into persisted queries, and possibly remove GraphQL entirely and replace these with REST endpoints.

---

While I appreciate the suddent pushback on the GraphQL hype cycle, I think the latest comments are just the same thing in the other direction. Where the hype cycle might lead you to incorrectly believe that GraphQL is the right solution for every problem, this latest anti-hype cycle could lead you in the other direction - that it's never the right solution.

I'm confident enough that it's a good solution to the specific problem that GOV.UK has to proceed with a proof of concept. Let's see where we end up!

PR #173, opened 2024-05-14
Update process documentation to reflect used conventions

kevindew:
This updates the process to reflect that RFCs should have a deadline and that senior tech members are the final voice of whether an RFC is accepted. I'm surprised this wasn't previously documented.

It also replaces notes about how to manage Action Plans, Standards and archiving since those concepts have never become established.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,15 +6,22 @@ GOV.UK staff use this repository as a forum to discuss and make technical decisi
 
 1. Create a new branch on this repo and copy `rfc-000-template.md` to `rfc-000-my-proposal.md` and edit.
 2. Include any images etc in a separate directory named `rfc-000` and link to them.
-3. Make a Pull Request (PR) for your branch.
-4. Rename your file and directory with the number of the PR and push as a new commits.
-5. Post a link to your PR in #govuk-developers on Slack and to the [govuk-tech-members][govuk-tech-members] Google Group. @mention teams in the PR description if you want particular attention from them.
+3. Make a Pull Request (PR) for your branch and open as a draft.
+4. Rename your file and directory with the number of the PR and push changes.
+5. When your RFC is ready to be commented on, mark you PR as "Ready for review" and set a deadline for comments (2 weeks is a common choice) and record that in the PR description.
+5. Post a link to your PR in #govuk-developers on Slack and to the [govuk-tech-members][govuk-tech-members] Google Group.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,31 +1,36 @@
...
+8. Stay active in the discussion and encourage other relevant people to participate. If you’re unsure who should be involved in a discussion, ask your Tech Lead or a Lead Developer. If you start an RFC it’s up to you to push it through the process and engage people.
+9. Once the deadline is reached you are able to merge if the PR has been approved by a member of GOV.UK Senior Tech (if they haven't or you're not sure, you can contact them on Slack with @govuk-senior-tech-people) - this action is the accepting of an RFC.
 
-## Managing Standards
+### If consensus isn't reached
 
-Standards RFCs shouldn’t be substantially altered after they are accepted, although it’s fine to correct typos and other mistakes via a new PR. In order to change a Standard, the original RFC must be superseded by a new one. The process for this is:
+Should consensus not be reached by the deadline, an RFC can either have the deadline extended or have the RFC closed as rejected. For a deadline extension, you should make an effort to inform #govuk-developers about this status and request extra input. If the RFC is being closed as rejected, leave a comment explaining the status and close the PR.
```
richardTowers, line 20: I wonder if we should have a softer status than rejected for things where we didn't conclusively reject the RFC, but didn't reach consensus.

I can imagine an RFC proposing something sensible but with controversial details getting "rejected", and then later pointed at to say "we agreed we wouldn't do this", when really its status means "we didn't agree that we would do this", which is a lot softer.

Maybe `no_consensus`?

I guess if we add that subtlety we'd also need to agree how to actually close them - close with a comment stating whether it's rejected, or just that consensus wasn't reached and future discussion / RFCs in the same area are welcome?

kevindew: Sure, makes sense - do agree that rejected feels a bit wrong. I'll try adjust the wording as I don't think an RFC that isn't accepted necessarily needs a formal status since it's not merged. 2 tics.

kevindew: I've made an edit, what do you think? https://github.com/alphagov/govuk-rfcs/compare/2dff53506e13771a5a60072f1b84149c426d67a6..9cd187633fc3b2b2bdbc103818fbb0f3c450f687

richardTowers: Yeah, lgtm - thanks for that!

PR #174, opened 2024-06-14
[WIP] RFC 000 - Consolidate Frontend Apps

leenagupte:


rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Code structure
+The code structure will be standardised. As well as using the [default Rails directory structure], we propose the following standard for additional classes (e.g. services, api calls). The standard expands on the [conventions] documented in the developer docs.
+
+The host app will follow the default MVC structure of Rails applications. In addition to that:
+
+Under `/app`:
+
+models
+- Ingest data from yaml / JSON / CSVs rather than database.
```
KludgeKML, line 155: Suggestion:
  "Frontend apps do not have database records, so models here are created to store data received from api calls (eg content items from content store), process data that will be written to apis (eg feedback items to zendesk), and information ingested from YAML/JSON/CSVs. Unless it's not otherwise advisable, static data in the app will be stored in the locale files to promote localisation."

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,307 @@
...
+
+Under `/lib`
+- Rake tasks
+
+Use `lib` rather than `app/lib` for custom code that enhances the app but doesn’t fit into a directory in `app`
+
+Under `/config`
+
+data 
+- YAML, JSON, CSV config files. Data files should be placed here rather than under `/lib/data`
```
KludgeKML, line 173: "Static data like lists of contact information should ideally be stored in the locale files to promote localisation"

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: The first bit sounds really good. Thanks. 👍 

What data are you think of here?
> static data in the app will be stored in the locale files 

I think static and dynamic content should be in the locale files, yes. But data? I think that should be in `config/data`. I don't think that needs to be mentioned in the models.

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: Ah, right! I understand now. Yes, that makes sense.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+Manually testing user journeys from a page is difficult, particularly where code changes are spread across more than one application. To test locally, developers have to have multiple applications running which slows down their local machines.
+
+Having fewer apps would also mean that GOV.UK Docker, our system for running apps locally, could be simplified or potentially removed entirely.
+
+Sharing changes for non-engineering disciplines to preview is also difficult. This can be done by deploying branches to integration, but these branches are frequently overwritten by dependency updates.
+
+We can use Heroku for previews but this is also complicated by multiple apps. Previewing a user journey that involves pages from more than one app cannot be done in Heroku because links from a page in one app do not link to another. For example, a journey from the homepage to a search page and then a mainstream browse page crosses multiple app boundaries, and the links would 404 in Heroku. Previews can be done using Heroku, but this would be greatly simplified by having only a single application.
+
+### `static` and `slimmer` are complicated and inflexible
```
andysellick, line 90: Should this section be the same heading level as the ones above? Currently it's `###` but the other problems above are `####`. Then I think `The system is not designed for dynamic content` and `Code locations are scattered` should be headings beneath this section?

This shows the heading structure currently:

Screenshot 2024-06-18 at 09 25 10

leenagupte: You're right, and thanks for show the heading structure, that's really helpful. I think most of the headings are supposed to be a level 3, not 4. Will edit and fix up.

leenagupte: @andysellick I've updated the headings to better match the heading structure what we had in the Google doc. I've updated and fixed up.

andysellick: Cool. Now looks like this, which I think is correct?

Screenshot 2024-06-18 at 10 55 42

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+We propose merging the apps into a single rendering app. This will solve some of the issues in this RFC and allow us to solve others in the future - for example, removing `static`.
+
+We propose merging all apps into an existing app. The benefit of this is that even if we aren’t able to consolidate all of the applications, we would still reduce the maintenance burden for each app merged. We also won’t have created anything new, and won’t have to manage any extra conflicting patterns.
+
+We can steadily move routes, controllers and models from the candidate (the app being merged) into the host (the app we are merging into) until the host is serving all the routes.
+
+We propose using [frontend] as the host app. This is because:
+
+* `frontend` contains some of the more complicated document types like licence_transaction and local_transaction, that have multiple pages and interactive elements, which would be harder to consolidate into another app
+* `frontend` has a one-controller per document type layout, which is preferable for document types with multiple pages as it follows [rails conventions for controllers] to do the orchestration.
```
sihugh, line 139: The controller-per-document-type approach does help in a number of ways. In order to do this though, (somewhat unusually for GOV.UK apps) we use [a routing constraint based on document type](https://github.com/alphagov/frontend/blob/af88146e66d7693085e44a92fb4281c25ecf1400/config/routes.rb#L60). This retrieves the content item from content store as part of the routing of a request and checks the document type to determine which controller should handle it.

The content item is cached in the request object, so should only be retrieved once - even though it may be used multiple times in route resolution. Additionally, the controller for document types that use content items also [retrieves the content item from the request object](https://github.com/alphagov/frontend/blob/main/app/controllers/content_items_controller.rb#L25).

I don't think this is a problem, and might well be a useful approach when introducing new types to the app, however it is probably not well known so should probably be socialised at least. It would probably be a Bad Thing:tm: if our single frontend app made duplicate requests to our APIs!

leenagupte: Oh that's interesting point. Thanks Simon! We were concerned that we'd be retrieving the content-item twice. But you're right, getting the content item from the request object rather than an api call is a slightly odd pattern.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+
+There are currently 8 rendering apps for GOV.UK that are supported and maintained by GOV.UK teams:
+
+1. [collections]
+2. [email-alert-frontend]
+3. [feedback]
+4. [finder-frontend]
+5. [frontend]
+6. [government-frontend]
+7. [smart-answers]
```
mtaylorgds, line 25: Although technically a rendering app, Smart Answers works very differently to other rendering apps (it doesn't render items from the Content Store, for example), and I'm not sure it's code structure would fit into the suggested format very well.

I'm sure lots of folks are aware of this, but I just want to call this out as something that will probably need special attention.

leenagupte: Oh yes, we included it for completeness. My thoughts about smart-answers are that it probably needs to be deconstructed into 3 apps. The templates for rendering, a publishing app for the content, and a gem / engine / app for the business rules / logic that can be imported into both the publishing and rendering apps.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+    * will need to consider approach for developer machines
+* Need to reconfigure Kubernetes deployment. We should be able to drop the total number of pods.
+    * optimise the number of pods needed in draft and production
+    * currently in production there are
+        * 4 pods for `frontend`
+        * 6 for `government-frontend`
+        * 4 for `collections`
+        * 9 for `finder-frontend`
+        * plus 2 or 3 per app for the others
+    * scaling up the pods for a single app will use more resources
```
theseanything, line 258: I think the net resource usage would be roughly be the same. You'll have more "frontend" pods, but also you're removing pods for the other frontend apps - (also potentially router/router-api). Likely there will be higher memory usage per pod, but wouldn't have thought it'd be significant. Scaling the number of pods is a super simple configuration change that takes minutes.

theseanything: Essentially think this will be a non-issue - but defo worth keeping an eye on how the resources usage changes.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+* Optimise the frontend code
+    * remove duplication
+* Remove the routes from the candidate
+* [Retire the candidate] once all routes have been removed
+
+### Expected issues
+Merging all of our apps together is not without difficulties and we acknowledge that a single app will present problems that our multiple apps do not.
+
+* A Rails update for one large app might be more complex than spread across individual apps.
+* The scaling of a single app might become more costly.
```
theseanything, line 240: The cost of scaling these pods is likely be minimal vs our overall infrastructure spend.

DaveLee-GDS:
Whilst i don't feel confident in commenting on the coding patterns as such, it does look like a good idea to combine everything as you describe. It will certainly contribute to savings and efficiencies (coding-wise, not necessarily in ££) in future in unforeseen ways.
Without knowing more about the reasons why we have so many front-end apps now, i'll hazard a guess that different teams created them over time to solve timely problems (a la Conway's Law). Also, i'll assume different teams currently keep them up to date as necessary. My point being, once combined into a single codebase, we will need to consider how different teams contribute to that single codebase and how that should be handled. For example, will there be a set list of Reviewers or CodeOwners that PR approval is required of before merging any changes? Basically, how will the Ways of Working look? That is of course assuming my assumptions stand. If not, feel free to ignore me!

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+    * use as a test case and use that to inform our next steps
+* Write a roadmap for app consolidation
+    * will be handled separately
+    * might be simply the order in which applications will be merged into the host
+    * or could be some apps merged together before merging into the host
+    * or some routes merging ahead of others
+* Consider and work on expected issues (see below)
+* Work to merge remaining apps
+
+### Steps as part of a merge
```
dgm34, line 219: You've flagged that the apps you're proposing to merge have inconsistent approaches to testing. And merging codebases as proposed here is one of the times we'd really benefit from having comprehensive test coverage. I wonder if there's a case for rewriting a candidate's tests in RSpec and ensuring good coverage prior to any merge in to `frontend`?

KludgeKML: You may be interested in this: https://github.com/alphagov/frontend/pull/4090 (the PR that rewrites Frontend's test suite, since it's currently minitest). Of the candidate apps to merge in, all of them already have RSpec as their teset runner with the exceptions of government-frontend (the biggie) and smart-answers (which as other people have noted, probably isn't in scope for this realistically).

leenagupte:
> once combined into a single codebase, we will need to consider how different teams contribute to that single codebase and how that should be handled. For example, will there be a set list of Reviewers or CodeOwners that PR approval is required of before merging any changes? Basically, how will the Ways of Working look?

@DaveLee-GDS Those are really good points. At the moment the frontend apps are mostly owned by the web teams, so it's customary for the teams to ask the "owning" team for a review.

When the frontend apps are merged, who "owns" it will become trickier. I think an option could be to have "contributor guidelines" where the conventions of the app are documented. For example, how to add a new document type, what sort of tests should be added and where etc.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+    * change the rendering app for the document type in the publishing app.
+    * republish content
+    * republish any special routes (things that don’t have a publishing app, like the homepage) - any special routes should be moved into special route publisher (we have an outstanding tech debt card relating to this)
+* Optimise the frontend code
+    * remove duplication
+* Remove the routes from the candidate
+* [Retire the candidate] once all routes have been removed
+
+### Expected issues
+Merging all of our apps together is not without difficulties and we acknowledge that a single app will present problems that our multiple apps do not.
```
hannako, line 237: Will we see any of the benefits described from having less frontend apps, or will we only get the benefits when we have a single app?

I'm just looking at the number of different controllers in government frontend and feeling a bit scared about all that coming into frontend! We will have to be so ninja with reviews to prevent a wild west (recently wept whilst wrapping my head around [all these services](https://github.com/alphagov/collections/tree/main/app/services) in collections that are doing basically the same thing)

Have we considered other ways of combining to get the number down to 2 (or 3)? Or do you not think that's a viable option?

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+    * or some routes merging ahead of others
+* Consider and work on expected issues (see below)
+* Work to merge remaining apps
+
+### Steps as part of a merge
+We anticipate that any merge will likely involve some or all of the following steps.
+
+* Identify any issues with existing code
+    * size these issues
+    * either fix issues or record them as tech debt - we need to avoid trying to fix everything and concentrate on progressing consolidation
```
hannako, line 224: Candidate for should be fixed before moving over https://github.com/alphagov/government-frontend/tree/main/app/views/histories

kevindew (commented):
Awesome work here, It’s super exciting seeing work to improve our frontend architecture.

I’m very sold on the idea of combining together the applications that have the basic responsbility of rendering content store content however I’m not convinced it is a good idea to try have a single application for all of www.gov.uk. As it seems to me that it seems hard to justify the existence of both frontend and government-frontend is a good architecture decision, whereas it seems much harder to justify that merging all the email account management functionality of email-alert-frontend into frontend is going to offer much simplification benefits.

I also think that having a single app for www.gov.uk will be quite limiting, for example if this was already the architecture it’d be a bit of a nightmare for us to have to embed GOV.UK chat in frontend rather than a distinct app. Also if we had to have anything in a private repo we’d have to take the whole of www.gov.uk private.

I think it’d be prudent to reduce the scope of the RFC a bit, even merging the apps that seem very clear contenders is a project that’ll easily span north of a year. It’s really hard to imagine having the business runway that we could commit towards completing this project, just that we could expect to make gradual progress over time. For this reason I’m really pleased you’ve picked an existing app is a merge target as that reduces the risk. I can see that making frontend the dominant frontend app would offer the ability to explore retiring router.

I am quite skeptical of the ideas later in the RFC about engines, I can see how they (or packwerk) would be helpful in a really complex Rails app, but I don’t think we should have the goal of building an app so complex we need them. I think the fact that these might be useful may be an indication that we’re considering stuffing too much stuff into a single app.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+`content-store` registers routes in `router-api`. This means that the same content data is stored in multiple places and has the potential to get out of sync.
+
+There was an attempt to switch `router-api` to [use PostgreSQL] but this was paused indefinitely due to other priorities and the complexity of the work.
+
+### Local preview and sharing changes to GOV.UK is difficult
+Manually testing user journeys from a page is difficult, particularly where code changes are spread across more than one application. To test locally, developers have to have multiple applications running which slows down their local machines.
+
+Having fewer apps would also mean that GOV.UK Docker, our system for running apps locally, could be simplified or potentially removed entirely.
+
+Sharing changes for non-engineering disciplines to preview is also difficult. This can be done by deploying branches to integration, but these branches are frequently overwritten by dependency updates.
```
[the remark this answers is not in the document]
leenagupte: Oh, that's much better thanks. I'll add a commit to update the paragraph.

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: I think anything that reduces the number of apps even a little bit is a benefit as it reduces the maintenance burden, so 5 apps is better than 8 etc.

The benefit of the single app is being able to remove static and slimmer. 

We will have to be quite strict in our reviews and making sure everything is properly namespaced, and in the agreed area of the code. I think that having contributor guidelines that details what is finally decided on here should make that a bit easier as there will be something to point to when doing a review.

When you say "other ways of combining" what do you mean?

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: Ah, yes this content should be moved into a locale file. I think that can be "fixed" as part of the migration.

maxgds:
Personally I'm not averse to a world of "many brains and one face" because my pain points are naturally around the rendering aspects. If we can find another way to a solution where we can run up a test instance that works for retracing user journeys end to end, can preserve states like logged in/accepted cookies, the header and footer know where the user is, it has one set of translations, and doesn't need to re-deliver assets repeatedly then I'm open to it. My end goal is probably consolidated rendering and I'm less bothered (but not unbothered) by the distributed dependency updates and so on.

leenagupte:
Hi @kevindew. Thanks for your comments. You made some really interesting points. I'm going to respond to them individually, so please bear with!

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+### How the proposal will be carried out
+This is obviously a very large piece of work that will take some time to complete. We have identified the following steps that will be needed.
+
+* Prepare `frontend` to be the host
+    * convert frontend tests to RSpec
+    * modernise the tests
+    * audit `frontend` and make changes to the frontend directory structure to match “new layout”
+    * remove redundant code from frontend (non-blocking)
+    * remove publishing_api tasks from frontend (non-blocking)
+* Merge `feedback` into `frontend`
```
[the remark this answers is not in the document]
leenagupte: Feedback was picked as a proof of concept for a few reasons:

1. It's a relatively small app with very few routes.
2. A lot of the functionality originally existed in frontend but was [moved back](https://github.com/alphagov/feedback/pull/1601) due to it being impossible to render a success message.
3. Frontend already renders content items with dynamic behaviour.
4. There are some complex bits in it, so lessons can be learnt before tackling something as complex as finder-frontend.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+Under `/app`:
+
+`models`
+- Frontend apps do not have database records, so models here are created to store data received from api calls (eg content items from content store), process data that will be written to apis (eg feedback items to zendesk), and information ingested from YAML/JSON/CSVs.
+
+`services`
+- Api calls for search and content-store, and apis external to GOV.UK
+
+`presenters`
+- Presentation related transformations on model data. It is a wrapper around a model for use in a view.
```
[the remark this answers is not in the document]
leenagupte: > Can we look into whether what we're considering as presenters in frontend apps is actually just models?

Yes! I think we should definitely look at this when evaluating the apps, especially `government-frontend`. I think some of the "presenters" in `frontend` are really models too (see [calendar_content_item](https://github.com/alphagov/frontend/blob/main/app/presenters/calendar_content_item.rb) 😱 ). 

Ideally the data from content-store would be retrieved and handled in a model before being passed to a presenter to make it more readable, however the routing constraint (mentioned in another comment) makes that impossible.

I find that the line between models / view helpers / presenters can get a bit blurry.
I think we agree that there shouldn't be any presentation logic in the model, and that before using a presenter, consider whether a helper will do.

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: I think whatever the solution to `smartanswers`, it will merit its own RFC. 

For this RFC, `smartanswers` was included for completeness. Realistically I think we could get to a place where most of the rendering is done in `frontend` and `frontend` does the redirects to `smartanswers` routes removing the need for `router` and `router-api`.

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: Oh, that's a shame, and also good to know. 😬

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+Even when applications use the same testing frameworks there are still a lot of inconsistencies in how the tests are set up, for example in how fixture files are used, or how `content-store` and `search-api` are stubbed.
+
+### Duplicated assets and complex asset model
+Our assets (CSS and JavaScript files) are well optimised but having multiple apps means that some problems remain. If a user journey involves more than one app this means that duplicate page assets can be downloaded, reducing the speed of the site.
+
+For example if a user visits a page in one app that requires the button component CSS and then moves to a page on another app with the same component, that CSS file is downloaded once per page, rather than once.
+
+We aim to build our pages and assets to be as performant as possible, but repeated downloads invalidate some of that effort. This situation happens for any repeated components across the estate, where each app will re-deliver CSS and JavaScript to the user unnecessarily.
+
+### `router` / `router-api` are complicated and hard to maintain
```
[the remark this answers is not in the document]
leenagupte: I'll add a section. The thought is that if we can get down to say 2 apps, e.g. `frontend` and `smartanswers` then `frontend` could just redirect routes to the `smartanswers` removing the need for `router` and `router-api`.

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
kevindew: > I think we agree that there shouldn't be any presentation logic in the model, and that before using a presenter, consider whether a helper will do.

Yup agree there 👍  Rails does make this area rather opaque.

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
kevindew: Ah so do you mean have frontend act as proxy similar to how router is?

I imagine the ideal would be if we had a more conventional proxy routing system where we used a path prefix for apps other than frontend, so /chat to chat, /email to email-alert-frontend, /smart-answers for Smart Answers then everything else to frontend. Then we could get rid of the need for the complex system. Only thing that seems to be problematic is the existing URLs and how divisive it'd be to change them (with Smart Answers seeming easily most divisive there)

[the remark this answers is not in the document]
theseanything: Agree, if have still have separate frontend apps. A path prefix would make it super easy to hand routing off to the load balancer. (I know the content folk like to have easy human readable URLs, but there needs to be compromise between the value that brings users vs engineering complexity here). Dealing with existing URLs would be tricky, but not impossible.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+
+We can use Heroku for previews but this is also complicated by multiple apps. Previewing a user journey that involves pages from more than one app cannot be done in Heroku because links from a page in one app do not link to another. For example, a journey from the homepage to a search page and then a mainstream browse page crosses multiple app boundaries, and the links would 404 in Heroku. Previews can be done using Heroku, but this would be greatly simplified by having only a single application.
+
+### `static` and `slimmer` are complicated and inflexible
+GOV.UK page layouts are rendered by `static` using the [slimmer] gem. This includes the header and footer common to all pages, even though the code for those elements comes from [govuk_publishing_components]. Having a single application would allow us to remove our dependency upon `static`.
+
+`static` and `slimmer` are both quite complicated and not well understood among the current developer community. The details are beyond the scope of this document, but several specific problems can be identified.
```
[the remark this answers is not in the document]
leenagupte: Eventually being able to remove `static` and `slimmer` is definitely one of the points of this RFC and a goal of these changes. 

We're in a bit of a strange situation where the layouts live in `govuk_publishing_components`, but are called via `static`. In the case of the [standard layout](https://github.com/alphagov/static/blob/03878acd740d0407fb35c807d5c53f7dccebf320/app/views/root/_gem_base.html.erb), the only extra work that's happening is rendering the global and emergency banners.

The current setup of `static` and `slimmer` where the templates are passed about as string, also make it very difficult to have dynamic headers. This is something that could cause a bit of pain when we need to integrate with One Login.

But the details of how static and slimmer are removed or replaced should have its own discussion.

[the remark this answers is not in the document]
kevindew: It does seem like that is many years off to combine the apps to a point where you could kill static, I think it'd be much more pragmatic to investigate killing static off without needing to merge all the apps if that is a bigger prize than just having less apps. Having to consolidate all the apps as a pre-requisite to not needing static does seem a bit of nuclear option. 

Like you say, it doesn't really seem to do a lot anymore (I hadn't realised how little, I thought it still had links configured in it, but they seem ). It seems that the only benefit of having apps use that over just rendering the layout themselves is that it means we only have to coordinate statics update of govuk_publishing_components over having to coordinate updating each frontend app when header or footer change (as an aside I do think it's quite an architecture smell the degree of content in govuk_publishing_components, really think we could benefit from apps able to inject that in from a content item).

I would be inclined to suggest that as part of the work to try make frontend the "main" frontend app, we try just dropping static from it and seeing how things go with using govuk_publishing_components directly and see if we can avoid waiting for the long migration.

rfc-174-consolidate-frontend-rendering-apps.md
KludgeKML: Do we know the history of why the history pages aren't just normal content items? Is there a way we can remove them entirely from government-frontend and turn them into editable items?

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+* Decide how to handle app ownership and maintenance and what that means now we have fewer or only one app
+* Consider renaming `frontend`
+
+## Other options to explore
+
+We write this knowing that new technologies are constantly emerging that may prove useful in this area. One such technology is [Rails Engines], which allows you to wrap a specific Rails application or subset of functionality and share it with other applications or within a larger packaged application.
+
+As a separate piece of related work, we propose exploring Rails Engines as follows.
+
+### Option 1: Engines in existing app
```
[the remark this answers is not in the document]
KludgeKML: There's a conceptual difference between a monolithic app and a monolithic repo - for instance we're already committed to a situation where some updates are done in gems, some in apps (this isn't a perfect analogy, of course, because generally we're hoping to share the gem code). It comes down a bit to which you think is more serious : the architectural problems caused by multiple apps or the developer-mindspace problems caused by multiple apps. It's worth bearing in mind too that this app (unlike, say, Whitehall), is likely to end up with multiple teams working on it, so separate repos in engines do have a slight cleanness advantage over one bit repo. 

I don't want to imply I'm married to the idea of engines, though. We could just remove the options section, or reframe it as "options we have considered and rejected".

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+* As with a monolith, Layouts could be in the “host” app, simplifying the engine code and allowing us to remove the static/slimmer app/gem complexity.
+* As with a monolith, the single host app would be the only one that had to be deployed, allowing us to reduce the number of apps/config in Kubernetes.
+* Re-engineering into engines would allow us to split functions in a more granular way than they have been with apps - for instance, it might make sense for frontend to be more than one engine, perhaps an engine for homepage, an engine for simple smart answers, and an engine for location-based content types like places and local_transactions, etc.
+
+Disadvantages of Engines:
+
+* We’d still have to know exactly which engine served a given route, so there’s still a little knowledge overhead in working out where changes/fixes would need to be made.
+* Little existing experience with engines in GOV.UK (the component guide is an engine, but it doesn’t get changed much and isn’t used outside of development)
+
+### Option 2: Engines in new app
```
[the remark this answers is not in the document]
KludgeKML: (We've simplified this section to "options considered", so I'm going to mark this as resolved as it's not a proposal for work)

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Option 1: Engines in existing app
+To consider converting the existing apps into engines that can be mounted inside a single host app. This would allow us to keep separate repos for each current application while still simplifying the overall architecture.
+
+Advantages of Engines:
+
+* Separate repos provide a natural encapsulation of code/tests, meaning that a test suite could be kept for just the routes mentioned in the engine, keeping test suites run times lower.
+* As with a monolith, Layouts could be in the “host” app, simplifying the engine code and allowing us to remove the static/slimmer app/gem complexity.
+* As with a monolith, the single host app would be the only one that had to be deployed, allowing us to reduce the number of apps/config in Kubernetes.
+* Re-engineering into engines would allow us to split functions in a more granular way than they have been with apps - for instance, it might make sense for frontend to be more than one engine, perhaps an engine for homepage, an engine for simple smart answers, and an engine for location-based content types like places and local_transactions, etc.
```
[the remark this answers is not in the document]
KludgeKML: (We've simplified this section to "options considered", so I'm going to mark this as resolved as it's not a proposal for work)

rfc-174-consolidate-frontend-rendering-apps.md
KludgeKML: > Don't discount merging it into a single app with both rendering and an admin area, would be way simpler to build than multiple apps.

I feel like we've recently seen a bunch of work in Whitehall to get away from the publishing/rendering in a single app paradigm, how would a single admin + rendering smart answers differ from that? We have things that are intermediate to that (eg places-manager, where we have an API consumed by the frontend and the admin console for that API in one app). A Smart-Answers manager, maybe, and then smart answers would get published as normal artifacts and attached to an API object? (in the same way a Place is an item in Publisher that is published as normal, but attached to a service in Places Manager which provides the backing data)

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
KludgeKML: As a stopgap, could Frontend (instead of Router) read from Router API and route to the other apps? What additional load would that put on the system? We're pretty heavily cached by the CDN, right? What would be the effect on the end rendering time if a rails app did the routing rather than the go app? I imagine it would be vastly slower, but does vastly slower actually equate to much in terms of what the end user sees?

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
KludgeKML: Updated (so will mark resolved). Thanks Jess!

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+
+The relationship between [router and router-api] is quite tangled. `router-api` is a Ruby on Rails application. The database that it shares with router is a Mongo database.  The version of Mongo we’re using is an old one, and has had to be pinned to keep [compatibility with the Mongo driver].
+
+`content-store` registers routes in `router-api`. This means that the same content data is stored in multiple places and has the potential to get out of sync.
+
+There was an attempt to switch `router-api` to [use PostgreSQL] but this was paused indefinitely due to other priorities and the complexity of the work.
+
+### Local preview and sharing changes to GOV.UK is difficult
+Manually testing user journeys from a page is difficult, particularly where code changes are spread across more than one application. To test locally, developers have to have multiple applications running which slows down their local machines.
+
```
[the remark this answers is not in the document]
KludgeKML: Updated, so marking as resolved.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+`static` and `slimmer` are both quite complicated and not well understood among the current developer community. The details are beyond the scope of this document, but several specific problems can be identified.
+
+#### The system is not designed for dynamic content
+The page template generated by `static` is passed as a string which is then [parsed by Nokogiri] into an object. This means that the header and footer can’t be dynamic without complex workarounds. This has presented problems with any efforts to integrate www.gov.uk with the One Login programme.
+
+#### Code locations are scattered
+The code for the page layout consists of components from `govuk_publishing_components`. Unlike other components that are called from applications directly they are rendered by `static`. This is overly complicated - if there was a single app, the layout components for that app would be included in it, and making changes would involve a single pull request. Currently this involves 3 pull requests, including publishing a new version of the components gem.
```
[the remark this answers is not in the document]
KludgeKML: Updated (marking as resolved)

rfc-174-consolidate-frontend-rendering-apps.md
kevindew: I'd not use Whitehall as a gold standard in this area because it spent many years being an app that wasn't rendering about 90% of the content and just 10%. I think if it had been the other way around and Whitehall was still rendering 90% of the content we'd have probably just migrated the other way to get it back to 100%. The debts of that system were always the inconsistencies and sheer complexity.

I think we shouldn't form a holy rule that publishing and rendering must be separate. Obviously we've invested a lot in having an architecture that supports separation and if we have basic content that goes into Publishing API and then a rendering app that just pulls data from content store same way as other apps then you'd be best continuing that pattern. But for something like Smart Answers where we've got complex modelling system, ability to define custom logic and a relatively small number of domain concepts, surely the overhead of having multiple apps to keep in sync would be a burden more than a benefit? Also seems unlikely that you'd want separate teams to have frontend and backend given the heavy coupling.

Obviously this is all rather academic at this point given the many years of desire for a Smart Answer management system and lack of solutions mind 😄

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
kevindew: I don't know for sure but I guess it'd be a humongous load difference given that router deals with all the 404s and then government-frontend gets the most app traffic. Whether we can cope with it though seems like something we could work out, particularly if we could applying auto-scaling given the app is pretty much stateless so should scale well. 

One thing I'm not clear on is whether Router conceals us a bit from some of the Ruby pains we have of only able to serve as many users as the number of threads we have per server, which can lead to requests in a backlog and eventually just aborted. But given Router already proxies to apps with that limitation I can't imagine it is that impactful.

What I'd probably suggest as a stop gap though is that rather than having Frontend start take Router API responsibilities you instead set Router to forward everything it doesn't know what to do with to Frontend, so Frontend can gradually take on handling error pages etc. You could then stop indexing in Router API for things to frontend so that the Router API database gradually gets smaller and smaller to be less impactful.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+└── .gitignore
+```
+
+Under `/app`:
+
+`models`
+- Frontend apps do not have database records, so models here are created to store data received from api calls (eg content items from content store), process data that will be written to apis (eg feedback items to zendesk), and information ingested from YAML/JSON/CSVs.
+
+`services`
+- Api calls for search and content-store, and apis external to GOV.UK
```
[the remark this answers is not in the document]
KludgeKML: Yeah, I think I phrased this line rather vaguely and included bad examples. Looking at the contents of the services folders currently in frontend and government-frontend there's only two things that really fit the NounVerb style: One where it's highly dubious and could be a lib class (DocumentBuilder) and one we want to get rid of anyway (CalendarPublisher), so I'll rewrite this to just remove services and in the full guidance we'll suggest avoiding it.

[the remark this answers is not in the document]
KludgeKML: (Updated, marking as resolved)

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
KludgeKML: I think we're _already_ doing that with government-frontend, but I may be wrong.

rfc-174-consolidate-frontend-rendering-apps.md
KludgeKML: Well, it wouldn't end up different teams if it were like Places (there the published item is quite light, and the backing service API management is the main thing, managed by a single team). But sure. We're probably not going to solve the problem of Smart Answers in this comment thread ;)

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
leenagupte: Removing static is a problem that has been investigated before: https://github.com/alphagov/govuk-rfcs/pull/84
It comes a bit unstuck when it comes to the banners. If static is replaced with a gem, then you have the issue of having to do a gem update and deploy in 8 rendering apps. That's better now that we have auto-merging and deployment on most of the apps, but could still leave things in an inconsistent state if a deployment fails.

There also the issue of duplicating assets if static is removed now before doing any consolidation.

I think that consolidating more apps gets us closer to the place where decommissioning static starts looking possible.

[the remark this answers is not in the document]
kevindew: Yeah there's further work after that RFC too:

https://github.com/alphagov/govuk-rfcs/pull/118 - which explored that content and other global aspects aren't hardcoded but are part of data from Content Store, which would allow banners to go out without being hardcoded.

And for assets there is the unimplemented approach to use a shared folder from a couple of RFCs: https://github.com/alphagov/govuk-rfcs/blob/109d05df9dcb9f06945ffa512f65d185b05f45a4/rfc-149-switch-to-per-page-asset-loading.md#make-assets-load-from-a-shared-folder and https://github.com/alphagov/govuk-rfcs/blob/109d05df9dcb9f06945ffa512f65d185b05f45a4/rfc-091-sharing-assets.md

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,308 @@
...
+* Begin work to remove static, slimmer and router (to be considered separately from this RFC)
+* Update component auditing
+* Decide how to handle app ownership and maintenance and what that means now we have fewer or only one app
+* Consider renaming `frontend`
+
+## Other options considered
+
+### Convert apps to engines to mount in Frontend
+
+If we wanted to keep the conceptually split functions of the apps but remove the rendering complexity of slimmer/static/router we could recreate the routes and business logic of the existing apps in Rails Engines that would be mounted inside the monolith. This option was rejected on the basis that it would not give us much code-separation advantage over simple namespace discipline within the monolith, and there would still be developer overhead in working with multiple repos.
```
richardTowers, line 277: We're trialling a new approach in whitehall, where we separate bits of code out into engines, but store them all in the same repo in `lib/engines`. There's only one example of this at the moment, [the content object store](https://github.com/alphagov/whitehall/tree/main/lib/engines/content_object_store).

This would allow us to use rails engines without the overhead of working with multiple repos.

It might be worth considering as a stepping stone, but I don't like it as a long term solution for merging frontends, because it retains the arbitrary split between "apps" (even if they're no longer actually apps).

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,308 @@
...
+
+## Summary
+We have too many frontend rendering applications. This increases complexity, reduces understanding and slows development. We propose consolidating our apps into a single app. This will improve standardisation, reduce maintenance costs and improve the developer experience.
+
+## Problem
+
+### Applications overview
+GOV.UK is served by multiple rendering applications (apps). Each app serves different groups of pages within the website.
+
+There are currently 8 rendering apps for GOV.UK that are supported and maintained by GOV.UK teams:
```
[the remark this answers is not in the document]
richardTowers: Although if you accept @kevindew 's suggestion that we focus this RFC on the top 4 most obviously mergeable apps, it's fine to ignore these edge cases.

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: A reason why I think we could break up smart-answers is because we already have something similar for simple-smart-answers. Yes, it's a much simpler wizard, but it is configured in a publishing app, but the rendering is separate. 

I think issue with regular smart-answers has always been the amount of developer toil to update the rates and content that really should be editable by content designers. So, I do think it needs to be torn up and re-thought, especially as the developer numbers aren't what they used to be.

I wouldn't ever take the conversation about smart-answers off the table. The scope of it is too large to be covered by this RFC, but it does do frontend rendering and needs to be considered with the other apps.

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: Oh they're published by [government-frontend](https://github.com/alphagov/government-frontend/blob/main/lib/publish_static_pages.rb) at the moment 😬 

It is a bit of a tangled mess what came first? [the template or the content item](https://github.com/alphagov/government-frontend/blob/main/lib/publish_static_pages.rb#L92)? 😬 

As a first step the task that publishes the pages should be moved to a publishing app. They at least do for the most part seem to follow a pattern: breadcrumbs, contents list, image, govspeak.

leenagupte:
@kevindew  thanks for your review. I'm going to respond to your big comment here, and then continue responding to the rest of them inline.

> it seems much harder to justify that merging all the email account management functionality of email-alert-frontend into frontend is going to offer much simplification benefits

There's a lot of weirdness happening with the email pages as the journey can be split over multiple apps, making it harder to test.

For example, if you want to sign up for emails about a taxon, you start in collections. Clicking on the "Get emails" link takes you to a page rendered by [email-alert-frontend].

If you're on the News and Communications finder and click on "Get emails" you're taken to a similar but different page rendered by finder-frontend, before continuing further into the email-alert-frontend journey.

There's also some other areas of [tech debt] that would be nice to tackle.

A lot of these issues would be easier to fix if the rendering was in one place. We could at least consolidate the templates the user is seeing and trying to smooth out the user journey from that perspective even if the routing in the backend is still all over the the place.

> it’d be a bit of a nightmare for us to have to embed GOV.UK chat in frontend rather than a distinct app

Unless I'm mistaken GOV.UK Chat purposefully doesn't look like the rest of GOV.UK? It's supposed to feel separate right? I also think that it doesn't have any of the same page furniture as a regular GOV.UK page? Is that right?

If that's true it feels fine to keep Chat separate. If it was truly going to be embedded into user journeys in GOV.UK we'd have to properly think about how / where it fits in any case.

> Also if we had to have anything in a private repo we’d have to take the whole of www.gov.uk private.

It's always felt weird that we do this. As in the content should be in draft, rather than github repos being made private.

> I think it’d be prudent to reduce the scope of the RFC a bit, even merging the apps that seem very clear contenders is a project that’ll easily span north of a year.

Yes, as with every other GOV.UK migration it's unlikely that we'd be able to complete the consolidation. However we've decided to be upfront with that. The goal is to reduce the complexity of the frontend stack. So even consolidating 2-3 apps can be considered a win. Starting with an existing app means that we won't start be adding even more tech debt. It should just be a case of slowly reducing the debt.

[email-alert-frontend]: (https://www.gov.uk/email-signup/confirm?topic=%2Feducation%2Ffurther-and-higher-education-skills-and-vocational-training&link=%2Feducation)
[finder-frontend]: (https://www.gov.uk/search/news-and-communications/email-signup?people%5B%5D=mark-stanhope&order=updated-newest)
[tech debt]: https://trello.com/c/eSxYNXB2/13-finders-and-finder-email-signup-pages-use-different-facets

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,330 @@
...
+6. [government-frontend]
+7. [smart-answers]
+8. [static]
+
+Each app is built in a similar way.
+
+* written in Ruby on Rails
+* no database
+* relies on data from API calls to populate pages
+* calls the content API from [content-store] to get the basic page data
```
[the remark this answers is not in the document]
leenagupte: > I think there's probably a couple of categories of apps, and we could do with being explicit about whether we're consolidating them all, or just the ones that match these similarity criteria. 

I'll update this list. When we're talking about consolidating apps, we meant the apps that render pages or parts of pages to the user (general public), so not APIs.

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
leenagupte: Being better in a better position to remove static and slimmer is one of the justifications of doing this work. The less rendering apps we have to deal with the more options we have. I don't we can omit mention of them entirely.

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
leenagupte: I think that rather than dropping the router section, we should add another that says how far we towards solving the problems with this RFC.

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
leenagupte: The APIs were ignored as they don't render what I'd think of user facing content, i.e. web "pages". But yes, given that those routes all follow a pattern, it would be easy to add a redirect for them in frontend.

Licensify is also out of scope as it's a Scala app that we (GOV.UK teams) don't maintain. I'd be happy for it to move to it's own domain though.

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,308 @@
...
+
+### How the proposal will be carried out
+This is obviously a very large piece of work that will take some time to complete. We have identified the following steps that will be needed.
+
+* Prepare `frontend` to be the host
+    * convert frontend tests to RSpec
+    * modernise the tests
+    * audit `frontend` and make changes to the frontend directory structure to match “new layout”
+    * remove redundant code from frontend (non-blocking)
+    * remove publishing_api tasks from frontend (non-blocking)
```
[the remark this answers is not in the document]
leenagupte: I think the bank holidays config should move to mainstream publisher. Is there a better app for it?

rfc-174-consolidate-frontend-rendering-apps.md
```diff
@@ -0,0 +1,308 @@
...
+* Remove the routes from the candidate
+* [Retire the candidate] once all routes have been removed
+
+### Expected issues
+Merging all of our apps together is not without difficulties and we acknowledge that a single app will present problems that our multiple apps do not.
+
+* A Rails update for one large app might be more complex than spread across individual apps.
+* The scaling of a single app might become more costly.
+* It won’t be possible to scale targeted parts of the stack (i.e. one of the existing apps) for expected increases in traffic, instead the whole site will need to be scaled.
+* It’s only possible to test one branch of code in a test environment at once. If there’s only one repo, only one change can be tested at a time.
```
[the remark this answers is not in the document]
leenagupte: That would be nice. That would also fix the problem we have at the moment with broken user journeys in the review apps.

rfc-174-consolidate-frontend-rendering-apps.md
leenagupte: Yeah, if we're going to do this in frontend I'd want to be based on functionality rather than the historical app.

rfc-174-consolidate-frontend-rendering-apps.md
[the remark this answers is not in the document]
richardTowers: Yeah, mainstream would be fine I reckon.

PR #175, opened 2024-08-05
RFC-175: Reduce number of frontend apps

KludgeKML:
Refinement of RFC-174 with reduced scope, proposing merging: collections, finder-frontend, and government-frontend into the frontend app.

Rendered version: https://github.com/alphagov/govuk-rfcs/blob/frontend-fewer-apps/rfc-175-frontend-fewer-apps.md

Closing date for comments: Thu 22nd August 2024

https://trello.com/c/54amoVVe/242-draft-new-reduced-scope-rfc-for-consolidating-apps


rfc-000-frontend-fewer-apps.md
```diff
@@ -0,0 +1,256 @@
...
+* It won’t be possible to scale targeted parts of the stack (i.e. one of the existing apps) for expected increases in traffic.
+* It’s only possible to test one branch of code in a test environment at once. If there’s only one repo for these four apps, only one change can be tested at a time.
+* It will be harder to find the pull request history for the retired repositories.
+
+In addition, we will need to address the following issues:
+
+* The test suite will take longer due to more tests.
+    * could potentially parallelise in CI
+    * will need to consider approach for developer machines
+* Need to reconfigure Kubernetes deployment. We should be able to drop the total number of pods.
```
leenagupte: In RFC-174 Sean commented that the increase in cost is negligible, so we probably don't need to bother with this.

KludgeKML: Have removed scaling cost references.

rfc-175-frontend-fewer-apps.md
```diff
@@ -0,0 +1,254 @@
...
+Again, retiring [static] by merging all the apps relies on all apps being merged, something that is far enough in the future that we’ve decided to explore other options for retiring [static]
+
+## <a name="full-exploration"></a>Full Exploration of the Problem
+
+### Applications overview
+GOV.UK is served by multiple rendering applications (apps). Each app serves different groups of pages within the website.
+
+There are currently 8 rendering apps for GOV.UK that are supported and maintained by GOV.UK teams, 4 of which are almost exclusively concerned with rendering content:
+
+1. [collections]
```
ryanb-gds, line 39: I don't think there's any reason this couldn't be merged to frontend so it's a bit out of scope of the RFC really, but do you see the merged frontend app as the long term home of the organisations API, which currently resides in collections?

KludgeKML: What currently uses that API?

ryanb-gds: [A fair few things](https://github.com/search?q=org%3Aalphagov+GdsApi%3A%3AOrganisations&type=code), although I think a lot of them are no longer active. Manuals Publisher and Short URL Manager are though

KludgeKML: Ah, I see it's here: https://github.com/alphagov/collections/blob/main/docs/api.md - mainly internal apps. Good question! It's probably not a terrible place for it to be, given that it's publicly available information and other people _might_ potentially use it.

rfc-175-frontend-fewer-apps.md
```diff
@@ -0,0 +1,254 @@
...
+status: proposed
+implementation: proposed
+status_last_reviewed:
+---
+
+# Consolidate Frontend Content-Rendering Apps
+
+## Summary
+
+This is a refinement of the suggestions in [RFC-174] about reducing the number of frontend rendering applications, reducing its scope to the applications that deal exclusively with rendering content items. By dealing only with the four applications that exclusively render content items, we improve standardisation and reduce maintenance costs, and the work is worthwhile even if we can only achieve sections of it. Because a large part of the problem and a lot of the solution remains the same, we have divided this RFC into two parts - an initial section that describes how we differ from [RFC-174], and a more verbose section detailing the approach. If you have already read or commented on [RFC-174], you can read the [How this proposal differs](#how-this-proposal-differs) section and skip the rest of the RFC. If you are new to this proposal, feel free to skip to the [Full Exploration of the Problem](#full-exploration) section.
```
ChrisBAshton, line 11: Is it worth being explicit that RFC-174 was closed and not accepted? The language in this RFC made me assume 174 had been merged and then descoped here in 175.

ChrisBAshton (approved):
LGTM 👍

rfc-175-frontend-fewer-apps.md
KludgeKML: Ah yes, good point - and we should properly merge 174 in marked as rejected(? or whatever we do with rejected RFCs)

KludgeKML: Done.

rfc-175-frontend-fewer-apps.md
KludgeKML: In some ways, the merged app would become a more appropriate place for it, because it's somewhat like the search results in Finder Frontend. Perhaps Search API is a natural home? I think it's a tricky question, because there's no particular good home for it, so maybe for the moment it makes as much sense to keep it here as anywhere. But when the routes are moved, we'll consider if there's a better place - we don't want to treat the grouping of the routes in the four current apps as handed down to us on stone tablets.

rfc-175-frontend-fewer-apps.md
```diff
@@ -0,0 +1,254 @@
...
+
+`tasks`
+- Rake tasks
+
+Use `lib` rather than `app/lib` for custom code that enhances the app but doesn’t fit into a directory in `app`
+
+Under `/config`
+
+`data`
+- YAML, JSON, CSV config files. Data files should be placed here rather than under `/lib/data`.
```
kevindew, line 150: In the time since the last RFC we on the AI team have had to deal in anger with considering the place for `data` files and found the `config/data` route was actually a bit surprising and `lib/data` was a better choice.

Our rationale was:

1. when people tend to be reaching for a `data` directory it tends to not actually be things that configure the app - so config/data seemed surprising for something that isn't doing any configuration
2. when putting things into the config directory you would typically want them to be accessible via `Rails.configuration` to be conventional config, and it seems strange to potentially have a directory of files that isn't
3. Rails already has the precedent for non namespaced storage in lib, with the assets and tasks precedence. It's really easy to flag a directory to be ignored by the autoloader: https://github.com/alphagov/govuk-chat/blob/54059d50aff6211328fb4f8302333adaaa03b02d/config/application.rb#L26-L29

The tricky bit is the blurring of the line between what is config and what is data - which you'll probably have even with this decision to choose between whether a file lives in `config` or `config/data`. The barometer we found worked well was whether it makes sense to be part of `Rails.configuration`

rfc-175-frontend-fewer-apps.md
```diff
@@ -0,0 +1,254 @@
...
+`locales`
+- Static content like lists of contact information should ideally be stored in the locale files to promote localisation.
+
+Namespaces will be used to group functionality e.g. for StepBySteps, and resources in the routes file. This structure will be documented in README of the host app.
+Each application will be analysed before being consolidated to ensure redundant code is not being moved over.
+
+### How the proposal will be carried out
+This is obviously a very large piece of work that will take some time to complete. We have identified the following steps that will be needed.
+
+* Prepare `frontend` to be the host
```
kevindew, line 161: Not sure if this is the right RFC or https://github.com/alphagov/govuk-rfcs/pull/176 but will this involve frontend taking over static responsibilities like error pages and common assets like favicon?

rfc-175-frontend-fewer-apps.md
```diff
@@ -0,0 +1,254 @@
...
+### After merging
+Once a candidate has been merged, the following steps would be needed.
+
+* Recalibrate Kubernetes deployment
+* Move components from govuk_publishing_components into the app if only being used in the frontend
+    * this information can already be found in our component auditing tools
+    * need to consider which ones are used by publishing applications
+* Update component auditing
+* Decide how to handle app ownership and maintenance and what that means now we have fewer apps.
+* Consider renaming `frontend`
```
kevindew, line 224: Time to reclaim govuk-frontend 💪

kevindew: Or could fall back to the old suggestion from https://github.com/alphagov/govuk-rfcs/blob/main/rfc-095-long-term-future-apps.md of `content-frontend`

kevindew (commented):
This sounds great @KludgeKML, thanks for putting this together. I really looking forward to seeing this come together, it feels like working on GOV.UK would be so much easier with these things combined.

rfc-175-frontend-fewer-apps.md
KludgeKML: The hardest problem in computer science, naming things!

leenagupte: This is the kind of thing that should probably be put to a vote. I'll have to veto "frontendy-mcfrontend-face" though 😉 .

rfc-175-frontend-fewer-apps.md
KludgeKML: For the moment I think that will be part of the work in RFC-176, but yes - current suggestion is that this will go in the static content gem, but it might be that during the implementation of that that we discover it makes more sense for some of those to got into Frontend. I think I'd like to keep that open (but I will mention it in RFC-176, which I'm just updating now). Until we can work up some proofs of concept, I'm reluctant to commit to moving them from one single app to another single app (just because there's outstanding tech debt related to nginx clobbering HTTP responses from the frontend apps, and part of solving that is making sure that some error responses are devolved to individual apps), but there will be a decision gate where we work out where it ends up.

rfc-175-frontend-fewer-apps.md
KludgeKML: Hmm. Looking at the current use of lib/data, only frontend uses it really (two apps use it for recruitment banner config, but as per RFC-176 we're looking to remove that code in app anyway), and it uses it to hold two content items that it publishes. (bank holiday and clock change content items). Given that those should probably not be in Frontend anyway, I think this might be a moot point in practice. In theory, though, we could have things in both directories depending on what they were?

leenagupte: > whether it makes sense to be part of `Rails.configuration`

I think that statement is open to a lot of interpretation too.
There doesn't seem to be a consensus in the wider world about where to data files either. It might be better to have a top level `/data` directory for model data.

rfc-175-frontend-fewer-apps.md
KludgeKML: Thanks for bringing this up!

rfc-175-frontend-fewer-apps.md
KludgeKML: I've added a note here that /config/data should be for data that's clearly for configuration, a root data directory `/data` for model data, but that in general we should try to avoid using data files for configuration in these apps.

rfc-175-frontend-fewer-apps.md
kevindew: Ok cool. I'm not quite sure I understand how the gem makes it easier because I would have thought that, without static, we don't want every app to be able to render things like favicon etc and just want one nominated one - since only one app should do www.gov.uk/favicon.ico for instance, although error pages might be different - but very happy to move that discussion away from this RFC since it'll be covered in the static one.

rfc-175-frontend-fewer-apps.md
```diff
@@ -0,0 +1,260 @@
...
+Under `/lib`
+
+`tasks`
+- Rake tasks
+
+Use `lib` rather than `app/lib` for custom code that enhances the app but doesn’t fit into a directory in `app`
+
+Under `/config`
+
+`data`
```
kevindew, line 150: Sorry to nitpick further, if you've got a separate /data directory surely you'd want this config just in /config directly - shouldn't be any need for a separate data directory.

rfc-175-frontend-fewer-apps.md
```diff
@@ -0,0 +1,260 @@
...
+- YAML, JSON, CSV config files. Data files should be placed here if they are clearly used in the configuration of the app. Ideally, though, we should **avoid using data files** for configuration where possible (see `/data` below)
+
+`locales`
+- Static content like lists of contact information should ideally be stored in the locale files to promote localisation.
+
+Under `/data`
```
kevindew, line 156: Nitpicking further but this seems less conventional than using `lib/data` since you'll be adding a new top level directory, whereas lib is already intended to be expanded

kevindew (commented):
This is great @KludgeKML 

Sorry that my last comments seem to have been on about the most minor of minor things in this - I really did want to just let it go but felt a bit guilty that I'm worried my last comments have led to you actually proposing something slightly worse than your initial proposal. Might be best to just take the data stuff out since it's so tangential to the actual core of this proposal?

KludgeKML:
> This is great @KludgeKML 
> 
> Sorry that my last comments seem to have been on about the most minor of minor things in this - I really did want to just let it go but felt a bit guilty that I'm worried my last comments have led to you actually proposing something slightly worse than your initial proposal. Might be best to just take the data stuff out since it's so tangential to the actual core of this proposal?

No need to apologise! It's better to bring up small points when we're being prescriptive.

I'm off until Tuesday now, but perhaps @leenagupte can do the honours? (Perhaps it's worth leaving in that we'll try to avoid data files where possible without defining exactly _where_ we'll avoid them 😉)

rfc-175-frontend-fewer-apps.md
kevindew: Yeah sounds good about it being a moot point - there's something smelly about a directory named data, reminds me of other dubious ones like "shared" and "util".

I think the main thing is to avoid there being two `data` directories - one under `/config/data` and another one as this introduces further ambiguity. 

I do agree that there is not a formally defined location in Rails, so does veer into the realm of individual interpretations of the enigmatic "Rails way". Would also agree that `app/data` is the most surprising suggestion so far. I wouldn't share your concern that it's weird something in `app` referencing a resource from `lib` but I know that veers into the other great Rails debate about the relationship of those directories.

leenagupte (approved):
It looks like all of the concerns raised in the comments have been addressed.

The only minor issue left is about where to put data files. I think that there was agreement that these types of files shouldn't exist if possible, and if they do, they should all be located in the same place. 

I think it's ok for this RFC to be accepted with the decision of where exactly to put data files left for another day, as the existing files are due to be removed.

kevindew (approved):
Yup, I think this is accepted.

Data files seems a very minor tangential issue 👍

PR #177, opened 2024-09-09
RFC-177: Make all content types editionable

brucebolt:
Adds RFC-177, which proposes making all content types editionable.

Deadline for comments: Monday 23rd September 2024.

[Rendered view](https://github.com/alphagov/govuk-rfcs/blob/rfc-177/rfc-177-edition-all-content.md)

[Trello card](https://trello.com/c/YZrRcF9o)

### DIFF
--- a/rfc-177-edition-all-content.md
+++ b/rfc-177-edition-all-content.md
@@ -1,0 +1,161 @@
+---
+status: proposed
+implementation: blocked
+status_last_reviewed: 2024-10-29
+---
+
+# Make all content types editionable
+
+## Summary
+
+GOV.UK has a number of types of content published using Whitehall Publisher. Some of these are editionable (i.e. publishers can create a draft version of the document before publishing) whereas others are not editionable (i.e. changes become live immediately).
+
+This RFC proposes using the edition workflow for all publishable entities in Whitehall Publisher.
+
+## Problem
+
+Documents published outside the edition workflow suffer from the following user-facing issues:
+
+* no draft preview (whether for the publisher themselves, or using a shareable preview link for fact checking)
+* no [2i functionality](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/reviewing-and-approving-content\#a-2i-checklist)
+* no scheduled publishing
+* high risk of content errors as changes go live immediately
+* inconsistent access control
+* no audit trail
+* validations are not consistent with other document types, e.g. the length of a document summary
+
+Additionally, there are a number of developer facing issues:
+
+* a high number of callbacks are manually implemented on models to republish associated non-editionable content when changes are made in other types of non-editionable content, which has previously led to incidents where callbacks are accidentally missed or called too frequently
+* two different sets of code exist for republishing the editionable and non-editionable content, with much code for the non-editionable content being specific to the content type; this affects the developer experience, making it more difficult to make changes, due to hardcoded special cases, a lack of flexibility and developers needing to understand additional code
+
+The edition workflow offers the following features, that are not available on non-editionable models:
+
+* 2i is mandatory; unless the user has permissions to force publish
+* all documents can be scheduled to publish in advance
+* all documents have a shareable preview link, which can be used for fact checking by third parties without a Signon account
+* there is reduced risk of errors being published on GOV.UK by giving publishers a draft preview
+* an audit trail is kept of all edits made (including the exact edit, when and by whom)
+* the workflow is uniform with other document types, so publishers will already be familiar with the process without needing to learn a new user interface specifically for the content type
+* documents can be searched and filtered, e.g. by world location
+* all documents can be found in a central location in Whitehall, negating the need for the ‘More’ option on the heading menu
+
+We have seen publishers work around some of these issues by using the integration environment as a draft preview. However this negates the usefulness of integration as a developer testing ground if we know publishers are using it for real publishing purposes. Data in the integration environment is overwritten each night, so drafting and previewing content over multiple days is not possible. Additionally, there is still a risk of making an error when copying the document content from integration to production.
+
+There have been times where a topical event page has needed to be updated or published overnight, to coincide with an announcement or embargo. Without scheduled publishing functionality, a publisher needs to remain awake overnight to manually publish the document or delay publishing until the morning.
+
+## Previous related work
+
+In late 2023 and early 2024, the GOV.UK Publishing Platform team migrated Worldwide Organisations to use the edition workflow. This involved creating a model that inherited from the Edition class, re-associating all the related objects to this new model and then migrating the data from the non-editionable model to the new one.
+
+As part of this work, the team ensured that the translations on sub-content were consistent with the parent page, since Worldwide Organisations had no validation on the translations that were permitted.
+
+The diagram below shows the relationship between models in Whitehall before Worldwide Organisations were made editionable:
+
+```mermaid
+---
+title: "Architecture of worldwide organisation model's relationship to other models before being made editionable in Whitehall"
+---
+
+erDiagram
+    WorldwideOrganisation {}
+    WorldwideOrganisationWorldLocation {}
+    SocialMediaAccount {}
+    Sponsorship {}
+    WorldwideOffice {}
+    WorldwideOrganisationRole {}
+    EditionWorldwideOrganisation {}
+    FeaturedImageData {}
+    WorldLocation {}
+    Organisation {}
+    Role {}
+    WorldNewsStory["WorldNewsStory (Edition)"] {}
+    CorporateInformationPage["CorporateInformationPage (Edition)"] {}
+
+    WorldwideOrganisation ||--|{ WorldwideOrganisationWorldLocation : "has many"
+    WorldwideOrganisation ||--|{ SocialMediaAccount : "has many"
+    WorldwideOrganisation ||--|{ Sponsorship : "has many"
+    WorldwideOrganisation ||--|{ WorldwideOffice : "has many"
+    WorldwideOrganisation ||--|{ WorldwideOrganisationRole : "has many"
+    WorldwideOrganisation ||--|{ EditionWorldwideOrganisation : "has many"
+    WorldwideOrganisation ||--|{ FeaturedImageData : "has many"
+
+    WorldwideOrganisationWorldLocation ||--|| WorldLocation : "has one"
+    Sponsorship ||--|| Organisation : "has one"
+    WorldwideOrganisationRole ||--|| Role : "has one"
+    EditionWorldwideOrganisation ||--|| WorldNewsStory : "has one"
+    EditionWorldwideOrganisation ||--|{ CorporateInformationPage : "has many"
+```
+
+The diagram below shows the relationship between models in Whitehall after Worldwide Organisations were made editionable:
+
+```mermaid
+---
+title: "Architecture of worldwide organisation model's relationship to other models after being made editionable in Whitehall"
+---
+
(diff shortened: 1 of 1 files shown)


rfc-177-edition-all-content.md
```diff
@@ -0,0 +1,147 @@
...
+The edition workflow offers the following features, that are not available on non-editionable models:
+
+* 2i is mandatory; unless the user has permissions to force publish
+* all documents can be scheduled to publish in advance
+* all documents have a shareable preview link, which can be used for fact checking by third parties without a Signon account
+* there is reduced risk of errors being published on website by giving publishers a draft preview
+* an audit trail is kept of all edits made (including the exact edit, when and by whom)
+* the workflow is uniform with other document types, so publishers will already be familiar with the process without needing to learn a new user interface specifically for the content type
+* documents can be searched and filtered, e.g. by world location
+* all documents can be found in a central location in Whitehall, negating the need for the ‘More’ option on the heading menu
```
yndajas, line 40: Although there are links to other things on the "More" page - cabinet ministers ordering, emergency banner settings, sitewide settings, and republishing content - but maybe we can come up with a more descriptive name for those!

rfc-177-edition-all-content.md
```diff
@@ -0,0 +1,148 @@
...
+
+    EditionWorldLocation ||--|| WorldLocation : "has one"
+    EditionOrganisation ||--|| Organisation : "has one"
+    EditionRole ||--|| Role : "has one"
+    EditionEditionableWorldwideOrganisation ||--|| WorldNewsStory : "has one"
+```
+
+## Proposal
+
+This proposal is that we shall apply the edition workflow to all the remaining non-editionable documents in Whitehall Publisher (and all their associated documents):
```
ChrisBAshton, line 133: Can I clarify who you mean by "we"? Is it the Publishing Platform team?

ChrisBAshton (commented):
Any attempt to make Whitehall behaviour more consistent is welcome by me! One important question below:

rfc-177-edition-all-content.md
JonathanHallam: It wouldn't be the Publishing Platform. We were only really in this space because of the work we were doing around Whitehall rendering. The use of "we" here is more of a royal "we" of GOV.UK, so realistically whichever team is responsible for Whitehall (sorry)!

As much as anything we wanted to write this to advertise that there's a bunch of problems that we've solved and that other content types should be way more straightforward than worldwide orgs were.

kevindew:
Sounds great for all the content in Whitehall to be editionable, it’s a major bit of product debt that’d be great to resolve.

It does however a little peculiar for a team to open an RFC without the intention to do the work, I think we should page @richardTowers to get his view on whether this RFC should continue.

The implementation proposal seems reasonable as it follows the existing approach, however I am concerned about the use of the `editions` table for trying to model all those distinct data structures -  particularly given the thing is already quite the monster. I would suggest that we invest in a couple of things as part of further migrations:

- setting up a JSON field that can store content type specific data so that we can stem the proliferation of columns and join tables (either on Edition or by creating a new model that could be a successor to Editions)
- exploring whether we can either remove the single table inheritance aspect of editions or do more with namespacing for it, so that it’s easier to understand in Whitehall what is a foundational model and what is actually a content type - as the model directory of Whitehall is rather daunting.

ryanb-gds:
Irrespective of whether or not we are committing to doing this work (I'd love the Whitehall team to take a shot at something along these lines but regrettably it's a way down our roadmap), I'd like to share a few thoughts.

I agree that all content on Whitehall should have a publishing workflow and share other aspects of edition-like behaviour (maybe the other active editors warning, audit trails, maaaybe translations). However, I would echo Kevin's concerns about coupling everything to the editions database table (and indeed its Active Record model via inheritance).

My suggestion would be that we explore a more composition based approach so that we can decouple the content types from each other in order to facilitate easier changes to individual types in accordance with user needs. For example, not all content types should be eligible for history mode, so it would be beneficial if we could only apply that behaviour where it is appropriate, rather than applying it to all content types and then having to override it where it is not wanted. Same applies for attachments, opening and closing dates on consultations, search indexing and numerous other features of Whitehall.

I think we can do this by making better use of decoupled concerns with tightly defined interfaces. We could achieve this by making the `document_id` column on the editions table into a polymorphic relationship so that editions could relate to models other than documents. If we made that change, organisations, roles, topical events etc. would sit at the same level as the document object in the class hierarchy, but still have access to the edition behaviour provided by a `has_many` polymorphic association with the edition model. We would also need to move most of the document content off of the edition model to other tables that relate to the document object, ideally using [Active Record’s delegated types](https://guides.rubyonrails.org/association_basics.html#delegated-types). That would keep the edition focused exclusively on managing the publishing workflow state, and let documents be responsible for the content state.

This is a similar problem to that faced by the content modelling team when they started to work on reusable content blocks. I wrote a Whitehall draft RFC about how to support reusable content blocks, which proposed [this database schema](https://github.com/alphagov/whitehall/blob/91f84363f6dae9f271410eebef1803a3cfe9fe51/docs/adr/0003-introduce-reusable-content-items/editions.mmd). A similar shape of schema could also support our current set of non-editionable Whitehall content types. The up front work to achieve this implementation would be quite substantial, but I believe it would significantly reduce the total delivery time required to make all Whitehall content types share an editionable publishing workflow.

leimdorfer:
> Irrespective of whether or not we are committing to doing this work (I'd love the Whitehall team to take a shot at something along these lines but regrettably it's a way down our roadmap)

Without understanding the detail, I definitely love this, both for the overall consistency and for the features it makes available to users. On the "way down our roadmap" thing. Having just joined the team, I'm not even really sure we have a longer term roadmap where this would live, but very keen to start pulling that together and it's good to have this as a reference to point to.

brucebolt:
>It does however a little peculiar for a team to open an RFC without the intention to do the work

This is true, however we wanted to make sure the intent was logged and that the previous work done (especially the knowledge acquired when editioning worldwide organisations) was not forgotten.

> The implementation proposal seems reasonable as it follows the existing approach, however I am concerned about the use of the `editions` table for trying to model all those distinct data structures - particularly given the thing is already quite the monster. I would suggest that we invest in a couple of things as part of further migrations:
> 
> * setting up a JSON field that can store content type specific data so that we can stem the proliferation of columns and join tables (either on Edition or by creating a new model that could be a successor to Editions)
> * exploring whether we can either remove the single table inheritance aspect of editions or do more with namespacing for it, so that it’s easier to understand in Whitehall what is a foundational model and what is actually a content type - as the model directory of Whitehall is rather daunting.

Indeed, this smell is a significant piece of tech debt in Whitehall.  It feels out of scope of this proposal, as that would require significant reworking in Whitehall.  Maybe it could be recorded as a separate piece of tech debt, as it can be resolved independently of any further content types being editioned.

>This is a similar problem to that faced by the content modelling team when they started to work on reusable content blocks. I wrote a Whitehall draft RFC (never published owing to a subsequent change of direction) about how to support reusable content blocks, which proposed this database schema. A similar shape of schema could also support our current set of non-editionable Whitehall content types. The up front work to achieve this implementation would be quite substantial, but I believe it would significantly reduce the total delivery time required to make all Whitehall content types share an editionable publishing workflow.

Maybe we could reword this proposal to suggest the implementer decides on the technical route they take?

rfc-177-edition-all-content.md
```diff
@@ -0,0 +1,148 @@
+---
```
richardTowers, line 1: General thoughts at the top, so that they can be threaded:

- Sorry it's taken me such a long time to get a chance to look at this 🙇🏻 
- I agree with the thrust of the other reviewers, which I'd summarise as:
    - Having an approved RFC doesn't necessarily mean we're going to prioritise the work to implement it, and the decision on if / when to prioritise the work is best left with the Whitehall team
    - There's some tech debt in Whitehall which it would be nice to clear up before continuing to build things on the foundation provided by the Edition model and its Single Table Inheritance approach
    - We have various ideas about how to clear up that tech debt, but we haven't got a decision on a way forward
    - Other than those concerns, everyone generally agrees that making everything Editionable is a good idea

So generally this is a good idea, but we just have practicalities to work through. That's left us a little unsure of whether to approve and merge the RFC, or whether to postpone it until we're ready to embark on the work.

I guess we could have it as

```
status: approved
implementation: blocked
```

With some note in the RFC explaining why it's blocked?

I wonder if all the non-editionable things are equally blocked - I could imagine that something like TakePartPages might be quite similar to Edition (e.g. it has a body), and therefore not too smelly to extend Edition and follow that pattern. Whereas some of the things that are knitted together like People / Roles / RoleAppointments / Organisations might be a worse fit for the current approach to Editions.

rfc-177-edition-all-content.md
```diff
@@ -0,0 +1,148 @@
...
+## Problem
+
+Documents published outside the edition workflow suffer from the following user-facing issues:
+
+* no draft preview (whether for the publisher themselves, or using a shareable preview link for fact checking)
+* no [2i functionality](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/reviewing-and-approving-content\#a-2i-checklist)
+* no scheduled publishing
+* high risk of content errors as changes go live immediately
+* no audit trail
+* validations are not consistent with other document types, e.g. the length of a document summary
```
richardTowers, line 24: Is there also a problem with inconsistent access control? For example non-editors being able to change organisation social media links?

rfc-177-edition-all-content.md
```diff
@@ -0,0 +1,148 @@
...
+* People
+* Role Appointments
+* Roles
+* Take part pages
+* Topical Events
+* World Location News
+
+This would be done in the same way as Worldwide Organisations: a new editionable model (and associated models) would be introduced alongside the existing non-editionable model, but its usage kept behind a feature flag. We would then create a migration to move the data from one model to the other, whilst also flipping the feature flag at the same time.
+
+Furthermore we propose that new document types must be editionable. If a new document type needs to be added to any publishing application, it must always follow an edition workflow.
```
richardTowers, line 148: So I'd suggest adding a `Blocked` section to the bottom here, to cover Kevin and Ryan's concerns that the code isn't factored appropriately to provide a good foundation for making more things editionable.

That said, some things (like "new document types must be editionable") do seem like actual decisions we could make. I guess we could split out two RFCs (1 - new document types must be editionable, 2 - make non-editionable document types editionable), but that's a bit of a faff... Maybe "partially blocked"?

rfc-177-edition-all-content.md
brucebolt: Yes, there are inconsistencies.  Especially the hardcoded VIP editor thing we added a few years ago.

rfc-177-edition-all-content.md
brucebolt: I'll stick with this being "partially blocked", but not dictate that future document types must be editionable from now.  This is because a new complex document type may hit the same tech debt problems.  I'll suggest that the new document should be editionable from the point that the underlying tech debt is fixed.

rfc-177-edition-all-content.md
brucebolt: That seems like a good idea to me.  I've updated the PR to reflect the blocked implementation status and added some reasons why.  [Link](https://github.com/alphagov/govuk-rfcs/compare/d1882e6c6b0e3b34ac4b1f1e2aed5308dfa13c2b..22aa2871e099b03f5ea56b8e6333b4bb0be13abe) to the diff.

richardTowers (approved):

PR #181, opened 2025-03-14
Asset update via publishing api rfc

ryanb-gds:
[Rendered version (including diagrams)](https://github.com/alphagov/govuk-rfcs/blob/asset-update-via-publishing-api-rfc/rfc-178-update-asset-lifecycle-via-publishing-api.md) 💅 

Looking for early feedback from Whitehall team for now (as we may not be able to commit to implementing this). Will kick off a formal review process if agreed that the Whitehall team is in favour and keen to undertake the work.

### DIFF
--- a/rfc-178-update-asset-lifecycle-via-publishing-api.md
+++ b/rfc-178-update-asset-lifecycle-via-publishing-api.md
@@ -1,0 +1,129 @@
+---
+status: proposed
+implementation: proposed
+status_last_reviewed:
+---
+
+# Update Asset Lifecycle State via Publishing API Message Queue
+
+## Summary
+
+Currently, each publishing application updates the state of each asset associated with an edition of a document when the edition's state changes. For example, when an edition is published, Whitehall updates the `draft` state of any assets associated with the edition to `false` by making an API call to Asset Manager. This will cause Asset Manager to serve the asset from the live domain. Other publishing applications which support asset uploads, such as Specialist Publisher and Manuals Publisher, make the same API call each time a document is published.
+
+To reduce this duplication and reduce the chatty API connections between the publishing applications and Asset Manager, we could have Asset Manager consume messages placed on RabbitMQ by Publishing API whenever an edition's state is updated. The message processor could then inspect the state of the edition's attachments and images and apply the appropriate updates to the relevant asset states, for example by updating the redirect URL associated with the assets if the edition has been unpublished.
+
+## Problem
+
+GOV.UK publishing has found asset management a difficult challenge in recent history. It has been the source of several incidents and bugs, such as [this high profile incident](https://docs.google.com/document/d/1niSvK0w-BkzHpkqj6-yqcHrRPI4OfYhOfHn5paEQH7Q/edit?usp=sharing) from 2024, and [this extant bug in Manuals Publisher](https://trello.com/c/5LqwCypV). There are two main problems:
+
+1. Each publishing app which supports asset uploads has duplicate code for updating the state of Asset Manager
+2. The code for synchronising the state of Asset Manager with the state of Publishing API in the publishing applications is generally very brittle and sensitive to race conditions.
+
+## Proposal
+
+There SHOULD be a single mechanism for updating the state of an asset based on the owning edition's state. This SHOULD be achieved via Publishing API, which will reduce the burden on the publishing applications to maintain consistency between Publishing API and Asset Manager. This will make the data flow consistent with how Search API and Router API are updated, and decouple the publishing applications from Asset Manager.
+
+In this scenario, publishing applications MUST upload binary assets directly to Asset Manager (as per the current process) They MUST send any access bypass information and access limiting information (also as per current process). Asset Manager can set the default lifecycle state for new assets as follows:
+
+```ruby
+{
+  draft: true,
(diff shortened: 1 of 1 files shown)


rfc-178-update-asset-lifecycle-via-publishing-api.md
```diff
@@ -0,0 +1,184 @@
...
+## Problem
+
+GOV.UK publishing has found asset management a difficult challenge in recent history. It has been the source of several incidents and bugs, such as [this high profile incident](https://docs.google.com/document/d/1niSvK0w-BkzHpkqj6-yqcHrRPI4OfYhOfHn5paEQH7Q/edit?usp=sharing) from 2024, and [this extant bug in Manuals Publisher](https://trello.com/c/5LqwCypV). There are two main problems:
+
+1. Each publishing app which supports asset uploads has duplicate code for updating the state of Asset Manager.
+2. The code for synchronising the state of Asset Manager with the state of Publishing API in the publishing applications is generally very brittle and sensitive to race conditions.
+
+## Proposal
+
+There SHOULD be a single mechanism for updating the state of an asset based on the owning edition's state. This SHOULD be achieved via Publishing API, which will reduce the burden on the publishing applications to maintain consistency between Publishing API and Asset Manager. This will make the data flow consistent with how Search API and Router API are updated, and decouple the publishing applications from Asset Manager.
```
ChrisBAshton, line 24: I'd remove reference to Router API, which was retired at the end of 2024. (Could instead refer to Email Alert Service as an additional example of something driven by the publishing API queue)

rfc-178-update-asset-lifecycle-via-publishing-api.md
```diff
@@ -0,0 +1,184 @@
...
+  "alt_text": "The launch of the Amelia Troubridge photography exhibition at Getty Images Gallery, London.",
+  "caption": "The launch of the Amelia Troubridge photography exhibition at Getty Images Gallery, London.",
+  "high_resolution_url": "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/image_data/file/65710/s960_Fanzi-Down.jpg",
+  "url": "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/image_data/file/65710/s300_Fanzi-Down.jpg"
+}
+```
+
+Again, we MAY include the Asset Manager IDs for each image, and we MUST provide deletion and replacement state. Deleted and replaced images MUST be filtered out of the data sent to content store.
+
+Publishing API already places a message on a RabbitMQ exchange each time an edition is published. We MUST add another topic for updates to draft editions so that Asset Manager can make additional updates to access limiting and access bypass states for draft documents. This may also be useful in the future to provide draft stack capability for Search API.
```
ChrisBAshton, line 77: May be worth linking to https://trello.com/c/5nN8vKG7/600-create-draft-search-index-to-enable-draft-search-preview here

rfc-178-update-asset-lifecycle-via-publishing-api.md
```diff
@@ -0,0 +1,184 @@
...
+    "id": "8485786",
+    "isbn": "",
+    "title": "Amendments 5 to 8 to Clause 37: Claim for relief on foreign income",
+    "unique_reference": "",
+    "unnumbered_command_paper": false,
+    "unnumbered_hoc_paper": false,
+    "url": "https://assets.publishing.service.gov.uk/media/67bf19a4750837d7604dbb96/Amendments_5_to_8_to_Clause_37-Claim_for_relief_on_foreign_income.odt"
+}
+```
+
```
ChrisBAshton, line 61: Is it worth writing up a paragraph explaining the 'legacy media paths' and any impact (or lack thereof) on this RFC? ISTR we're no longer generating those paths but that there are still plenty of assets that are referenced via them.

ChrisBAshton (commented):
Love the idea in principle 👏

rfc-178-update-asset-lifecycle-via-publishing-api.md
ryanb-gds: Didn't realise that, excellent

rfc-178-update-asset-lifecycle-via-publishing-api.md
ryanb-gds: I don't think there's any impact here. I was able to confirm in the Whitehall console that there are no assets without an Asset Manager ID and that any attachment datas and image datas without assets are either deleted or replaced. Would be good to know if that matches the expectations of @lauraghiorghisor-tw

lauraghiorghisor-tw: Yep, we "migrated" everything. I think it was driven from WH's AttachmentData's (and other models that have the carrierwave mounted on them). We fetched the asset by corresponding legacy path and populated WH with the asset manager ID. I do not think we removed the legacy field from any of those assets in AM, so they are still discoverable by accessing the path URL AND the ID url (which is why you would have seen it around @ChrisBAshton). 

I believe other apps other than WH fetch their assets by ID too.

There are - if I remember correctly - hanging assets in AM that were not touched by the migration - for whatever reason they did not have corresponding `AttachmentData`s anymore. These were not deleted at the time.

PR #183, opened 2025-05-13
WIP: Technical approach for publishing 'flexible pages'

ChrisBAshton:
See [rendered version](https://github.com/alphagov/govuk-rfcs/blob/publishing-flexible-pages/rfc-183-publishing-flexible-pages.md) 💅 

We will iterate this draft RFC as we work through a series of mini technical spikes, and eventually aim to open it up for review, to go hand in hand with the frontend RFC (RFC-182).

### DIFF
--- a/rfc-183-publishing-flexible-pages.md
+++ b/rfc-183-publishing-flexible-pages.md
@@ -1,0 +1,123 @@
+---
+status: draft (not yet ready for review)
+implementation: draft (not yet ready for review)
+status_last_reviewed:
+---
+
+# Technical approach to publishing Flexible pages
+
+## Summary
+
+Key decisions to make:
+
+1. Should the flexible page content schema provide a layout within which various block components are specified, or a list of sections with the layout encoded in the frontend?
+2. How strict should the Publishing API content schema be in enforcing the content?
+3. Should the publishing interface for flexible pages be built in Whitehall, Specialist Publisher, or another application?
+4. How will we model the flexible page content in our data store?
+
+## Decisions
+
+### 1. Should the flexible page publishing schema provide a layout within which various block components are specified, or a list of sections with the layout encoded in the frontend?
+
+We feel that providing the layout within the publishing schema will enable us to support a wider range of user needs faster than if each layout is encoded in the frontend. By encoding the layout in the schema, we hope to minimise the scope of changes required when new content types are requested to just a single schema file. Encoding the layout in the frontend would mean that changes or new flexible page types would require code changes in both the publishing app and the frontend app.
+
+With the above consideration in mind, here is the schema we propose for the Publishing application:
+
+```yaml
+- document_type: history_page
+  rows:
+    - columns:
+        - width: full
+          chunks:
+            - type: page_title
+              uid: page_title
+              heading_text:
+                type: text
+                required: true
+              context:
+                type: text
+              lead_paragraph:
+                type: text
+    - columns:
+        - width: one-third
+          chunks:
+            - type: image
+              uid: sidebar_image
+              src:
+                type: text
+              alt:
+                type: text
+            - type: contents_list
+              uid: contents_list
+              derived_from: govspeak_body # references uid from chunk from which contents will be extracted
+
+        - width: two-thirds
+          chunks:
+            - type: govspeak
+              uid: govspeak_body
+              body:
+                type: govspeak
+```
+
+We believe this schema allows for clear iteration in code, as follows:
+
+```ruby
+flexible_page.rows.each do |row|
+  row.columns.each do |column|
+    column.chunks.each do |chunk|
+      #render chunk
+    end
+  end
+end
+```
+
+We include `uid` values for each chunk so that chunks can reference each other, as will be required for the contents list chunk.
+
+The width value for each column can be appended to `govuk-grid-column` to generate the appropriate class name for the frontend.
+
+#### Decision
+
+### 2. How strict should the Publishing API content schema be in enforcing the content?
+
+There are two simple approaches to validating the flexible page content sent to Publishing API:
+
+1. No validation of the flexible page content
+2. Write a schema for each flexible page type specifying exactly what attributes should be present within the content
+
+If we were to adopt the second design from decision 1, which places sections within a particular region, then an additional approach is possible:
+
+3. Write a single schema for all flexible page types which validates the individual sections, but allows for any sections to be used in any region.
+
+#### Decision
+
+### 3. Should the publishing interface for flexible pages be built in Whitehall, Specialist Publisher, or another application?
+
+From our existing publishing applications, there are two obvious candidates to host flexible page publishing: Whitehall and Specialist Publisher.
+
+Factors in favour of hosting flexible page publishing in Whitehall:
+
+1. Whitehall has the most mature feature set. Extending the Whitehall edition model would enable flexible pages with the full publishing workflow, including scheduling and 2i.
+2. Whitehall is the most commonly used publishing tool, so publishers will not need to adapt to a new tool.
+
+Factors in favour of hosting flexible page publishing in Specialist Publisher:
+
+1. Specialist Publisher already deals with dynamically generating content forms from a schema, so we have some useful primitives in place.
+2. We're less encumbered by Whitehall's hard-to-change publishing workflow and deep inheritance trees.
+3. We're storing the data directly in Publishing API, so we could rely on Publishing API's schema validation rather that implementing additional validation within the publishing application.
+
+We should also consider whether extending an open source content management system may be appropriate for our needs. However, we would need to understand how to integrate such a system with our Publishing ecosystem tools such as SignOn, Publishing API and Asset Manager.
+
+#### Decision
+
+### 4. How will we model the flexible page content in our data store?
+
+We are expecting to store the flexible page content in a denormalised data structure, because we don't plan to reuse flexible sections across documents. Denormalising the data will reduce the effort for developers to add new flexible page types, at the cost of making changes to the data structures more difficult owing to a lack of support from common database migration tools. If we choose to host flexible page publishing in Whitehall, we MAY use a JSONB column on the `editions` table to store the flexible page content in. We MAY also serialize the data into the existing `body` column.
+
+We will need a plan to handle migrations when the data structure needs to change. There are two options available:
+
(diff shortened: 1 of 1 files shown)


rfc-183-publishing-flexible-pages.md
```diff
@@ -0,0 +1,126 @@
...
+Key decisions to make:
+
+1. Should the flexible page content schema provide a layout within which various block components are specified, or a list of sections with the layout encoded in the frontend?
+2. How strict should the Publishing API content schema be in enforcing the content?
+3. Should the publishing interface for flexible pages be built in Whitehall, Specialist Publisher, or another application?
+4. What will the migration path be when we want to make a change to a flexible page type?
+
+## Decisions
+
+### 1. Should the flexible page content schema provide a layout within which various block components are specified, or a list of sections with the layout encoded in the frontend?
```
ChrisBAshton, line 20: For context, I think we're proposing a small tweak to RFC-182 (which the P&P team have indicated they're open to, though I agree we'll want to get that formally approved before going too far down this path) which makes the frontend even simpler and more procedural, and allows for more flexibility and potentially faster turnaround times for new preset layouts.

Ultimately the below would be captured in RFC-182 rather than here, but we should capture it here for now. Instead of the Content Store having sections defined like this:

```yml
flexible_sections:
- type: page_title
  heading_text: "11 Downing Street"
  context: "History"
  lead_paragraph: "This is the history of 11 Downing Street."
- type: rich_text_with_contents_list
    contents_list:
      image:
        source: (image path)
        alt: ""
        caption: ""
      items:
      - href: “#roots”
        text: “Roots of the Treasury”
      - href: “#rise”
        text: “The rise of Downing street”
    content: (govspeak content)
```

...it would have sections defined something like this:

```yml
flexible_sections:
- flexible_section:
  - container: column-full-width
    elements:
      - type: page_title
        heading_text: "11 Downing Street"
        context: "History"
        lead_paragraph: "This is the history of 11 Downing Street."
- flexible_section:
  - container: column-one-third
    elements:
    - type: contents_list
      image:
        source: (image path)
        alt: ""
        caption: ""
      items:
      - href: “#roots”
        text: “Roots of the Treasury”
      - href: “#rise”
        text: “The rise of Downing street”
    - container: column-two-thirds
      elements:
      - type: govspeak
        content: (govspeak content)
```

Let's work through an example to explore the advantages of the change. Imagine there's a new requirement for a 'rich text with contents list' where the contents list is in a sidebar on the right instead of on the left. Under the original RFC-182 proposal, that would require either:

1. A whole new flexible section definition (e.g. `rich_text_with_contents_list__right_aligned`) - taking time to build and add to the schemas
2. A new property on the existing flexible section definition (e.g. `sidebar_on_the_right: true`) - taking time to build and add to the schemas
3. Some other workaround (adding some CSS grid stuff to Frontend to hardcode the fact that the `history_page` document type should have its sidebar on the right instead of the left) - taking time to build and going against the "flexible" aim of the project

Now let's look at the same example with the 'layout-aware' schema we're proposing instead. A requirement for a right-aligned sidebar would simply mean moving the `container: column-one-third -> type: contents_list` definition below the `container: column-two-thirds -> type: govspeak` definition. I.e. render the two-thirds-column of GovSpeak first, and then the one-third column of contents list. Such a change wouldn't require any changes to Frontend, nor to the Content Store / Publishing API schemas, so we could deliver changes more quickly whilst not adding any extra code 🎉 

The problem, of course, is that P&P are keen to not give publishers free reign to add any arbitrary elements to any arbitrary column layout - it has the potential to get messy very quickly. But there's no need to give *publishers* this level of flexibility in the Whitehall UI - we can still offer a fixed "here's a contents list on the left, GovSpeak on the right" UI for the publishers, but render the more flexible layout-centric Content Store schema at publish time. There's flexibility _under the hood_, rather than in the UI, where we want only preset layouts (for now).

With all that context, we can now start to think about how we might want to describe our preset layouts in Whitehall, as a consumable 'thing' for Whitehall to dynamically render a form with inputs for all the necessary bits. I imagine that would look something like this (taking the example schema from the [Publishing strategy for flexible pages](https://docs.google.com/document/d/1VrmJrAH7SXL8HiBUHilT15FGSKvhBswm2JYqdKWHqYE/edit?pli=1&tab=t.0) but adding the container metadata):

```yml
document_type: history_page
flexible_sections:
- flexible_section:
  -  container: column-full-width
      type: page_title
      uid: title
      inputs:
      - name: heading_text
        type: text
        required: true
      - name: context
        type: text
      - name: lead_paragraph
        type: text

- flexible_section:
  type: rich_text_with_contents_list
  uid: rtcwcl
  inputs:
  - subsection: contents_list
    container: column-one-third
    inputs:
    - name: image
      inputs:
      - name: source
        type: url
        required: true
      - name: alt
        type: text
      - name: caption
        type: text
    - name: items
      children:
      - name: href
        type: text
        required: true
      - name: text
        type: text
        required: true
    - name: content
      container: column-two-thirds
      type: govspeak
      required: true
```

Or, more closely following what will be in Content Store, we could do away with the idea of `subsection`s altogether and define it as:

```yaml
flexible_sections:
- flexible_section:
  - container: column-full-width
    elements:
      - type: page_title
        heading_text:
          type: text
          required: true
        context: text
        lead_paragraph: text
- flexible_section:
  - container: column-one-third
    elements:
    - type: contents_list
      image:
        source: url
        alt: text
        caption: text
      items:
        children:
        - name: href
          type: text
          required: true
    - container: column-two-thirds
      elements:
      - type: govspeak
        required: true
```

rfc-183-publishing-flexible-pages.md
```diff
@@ -0,0 +1,126 @@
...
+status_last_reviewed:
+---
+
+# Technical approach to publishing Flexible pages
+
+## Summary
+
+Key decisions to make:
+
+1. Should the flexible page content schema provide a layout within which various block components are specified, or a list of sections with the layout encoded in the frontend?
```
lauraghiorghisor-tw, line 13: Perhaps a follow up question then is, if we go with the layout-aware option: 
- What are the basic building blocks we can start with that are enough for MVP? I think currently we're mainly talking about title, image, contents list, govspeak.

rfc-183-publishing-flexible-pages.md
```diff
@@ -0,0 +1,126 @@
...
+    uid: body
+    inputs:
+      - name: content
+        type: govspeak
+        required: true
+```
+
+In this design the schema defines a number of page regions which can be populated by one or more sections.
+
+### 2. How strict should the Publishing API content schema be in enforcing the content?
```
lauraghiorghisor-tw, line 117: This can then be either:
```yaml
document_type: history_page
schema_name: history_page
// presence of certain components can be required
```
-> in which case when there is a request for new document types, a step in the dev wokflow will have to be adding a new schema to Publishing Api.

or
```yaml
document_type: history_page
schema_name: flexible_page
//nothing can be required, everything is allowed, as long as it is a recognised "building block"/component/item/element or whatever name we land on
```
-> in which case we only have to register a new document type but we can reuse the schema. This separation means we maintain open the future possibility of amending this document type, making things required if we want or reorganising the layout. This allows us to cater to different document type needs.

or
```yaml
document_type: flexible_thing
schema_name: flexible_page
//nothing can be required, everything is allowed, as long as it is a recognised "building block"/component/item/element or whatever name we land on
```
-> we cannot treat a history page any differently from any other page. There is basically no history page.

Does the URL we want to serve these at enforce how we understand the "document_type"? Meaning, if we want to serve all history pages from `/history_pages/...` then they need to be grouped as such?

rfc-183-publishing-flexible-pages.md
lauraghiorghisor-tw: - Ideas on granularity - 🥘 💭 
I was really sold (in my head) to the idea of a more granular Design-System-like set of components that we could use as "building blocks" for the grid layout - with the benefit that you'd have consistent sizing/granularity across the blocks. Govspeak is the outlier/blocker here. 

The implementation suggested above, with govspeak as a block, gets a bit messy when you're trying to reconcile, for example, the concept of an image that is a stand-alone component than can span the entire width of a page, with that of an image that can be embedded in govspeak. Especially when we have to also consider the complications of how the Image Tab is behaving in whitehall currently.

Let's say we do the POC in Whitehall, with govspeak as a building block. I'd argue it's still worth considering, now, or as a futurespective exercise - if you were to build towards ✨ a very flexible and very interactive flexible pages UI 🤖 - **how would we break down the govspeak into smaller large-enough components**? Perhaps the basic unit is  text only (but with all the callouts and whatnot supported). Anything media-like would be added as a stand-alone component.

rfc-183-publishing-flexible-pages.md
ryanb-gds: That's a good point about the URLs. I would imagine your option 2 above will be the best bet, which I believe is how specialist documents work.

rfc-183-publishing-flexible-pages.md
```diff
@@ -0,0 +1,162 @@
...
+#### Decision
+
+### 3. Should the publishing interface for flexible pages be built in Whitehall, Specialist Publisher, or another application?
+
+From our existing publishing applications, there are two obvious candidates to host flexible page publishing: Whitehall and Specialist Publisher.
+
+Factors in favour of hosting flexible page publishing in Whitehall:
+
+1. Whitehall has the most mature feature set. Extending the Whitehall edition model would enable flexible pages with the full publishing workflow, including scheduling and 2i.
+2. Whitehall is the most commonly used publishing tool, so publishers will not need to adapt to a new tool.
```
richardTowers, line 139: I wonder if there's an additional benefit to Whitehall, in that it has fairly robust support for assets (e.g. uploading images, cropping them etc.) which AFAIK specialist doesn't have.

I imagine assets is going to be a big part of this whole shindig.

rfc-183-publishing-flexible-pages.md
ryanb-gds: Have largely adopted @lauraghiorghisor-tw approach above, with minor tweaks to avoid the extra `row` key which isn't strictly needed

rfc-183-publishing-flexible-pages.md
ryanb-gds: That's correct, as required for history pages

PR #186, opened 2025-10-03
RFC for Replacement of Email-based Fact Checking process

baisa:
Proposal to replace Publisher’s current email-based fact check workflow with a secure, auditable system that supports multiple reviewers and reduces manual effort. The supporting email service is being deprecated, and its unreliability has already created ongoing toil for the Mainstream team (e.g. issues in email processing).

Closing date for comments: 17th October 2025

### DIFF
--- a/rfc-186-replacement-of-email-based-fact-checking-process-in-publisher.md
+++ b/rfc-186-replacement-of-email-based-fact-checking-process-in-publisher.md
@@ -1,0 +1,66 @@
+---
+status: proposed
+implementation: proposed
+status_last_reviewed:
+---
+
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Agree with Martyn - seems like a good approach to start with a rails engine and separate it out later if it turns out that's a requirement.

I'm assuming that's the main decision this RFC is inviting comment on, but there are obviously lots of subsequent decisions that will need to be made (how users get access to the tool, etc.). I'm happy to take it one decision at a time though, so 👍🏻  on the rails engine approach.

PR #187, opened 2026-01-14
[PNP-9741] Update RFC-176 to done

KludgeKML:
Static is now retired, so update this RFC status to done.

[JIRA Card](https://gov-uk.atlassian.net/browse/PNP-9741)


deborahchua (approved):
🎉

PR #188, opened 2026-02-12
RFC-188: Make Whitehall the Authority of its own Taxons [WHIT-3066]

ChrisBAshton:
This RFC has been **closed** as **not accepted** - see [summary](https://github.com/alphagov/govuk-rfcs/pull/188#issuecomment-4021963250). ⛔ 

---

[Rendered view](https://github.com/alphagov/govuk-rfcs/blob/whitehall-taxons-refactor/rfc-188-make-whitehall-the-authority-on-its-taxons.md) 💅 

## Summary

This RFC proposes that Whitehall should be the canonical data store for which pieces of content are tagged to which topic and world taxonomies. It will be responsible for pushing the taxons to Publishing API, in the same way that it is already responsible for pushing other kinds of links, such as organisations and ministers.

Benefits:

- Enables different topic/world taxonomies on draft vs live content.
- Eradicates the root cause of a race condition whereby chosen taxonomies can take a while to 'stick' in Publishing API.
- Simplifies Whitehall and makes it more internally consistent.
- Removes runtime coupling between Whitehall and Publishing API for tagging.

### Recent context

1. We're currently working on a rewrite of Topical Events. It was mooted by the Content stakeholders that that format should not have the option of tagging to a topic taxonomy. Our other kinds of associations can be easily configured to be enabled or disabled via the StandardEdition 'config driven' architecture, but because taxons are implemented so differently, it's not clear that we'd be able to extend the same approach there.

2. Having an early look at multi-page architecture in Whitehall's new config-driven approach, we're likely to want to disable topic taxonomy choice from 'child documents' which should otherwise inherit from their parents. As per 1, the most idiomatic way to implement this would be to make taxonomy tagging a configurable property in the content type schema, which would be made easier if taxons worked in the same way as other associations.

Jira: https://gov-uk.atlassian.net/browse/WHIT-3066

### DIFF
--- a/rfc-188-make-whitehall-the-authority-on-its-taxons.md
+++ b/rfc-188-make-whitehall-the-authority-on-its-taxons.md
@@ -21,79 +21,82 @@ Whitehall will:
 
 - Persist selected taxons locally, scoped to Editions.
 - Render taxon tagging from local persistence (removing eventual-consistency delays).
-- Push a complete, authoritative set of taxon links to Publishing API:
-  - as edition_links on draft save
-  - as edition_links and link set updates on publish/republish
+- Push a complete, authoritative set of taxon links to Publishing API, as edition_links
 
 The list of available taxons (the taxonomy catalogue) will continue to be fetched and cached from the publishing stack.
 
-## Problem
+It is also worth noting that this proposal makes Whitehall the canonical source of Topic/World Taxonomy associations for Whitehall content _only_. The RFC does not propose changing the canonical source of these associations for any non-Whitehall content, which remains the responsibility of Content Tagger.
 
-### User problem
+### Benefits
 
-After tagging an Edition with taxons in Whitehall, subsequent refreshes often do not reflect the change immediately, especially in the local dev environment.
-
-This is caused by Whitehall reading taxon tagging from Publishing API at runtime while the update is processed asynchronously or cached (TBC: which?).
-
-This results in:
-
-- confusing UX ("did it save?")
-- reduced trust in the system
-- runtime coupling between Whitehall and Publishing API for editorial workflows
-- friction and delays in local development
-
-### Technical problem
-
-Taxon tagging for Whitehall-managed content is currently treated as externally-owned state:
-
-- Whitehall must query Publishing API at runtime to display tagging.
-- Draft and live semantics are harder to reason about.
-- Topic and world taxonomies must both be considered when reading/writing.
-- Data flow differs from other Whitehall associations.
+- Eradicates the root cause of a race condition.
+- Simplifies Whitehall and makes it more internally consistent.
+- Enables different topic/world taxonomies on draft vs live content.
+- Removes runtime coupling between Whitehall and Publishing API for tagging.
 
 ## Context
 
-- Taxonomies (topic + world) are maintained outside Whitehall (in Content Tagger).
-- Taxons are stored in Publishing API as content items.
-- Publishing API stores links by `content_id`.
-TODO: double-check the following definitions:
-- Link sets are document-scoped (not edition-scoped).
-- Edition links are draft-scoped and allow draft/live divergence.
+- Taxonomies (both topic + world) are:
+  - Maintained outside Whitehall (in Content Tagger).
+  - Stored in Publishing API as content items (see examples below).
+  - Presented as a combined `taxons` link on content items ([example](https://www.gov.uk/api/content/government/publications/latvia-list-of-medical-facilities)).
+- Publishing API [link lifecycles](https://docs.publishing.service.gov.uk/repos/publishing-api/link-expansion.html#link-lifecycles):
+  - Link sets are document-scoped (not edition-scoped).
+  - Edition links are draft-scoped and allow draft/live divergence.
+- Whitehall prefers draftable associations (e.g. organisations via edition_links) so that changes made on a draft edition do not affect the live edition until publish.
 
-Whitehall prefers draftable associations (e.g. organisations via edition_links) so that changes made on a draft edition do not affect the live edition until publish.
+Taxons are not special in Publishing API — they are stored as a standard `"taxons"` link type. There are no hardcoded rules requiring Content Tagger to be the canonical writer.
 
-This RFC aligns taxon tagging with that model.
+```rb
+# Publishing API
 
-## Investigation: Publishing API behaviour
+topic_taxon = Document.find_by(content_id: "71f685fe-0dc3-469d-92c2-1000ad86d7d9").editions.last
+=> 
+#<Edition:0x0000ffff68066130
+ id: 3359281,
+ title: "River maintenance, flooding and coastal erosion",
+ publishing_app: "content-tagger",
+ rendering_app: "collections",
+ update_type: "major",
+ document_type: "taxon",
+ schema_name: "taxon",
+ first_published_at: "2018-03-08 16:32:37.000000000 +0000",
+ last_edited_at: "2018-09-16 10:36:13.350772000 +0000",
+ state: "published",
+ content_store: "live",
+ routes: [{"path" => "/environment/river-maintenance-flooding-coastal-erosion", "type" => "exact"}]>
 
-### Key findings
+world_taxon = Document.find_by(content_id: "7283f126-c578-4df7-949e-44662d4c086b").editions.last
+=> 
+#<Edition:0x0000ffff5e47a780
+ id: 4682028,
+ title: "Living in Latvia",
+ publishing_app: "content-tagger",
+ rendering_app: "collections",
+ update_type: "major",
+ document_type: "taxon",
+ schema_name: "taxon",
+ first_published_at: "2017-06-29 14:02:28.826713000 +0000",
+ last_edited_at: "2020-03-03 16:56:27.160273000 +0000",
+ state: "published",
+ content_store: "live",
+ routes: [{"path" => "/world/living-in-latvia", "type" => "exact"}]>
+```
 
-TODO: double-check and verify the below.
+## Problem
 
-1. Taxons are not special in Publishing API.
-   - They are just another link type: `"taxons"`.
-   - There are no hardcoded business rules restricting which publishing app may set taxons.
+Taxon tagging for Whitehall-managed content is currently treated as externally-owned state. Whitehall queries Publishing API at runtime to display tagging.
 
-2. Publishing API does not distinguish topic vs world taxonomy at storage level.
-   - Both are stored under `"taxons"`.
+Consequences:
 
-3. Link sets are tied to `content_id` and are not edition-scoped.
-   - Writing to link sets affects the document-level associations.
-
-4. Edition links are draft-scoped and allow draft/live divergence.
-
-5. Publishing API does not automatically merge partial link updates.
-   - PATCHing links requires sending the full desired set.
-   - Omitting links risks overwriting them.
-
-6. Content Tagger writes taxons by mutating Publishing API link sets.
-   - There is no enforcement that only Content Tagger may do so.
-
-Conclusion:
-
-There are no hidden business rules in Publishing API that prevent Whitehall from becoming authoritative for taxon tagging.
-
-The main risk is partial overwriting of link sets, which can be mitigated by always sending the full combined set.
+- Inconsistent with how Whitehall models other associations.
+- Makes consolidation of a config-driven association architecture more difficult.
+- Introduces runtime coupling to Publishing API.
+- Code is more complex to reason about.
+- Race conditions have been observed whereby saving new taxon selections takes time to “stick”, showing stale data on refresh.
+  - This hinders local development (topic-less documents are invalid and cannot be force-published).
+  - The issue has also been observed in integration.
+- Taxon tagging is currently implemented as link set links, meaning draft changes immediately affect live associations.
 
 ## Proposal
 
@@ -101,16 +104,16 @@ The main risk is partial overwriting of link sets, which can be mitigated by alw
 
 Whitehall will store selected taxons per Edition.
 
-This enables draft editions to differ from live editions.
-
-Proposed model will be something like:
+Proposed model:
 
 - `edition_taxon_links`
   - `edition_id`
   - `taxon_content_id`
   - `taxonomy_kind` (enum: `topic`, `world`)
 
-When a new edition is created, it will inherit taxons from the previous edition.
+When a new edition is created, it will duplicate taxon links from the previous edition.
+
+This enables draft editions to differ from live editions.
 
 ### 2. Render from local persistence
 
@@ -118,86 +121,46 @@ Whitehall UI will render taxon tagging from `edition_taxon_links`.
 
 Whitehall will no longer call Publishing API at runtime to determine which taxons an Edition is tagged to.
 
-This removes:
-
-- eventual-consistency issues
-- runtime dependency on Publishing API
-- cross-app state ambiguity
-
-### 3. Draft vs Live semantics
-
-Whitehall prefers draftable associations.
-
-Therefore:
-
-#### On save draft
-
-- Taxons are written as edition_links only.
-- Link sets are NOT updated.
-- Live associations remain unchanged.
-
-TODO: is there a case for dropping link set links altogether?
-
-#### On publish / republish
-
-- Taxons are written as edition_links.
-- The document’s link set is updated via `PATCH /v2/links/:content_id`.
-- The full combined set of topic + world taxons is sent.
-
-This ensures:
-
-- Draft changes do not affect live content prematurely.
-- Link sets remain authoritative for live content.
-- Downstream consumers receive correct associations.
-
-### 4. Always send full combined taxon set
-
-When updating link sets:
-
-- Whitehall must always send the complete combined set of topic + world taxons.
-- Partial updates are not permitted.
-
-This avoids accidental overwriting of one taxonomy with another.
-
-### 5. Continue fetching taxonomy catalogue
-
 Whitehall will continue to fetch and cache the list of available taxons from the publishing stack.
 
-This proposal changes tagging ownership, not taxonomy ownership.
+This proposal changes tagging ownership, not taxonomy ownership. The question of taxonomy ownership and governance is being investigated separately by the `#govuk-publishing-tagging-workflow` team.
 
-## Migration Plan
+### 3. Update Publishing API integration
 
-### 1. Backfill
+Taxon links will no longer be managed via `patch_links`.
+
+Instead, taxons will be sent via `put_content` as edition links, like Whitehall's other associations.
+
+Whitehall must always send the complete combined set of topic + world taxons. Partial updates are not permitted.
+
+At this stage we have fully editionable taxonomies — no further use of link sets for taxons.
+
+## Implementation Plan
+
+### Phase 1 – Introduce local persistence
+
+- Create `edition_taxon_links`.
+- Continue writing to Publishing API as currently implemented. In addition, write the taxon selections locally.
+- Duplicate taxon links when new editions are created.
+- Continue reading taxon selections from Publishing API. Do not yet switch rendering.
+
+### Phase 2 – Backfill existing data
 
 For each Whitehall document:
 
 - Fetch existing taxon link sets from Publishing API.
-- Populate `edition_taxon_links` for the latest edition.
+- Populate/overwrite `edition_taxon_links` for both the live and draft editions (if applicable).
+- Ensure both topic and world taxons are captured.
 
-Both topic and world taxons must be included.
+### Phase 3 – Switch read path
 
-### 2. Dual-read period (temporary)
+- Render taxon tagging from `edition_taxon_links`.
+- Remove runtime reads from Publishing API.
 
-For a limited period:
+### Phase 4 – Remove link set updates
 
-- Compare local taxon set vs Publishing API link set.
-- Log diffs.
-- Identify edge cases.
-
-### 3. Switch UI reads
-
-Switch Whitehall to render from local persistence only.
-
-### 4. Switch publish writes
-
-Update publish/republish flow to:
-
-- Write edition_links
-- PATCH link set with full combined taxon set
-
-### 5. Remove runtime dependency
-
-Remove all runtime reads of taxon tagging from Publishing API.
+- Remove `patch_links` usage for taxons.
+- At this stage, taxonomies are updated solely on edition_links via `put_content`.
 
 ## Governance Considerations
 
@@ -205,6 +168,7 @@ After this change:
 
 - Whitehall is authoritative for taxon tagging on Whitehall-managed content.
 - If Content Tagger mutates taxons for Whitehall content, those changes will be overwritten on next republish.
+- NB: Content Tagger as an application has [low and poorly defined usage](https://gds.slack.com/archives/C0A5B765LUV/p1771244116138289?thread_ts=1770918971.536859&cid=C0A5B765LUV) and has not received meaningful updates in several years.
 
 Optional mitigations:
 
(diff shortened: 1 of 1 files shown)


rfc-188-make-whitehall-the-authority-on-its-taxons.md
```diff
@@ -0,0 +1,282 @@
...
+### 5. Remove runtime dependency
+
+Remove all runtime reads of taxon tagging from Publishing API.
+
+## Governance Considerations
+
+After this change:
+
+- Whitehall is authoritative for taxon tagging on Whitehall-managed content.
+- If Content Tagger mutates taxons for Whitehall content, those changes will be overwritten on next republish.
```
leimdorfer, line 170: Is this a change? What happens currently? If a user mutates a taxon would they expect existing content to reflect that without a republish or is this consistent with current expectations?

ryanb-gds (commented):
Would like to see some data confirming that users are having poor experiences in production - I've been under the impression  this mostly affects us in development/integration. Also, do we know enough about how users work with Content Tagger to be sure this won't cause friction?

One other option to improve this area would be to poll Publishing API for taxon data occasionally, similar to what we're considering doing for image uploads.

I think this is probably a good idea but just want to make sure we've done our due diligence. Also, we need to make sure we run this past the tagging workflow team who are working on tagging flows at the moment. Could cause problems if they were banking on making Content Tagger the source of truth or something.

RukhsanaAli-debug:
How would this work align to future ambitions to create a graph database of GOV.UK content aligned to out core ontology? It's a bit in the future for the moment but I wonder whether the need in the RFC is covered in that work?

rfc-188-make-whitehall-the-authority-on-its-taxons.md
ChrisBAshton: Yep - currently it's possible to edit the tags of Whitehall content, inside Content Tagger. The changes are immediately applied to the live content. The changes are also surfaced in the Whitehall UI, so it is possible for the Whitehall publisher to 'undo' the change by explicitly changing the tagged topics in Whitehall before republishing.

Tested this by:

- Loading up a news article: https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/standard-editions/1582312
- Editing its tags in Content Tagger: https://content-tagger.integration.publishing.service.gov.uk/taggings/5eb5c394-7631-11e4-a3cb-005056011aef
- Seeing the tags immediately applied/published in https://www.integration.publishing.service.gov.uk/government/news/scottish-secretary-attends-royal-national-mod

ChrisBAshton:
@ryanb-gds yep there's an ongoing [thread in #govuk-publishing-tagging-workflow](https://gds.slack.com/archives/C0A5B765LUV/p1770918971536859) to give them an early look at the RFC.

I do expect that the poor UX issue doesn't really apply to Production - but probably more through luck than good design.

ChrisBAshton:
@RukhsanaAli-debug definitely one to discuss and align on. I see this as a step in the right direction that we can iterate on later. Importantly the taxons themselves are still defined outside of Whitehall - this RFC is just about how those taxons are integrated with the app.

chrisroos:
Thanks @ChrisBAshton. This seems like an improvement to me. 

I like that it solves (or at least makes it easier to solve) the existing bug which allows publishers to modify live Topic/World Taxonomy associations by editing them in the draft of a published document in Whitehall.

It should also make it easier to modify the [TaxonValidator in Whitehall](https://github.com/alphagov/whitehall/blob/main/app/validators/taxon_validator.rb) so that it differentiates between the Topic and World Taxonomy associations. At present it's possible to satisfy the TaxonValidator by tagging to the World Taxonomy _instead of_ the Topic Taxonomy and it's not clear whether that's by design or not.

If I understand correctly this won't require changes to any rendering apps. Is it worth making that explicit?

Is it worth making it clear that this makes Whitehall the canonical source of Topic/World Taxonomy associations for Whitehall content _only_ (i.e. Content Tagger remains the canonical source of these associations for any non-Whitehall content)?

In case you're not already aware: not all Whitehall-published content items in the content store have Topic Taxonomy associations even though the validation in Whitehall now requires them to. What you're proposing won't make the situation any worse/better but it'll be useful to be aware of when you come to backfilling data.

In case it's useful: as part of the work we're doing we've created [this spreadsheet](https://docs.google.com/spreadsheets/d/14j-BZ-ka935s0XTjCEFilS4yVYcj8Mo3K7qUd0iCVCU/edit?gid=26672664#gid=26672664) which (among other things) lists all 174 document types in the content store and the counts of those documents with associations to the Topic Taxonomy. The queries for extracting this data from content-store are in [this gist](https://gist.github.com/chrisroos/c03dde835b0426cc788256160db32c57) and we'd be very happy to discuss it further.

Finally, I have very vague recollections from the distant past about a desire to make Publishing API/Content Store the source of truth for all data and that the publishing apps should/would be fairly thin clients on those APIs. I'm guessing that may no longer be a driving factor but thought I'd mention it in case it's useful.

rfc-188-make-whitehall-the-authority-on-its-taxons.md
```diff
@@ -0,0 +1,210 @@
...
+- Remove `patch_links`.
+- Use edition_links only, ensuring draft changes do not affect live until publish.
+
+### Risk: Backfill misses world taxonomy
+
+Mitigation:
+- Explicitly backfill both topic and world taxons.
+- Validate counts before and after switching.
+
+### Risk: A taxon is deleted from Content Tagger
```
sihugh, line 199: Do we have a good feel for how often this occurs? (I also assume that moves/renames don't really matter?)

If there _is_ major rework of the topic taxonomy (eg to align with app taxon, or just a refresh because it's been a while), would this this introduce any different problems for the team doing it?

rfc-188-make-whitehall-the-authority-on-its-taxons.md
```diff
@@ -0,0 +1,282 @@
...
+
+For a limited period:
+
+- Compare local taxon set vs Publishing API link set.
+- Log diffs.
+- Identify edge cases.
+
+### 3. Switch UI reads
+
+Switch Whitehall to render from local persistence only.
```
leenagupte, line 189: Is the plan to do a document type at a time, or a big bang of everything at once? Considering the number of documents in whitehall, I'd pitch for the former.

rfc-188-make-whitehall-the-authority-on-its-taxons.md
```diff
@@ -0,0 +1,282 @@
...
+## Governance Considerations
+
+After this change:
+
+- Whitehall is authoritative for taxon tagging on Whitehall-managed content.
+- If Content Tagger mutates taxons for Whitehall content, those changes will be overwritten on next republish.
+
+Optional mitigations:
+
+- Prevent Whitehall content from being edited in Content Tagger.
```
leenagupte, line 211: So you don't want to remove Whitehall content from Content Tagger, or you'll use the ability to tag it to Mainstream browse.

But, as with Organisations, I'd remove the ability to tag Whitehall content with Taxons in Content Tagger.

rfc-188-make-whitehall-the-authority-on-its-taxons.md
leenagupte: This is a good point. And a bit of a worry.

In the past we've only removed a taxon at time. And in that case the content tagged to the removed taxon has been re-tagged to a higher level, if the removed taxon had a parent. (I can't remember what happened if a removed taxon didn't have a parent)

rfc-188-make-whitehall-the-authority-on-its-taxons.md
```diff
@@ -216,67 +180,31 @@ Optional mitigations:
...
-
-## Impact
-
-- Immediate UX improvement for tagging.
-- Reduced runtime coupling.
-- Consistent association model across Whitehall.
-- Clear ownership model for taxon tagging.
-- Explicit handling of topic + world taxonomy together.
-- Paves the way for easily configuring whether topic/world taxonomies should be required on a per-content type basis (following Whitehall's new config-driven approach).
+- Update the [Taxonomy::RedisCacheAdapter](https://github.com/alphagov/whitehall/blob/07d624404ba3dd3fd06d5bbc241354c4a57ecb46/lib/taxonomy/redis_cache_adapter.rb#L1-L2) such that, on update, it cross-references all taxons by their content ID in the `edition_taxon_links` table and drops any rows where the taxon no longer exists.
```
leenagupte, line 202: So the expectation wouldn't be to move the documents up a level in the taxon tree?

leenagupte (commented):
On the whole I very much agree with this approach.

I have a couple of inline questions.

nacnudus:
Will breadcrumbs of taxon links still work?

The breadcrumb links "Home > Going and being abroad > British nationals overseas" in the screenshot are derived from a single taxon (in this case) that the document is linked to. If that link is an edition link, then Publishing API won't follow the taxonomy up to "Going and being abroad" and "Home". Or does this RFC propose that Whitehall will do link expansion?

image

https://www.gov.uk/government/publications/latvia-list-of-medical-facilities 

> edition links have a substantial limitation, which is that [recursive link expansion](https://docs.publishing.service.gov.uk/repos/publishing-api/link-expansion.html#recursive-links) is not applied to them. If a document needs to access data from documents more than one link away, then link set links must be used.
https://docs.publishing.service.gov.uk/repos/publishing-api/link-expansion.html#link-lifecycles

A second question, regarding

> Enables different topic/world taxonomies on draft vs live content.

Links between taxons themselves are link-set links, so while this proposal (as I understand it) would allow draft and live editions of a given document to be tagged to a different taxon, it wouldn't allow for draft and live editions of a given _taxon_ to be tagged to different parents. The taxonomy itself still couldn't differ between draft vs live content. Is that right?

rfc-188-make-whitehall-the-authority-on-its-taxons.md
```diff
@@ -0,0 +1,210 @@
...
+
+Whitehall will continue to fetch and cache the list of available taxons from the publishing stack.
+
+This proposal changes tagging ownership, not taxonomy ownership. The question of taxonomy ownership and governance is being investigated separately by the `#govuk-publishing-tagging-workflow` team.
+
+### 3. Update Publishing API integration
+
+Taxon links will no longer be managed via `patch_links`.
+
+Instead, taxons will be sent via `put_content` as edition links, like Whitehall's other associations.
```
ChrisBAshton, line 132: Starting a thread here to address @nacnudus 's question:

> Will breadcrumbs of taxon links still work?
> The breadcrumb links "Home > Going and being abroad > British nationals overseas" in the [screenshot](https://private-user-images.githubusercontent.com/3522552/554219785-8e3e32d1-0b85-4177-9c98-30971214a54b.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzIwOTE5MzIsIm5iZiI6MTc3MjA5MTYzMiwicGF0aCI6Ii8zNTIyNTUyLzU1NDIxOTc4NS04ZTNlMzJkMS0wYjg1LTQxNzctOWM5OC0zMDk3MTIxNGE1NGIucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI2MDIyNiUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNjAyMjZUMDc0MDMyWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MmE4MDRlNDVkYTRkOTI2NTcwOGM4NzVmNGM1NTdkYjVhNzZhODQ1Y2EyNzk2M2ZiMjA3ZGU2ZWE0MTQ3OWQxZiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.4eGDy0_2FCoaegWUoAD3FD3suygylq8kOuk4-VOLUuo) are derived from a single taxon (in this case) that the document is linked to. If that link is an edition link, then Publishing API won't follow the taxonomy up to "Going and being abroad" and "Home". Or does this RFC propose that Whitehall will do link expansion?

Oof, good catch.

Currently, topics are [sent to Publishing API via `patch_links`](https://github.com/alphagov/whitehall/blob/99385ffc650d1692e2019ef241de2800fe10c42f/app/services/edition_taxon_link_patcher.rb#L5), i.e. link set links. In our new StandardEdition way of doing things, we send both [edition links (as part of the content payload)](https://github.com/alphagov/whitehall/blob/2b406d9bb189b7cce7956cb6615aa66c29bad7ed/app/presenters/publishing_api/standard_edition_presenter.rb#L24) and [link set links (via patch_links)](https://github.com/alphagov/whitehall/blob/2b406d9bb189b7cce7956cb6615aa66c29bad7ed/app/presenters/publishing_api/standard_edition_presenter.rb#L32-L34) (introduced in https://github.com/alphagov/whitehall/commit/2d61dfb2d45c3475cd0b038213eef1e8879c588e), which is how many legacy Whitehall content types are configured ([example](https://github.com/alphagov/whitehall/blob/7b480cc0ab2d5668951064fdd725a3108068f712/app/presenters/publishing_api/call_for_evidence_presenter.rb#L31-L51)). I don't really know why - it was [set up that way 8 years ago](https://github.com/alphagov/whitehall/commit/69a1ee453b5fe90fa6dfebab5bcb9966026220b2) with a view to gradually tweasing the link set links and edition links apart.

Reading [the docs](https://docs.publishing.service.gov.uk/repos/publishing-api/link-expansion.html#put-content-edition-links), if we send both edition links and link set links, then the latter gets used - and the latter doesn't do link expansion. So you're right that if we adopt the above approach for topics, we'd send edition links and we'd lose the breadcrumb trail.

## Overview of Frontend's breadcrumb logic

The rendering logic for the frontend (in the example, at least) is:

- [Follows the `use_taxon_breadcrumbs?` path](https://github.com/alphagov/govuk_publishing_components/blob/20cb813d653187f408727a27a28d8f6c1f244a45/lib/govuk_publishing_components/presenters/breadcrumb_selector.rb#L63-L67) of the BreadcrumbSelector of the ContextualBreadcrumbs component
- Which [calls](https://github.com/alphagov/govuk_publishing_components/blob/20cb813d653187f408727a27a28d8f6c1f244a45/lib/govuk_publishing_components/presenters/contextual_navigation.rb#L33) the ContentBreadcrumbsBasedOnTaxons class
- Which [loops through all the 'parents'](https://github.com/alphagov/govuk_publishing_components/blob/de9dc45a2fd57af679c1cbb8e40c615d1319dab1/lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_taxons.rb#L14-L20) of a [content item's (single) "parent taxon"](https://github.com/alphagov/govuk_publishing_components/blob/de9dc45a2fd57af679c1cbb8e40c615d1319dab1/lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_taxons.rb#L37)
- Which grabs the [first linked taxon, alphabetically](https://github.com/alphagov/govuk_publishing_components/blob/de9dc45a2fd57af679c1cbb8e40c615d1319dab1/lib/govuk_publishing_components/presenters/content_item.rb#L22-L26) if there is more than one (!!)

In the [example's content item](https://www.gov.uk/api/content/government/publications/latvia-list-of-medical-facilities) we see `links.taxons` for:

- British nationals overseas
- Living abroad
- Travel abroad
- Living in Latvia

...so the rendering logic explains why "British nationals overseas" wins out as the choice of breadcrumb.

That taxon has `links.parent_taxons` for "Going and being abroad", which itself has a `links.parent_taxons` for the root taxon "GOV.UK homepage". These form the first two links of the three breadcrumbs.

## Could Whitehall send expanded links?

> does this RFC propose that Whitehall will do link expansion?

It didn't, but let's explore that.

We already have [logic that filters out ancestors](https://github.com/alphagov/whitehall/blob/99385ffc650d1692e2019ef241de2800fe10c42f/app/services/edition_taxon_link_patcher.rb#L7-L31) when determining which topics to send, so that we only send the most specific taxon.

We could drop that logic altogether and send all the parent taxons, e.g. "GOV.UK homepage" and "Going and being abroad", in the example above. But I don't think that would work. Frontend would just choose the first tagged taxon alphabetically (which would still be "British nationals overseas"), but because it's an edition link rather than a link set link, my assumption is that that would have no `links.parent_taxons` value? In which case Frontend would have a breadcrumb of one.

Are you suggesting that we could send a nested `links.taxons` structure? I'm not sure that's possible, given we're only sending a [flat array of content IDs](https://github.com/alphagov/whitehall/blob/99385ffc650d1692e2019ef241de2800fe10c42f/app/services/edition_taxon_link_patcher.rb#L24) currently 🤔 

## Where does that leave us?

In the long term, breadcrumbs probably ought to be rearchitected somewhat. It does seem a pretty big oversight that [publishers can't specify which tagged topic drives the breadcrumbs of their page](https://gov-uk.atlassian.net/browse/WHIT-2633). As part of that rearchitecture, we could consider changing the publishing platform to allow Whitehall to send a nested structure of topic taxons. Or, probably better, [complete the work to allow recursive expansion of edition links](https://github.com/alphagov/publishing-api/pull/2605).

In the short to medium term, I'd suggest we can still proceed with this RFC, but would have to add some unwanted configurable logic to the [content.links](https://github.com/alphagov/whitehall/blob/2b406d9bb189b7cce7956cb6615aa66c29bad7ed/app/presenters/publishing_api/standard_edition_presenter.rb#L24) return value in StandardEditionPresenter, to filter out some links (topics) from the edition links, sending them only via patch_links in the (link set links) `links` method.

ChrisBAshton:
> While this proposal would allow draft and live editions of a given document to be tagged to a different taxon, it wouldn't allow for draft and live editions of a given _taxon_ to be tagged to different parents. The taxonomy itself still couldn't differ between draft vs live content. Is that right?

Correct - we wouldn't touch the taxon hierarchy side of things. We just receive the [latest expanded topic taxons (without drafts)](https://github.com/alphagov/whitehall/blob/a9aeed7c09d0490fc1cdbe8f3777b5db5b9e2341/lib/taxonomy/publishing_api_adapter.rb#L33) and surface those as options in the UI.

rfc-188-make-whitehall-the-authority-on-its-taxons.md
ChrisBAshton: Good spot. I didn't know that Content Tagger offered an auto-retagging-to-the-parent option when deleting a taxonomy (though it's also optional, which might explain why we've seen a number of live pieces of Whitehall content with no topic tagged - the assumption is that the topic they were once tagged to has been deleted!).

That does complicate things.

## Testing

Tagged a document to a topic on integration:

> Screenshot 2026-02-26 at 09 04 38

Deleted the topic in Content Tagger. Left the "Tag content to parent?" checkbox checked, as per the default:

> Screenshot 2026-02-26 at 09 05 36

Topic now deleted:

> Screenshot 2026-02-26 at 09 11 54

Sure enough, after running `rails taxonomy:rebuild_cache` inside Whitehall, the child taxon option was removed from the UI and the document had been automatically retagged to the parent taxon:

> Screenshot 2026-02-26 at 09 08 37
> Screenshot 2026-02-26 at 09 15 31

## What can we do about this?

Whitehall repopulates its Taxons cache via [calls](https://github.com/alphagov/whitehall/blob/a9aeed7c09d0490fc1cdbe8f3777b5db5b9e2341/lib/taxonomy/publishing_api_adapter.rb#L33) to `get_expanded_links`.

We can then cross-reference the content IDs of received taxons, with the stored taxon choices of content items in Whitehall, surfacing any that are now 'missing'.

We can verify that a missing taxon has been deleted, by calling `get_content` and checking its `publication_state`:

```
Services.publishing_api.get_content("d44bcec9-4da8-4fa4-b548-993a826c5327").to_h
=> 
{"auth_bypass_ids"=>[],
 "base_path"=>"/business/uk-energy-security",
 "content_store"=>"live",
 "details"=>{"internal_name"=>"UK energy security [P]", "notes_for_editors"=>"", "visible_to_departmental_editors"=>false},
 "document_type"=>"taxon",
 "first_published_at"=>"2018-03-08T16:23:44Z",
 "last_edited_at"=>"2018-09-03T15:41:05Z",
 "major_published_at"=>"2018-09-03T15:41:19Z",
 "phase"=>"live",
 "public_updated_at"=>"2018-09-03T15:41:04Z",
 "published_at"=>"2018-09-04T09:24:49Z",
 "publishing_app"=>"content-tagger",
 "publishing_api_first_published_at"=>"2018-03-08T16:23:44Z",
 "publishing_api_last_edited_at"=>"2018-09-04T09:24:48Z",
 "redirects"=>[],
 "rendering_app"=>"collections",
 "routes"=>[{"path"=>"/business/uk-energy-security", "type"=>"exact"}],
 "schema_name"=>"taxon",
 "title"=>"UK energy security",
 "user_facing_version"=>5,
 "update_type"=>"minor",
 "publication_state"=>"unpublished",
 "content_id"=>"d44bcec9-4da8-4fa4-b548-993a826c5327",
 "unpublishing"=>
  {"type"=>"redirect",
   "explanation"=>nil,
   "alternative_path"=>"/education/academy-funding",
   "redirects"=>[{"path"=>"/business/uk-energy-security", "type"=>"exact", "destination"=>"/education/academy-funding"}],
   "unpublished_at"=>nil},
 "locale"=>"en",
 "lock_version"=>9,
 "updated_at"=>"2026-02-26T09:06:13Z",
 "state_history"=>{"4"=>"superseded", "5"=>"unpublished", "3"=>"superseded", "2"=>"superseded", "1"=>"superseded"},
 "links"=>{}}
```

And we can retrieve its parent taxons by calling `get_expanded_links`:

```
Services.publishing_api.get_expanded_links("d44bcec9-4da8-4fa4-b548-993a826c5327", with_drafts: false).to_h
=> 
{"generated"=>"2026-02-26T09:06:13Z",
 "expanded_links"=>
  {"parent_taxons"=>
    [{"links"=>
       {"parent_taxons"=>
         [{"links"=>
            {"root_taxon"=>
              [{"links"=>{},
                "title"=>"GOV.UK homepage",
                "locale"=>"en",
                "api_path"=>"/api/content/",
                "base_path"=>"/",
                "withdrawn"=>false,
                "content_id"=>"f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a",
                "schema_name"=>"homepage",
                "document_type"=>"homepage",
                "public_updated_at"=>"2026-02-18T13:00:23Z",
                "analytics_identifier"=>nil}]},
           "phase"=>"live",
           "title"=>"Business and industry",
           "locale"=>"en",
           "details"=>{"internal_name"=>"Business and industry", "notes_for_editors"=>"", "visible_to_departmental_editors"=>true},
           "api_path"=>"/api/content/business-and-industry",
           "base_path"=>"/business-and-industry",
           "withdrawn"=>false,
           "content_id"=>"495afdb6-47be-4df1-8b38-91c8adb1eefc",
           "description"=>"",
           "schema_name"=>"taxon",
           "document_type"=>"taxon",
           "public_updated_at"=>"2018-09-03T15:41:02Z",
           "analytics_identifier"=>nil}]},
      "phase"=>"live",
      "title"=>"Business and the environment",
      "locale"=>"en",
      "details"=>{"internal_name"=>"Business and the environment", "notes_for_editors"=>"", "visible_to_departmental_editors"=>false},
      "api_path"=>"/api/content/business-and-industry/business-and-the-environment",
      "base_path"=>"/business-and-industry/business-and-the-environment",
      "withdrawn"=>false,
      "content_id"=>"13fb7519-c48e-4577-90f8-aaf483fa0c03",
      "description"=>nil,
      "schema_name"=>"taxon",
      "document_type"=>"taxon",
      "public_updated_at"=>"2018-09-03T19:57:48Z",
      "analytics_identifier"=>nil}]}}
```

So **it should be possible to retain existing behaviour** of retagging Whitehall content to the parent of the deleted taxon.

## Manual review step

Thinking this through, there is a potential incident if we're not careful: if Publishing API, for whatever reason, fails to return the expanded links of taxons when Whitehall refreshes its cache, there's a risk that *we could auto-delete all topic taxon selections in Whitehall*.

So in hindsight, it's probably best to _not_ automate removing deleted taxons (and tagging their parents). We can make it a manual step instead - something like:

- Raise exception (captured in Sentry) if any missing taxons are discovered on taxonomy refresh
- Developer checks, then runs rake task to normalise all the tags

I'd hypothesise that taxons are not edited very often ([based on user signin data](https://gds.slack.com/archives/C0A5B765LUV/p1771245262122069?thread_ts=1770918971.536859&cid=C0A5B765LUV)), so it shouldn't happen too often and wouldn't be too arduous a support task.

### Outstanding risks?

There'd still be a risk of a publisher editing a document in between its tagged topic being deleted and a developer patching up the tags to the parent. The numbers of affected documents should be low, and publishers wouldn't be able to publish updates to their document if the deleted taxon meant that they were no longer tagged to any topic. But if a publisher had tagged their doc to two or more topics, and then one topic was deleted, they'd lose the topic on subsequent publish.

This is either a risk we can live with, or we can implement some sort of halfway house. For example we may want to automate the removal and retagging of content *where only one or two topics have been deleted since last cache rebuild*, and anything higher than that indicates possible system/network failure where we want to have the manual review safeguard described above.

rfc-188-make-whitehall-the-authority-on-its-taxons.md
ChrisBAshton: Moving this discussion to https://github.com/alphagov/govuk-rfcs/pull/188#discussion_r2857999471

rfc-188-make-whitehall-the-authority-on-its-taxons.md
ChrisBAshton: Noted - thanks. Gosh, I can see why things have just been left as-is! There are more moving parts than I'd thought.

So for the record, we can [tag a piece of Whitehall content in Content Tagger](https://content-tagger.integration.publishing.service.gov.uk/taggings/e95d0121-d252-4c1a-9380-791e2a715312):

> Screenshot 2026-02-26 at 09 58 49

I see it reflected in the Whitehall content item under `links.mainstream_browse_pages`:

Screenshot 2026-02-26 at 09 55 40

It doesn't appear to have any effect, but maybe that's because Frontend [only looks for a mainstream browse tag on the _parent_ document](https://github.com/alphagov/govuk_publishing_components/blob/20cb813d653187f408727a27a28d8f6c1f244a45/lib/govuk_publishing_components/presenters/contextual_navigation.rb#L75-L77), and my example didn't have a parent 🤔 

But yes, I can tweak the above part of the RFC so that Whitehall content is _not_ removed from / disabled in Content Tagger, but perhaps just remove the "Topic taxonomy tags" section of the form in Content Tagger when the content originates from Whitehall.

rfc-188-make-whitehall-the-authority-on-its-taxons.md
hannako: > In the long term, breadcrumbs probably ought to be rearchitected somewhat. It does seem a pretty big oversight that [publishers can't specify which tagged topic drives the breadcrumbs of their page](https://gov-uk.atlassian.net/browse/WHIT-2633).

It's perhaps not oversight but the lack of a good alternative. Giving publishers the tool to curate their breadcrumbs would be a non-trivial piece of work for content teams - for example providing guidance so that similar pieces of content aren't being tagged to different taxons.

ChrisBAshton:
I am **closing** this RFC as **not accepted**.

---

The proposal aimed to give Whitehall more responsibility for tracking what topic/world taxonomies its documents are tagged to. The main reasons this felt like a good idea:

1. Solves the bug where sometimes a document's tagged topics/taxons are 'delayed' before they're shown in the UI
2. Simplifies Whitehall's codebase, to have taxons work more like Whitehall's other 'associations'
3. Gives us edition level topics/taxons 'for free', making it possible to stage changes to a document's tagged taxons

However, 3 [turned out not to be true](https://github.com/alphagov/govuk-rfcs/pull/188#discussion_r2857704818). Edition links don't support recursive expansion, so there'd be a negative side-effect of switching to them in that breadcrumb hierarchies would no longer be available to the frontend. This is a **breaking change that is not an acceptable outcome of the RFC**.

To work around 3, we'd have to introduce bespoke logic in Whitehall's StandardEditionPresenter to omit taxons from the 'edition links' sent to Publishing API, retaining it only in the 'link set links' payload. That already lessens the potential benefit of 2.

In addition, the RFC failed to account for existing behaviour in Content Tagger, whereby [documents tagged to a topic that is being deleted can be automatically 're-tagged' to the parent topic](https://github.com/alphagov/govuk-rfcs/pull/188#discussion_r2826558810). Losing this functionality is an **additional breaking change that did not feel like an acceptable outcome of the RFC**.

To implement said behaviour in Whitehall, we'd have to add complex logic at the point where Whitehall refreshes its taxonomy cache. If any Whitehall content continued to reference taxons that no longer exist, we'd need to perform additional lookups in Publishing API to figure out the deleted taxon's nearest "ancestor" and re-tag all documents to that. If no such ancestors exist, we'd delete the tagging altogether.

This already reduces much of the alleged benefit of 2, because we'd be adding this re-tagging logic that Whitehall currently doesn't need to know about. But it also underlined a significant risk, late in the day - that auto-deleting tags from Whitehall content in response to missing topics at revalidation time is **inherently risky**. We run the risk of one bad API response causing Whitehall to delete all of its tagging data. This, again, is not a risk under the current Publishing API led system.

All of the above are reasons alone not to proceed with the RFC, especially given how limited the remaining value would be versus the time it would take to implement.

For completion, there were other reasons not to proceed with the RFC further.

First off, this proposal only helps Whitehall - not other apps. We'd be encoding Whitehall specific domain knowledge into Content Tagger (to stop people from being able to do general topic tagging to Whitehall content in the Mainstream interface, but still having to surface Whitehall content in the first place so that it can be tagged to Mainstream).

There is also another team considering tagging more widely, and given the above caveats, it does seem like a proper architectural solution to tagging is team sized, rather than something that can be chipped away at by individual teams.

PR #189, opened 2026-04-20
✨ Add RFC-178: Require Dependabot cooldown for external auto-merging

jasonBirchall:
## 👀 Purpose

- Amend RFC-167 to restore external dependency auto-merging in govuk-dependabot-merger, gated behind a Dependabot cooldown requirement of at least 3 days.

## ♻️ What's changed

- Added rfc-178-require-dependabot-cooldown-for-external-auto-merging.md proposing the new constraint
- Updated rfc-167-auto-patch-dependencies.md Status section to reference RFC-178

## 📝 Notes

- PR 131 in govuk-dependabot-merger removed external auto-merging entirely. This RFC brings it back with one
guard rail: the repo's .github/dependabot.yml must have cooldown default-days >= 3. If it doesn't, external PRs get blocked and a warning is logged.
- No code changes yet. Implementation follows once the RFC is accepted.

### DIFF
--- a/rfc-178-require-dependabot-cooldown-for-external-auto-merging.md
+++ b/rfc-178-require-dependabot-cooldown-for-external-auto-merging.md
@@ -1,0 +1,94 @@
+---
+status: proposed
+implementation: proposed
+status_last_reviewed: 2026-04-20
+---
+
+# Require Dependabot cooldown for external dependency auto-merging
+
+## Summary
+
+[RFC-167][] allowed repositories to opt into auto-merging external dependencies via [govuk-dependabot-merger][]. [PR #131][pr-131] removed that ability as a supply chain hardening measure. This RFC brings it back, with one new rule: the repository's `.github/dependabot.yml` MUST configure a [cooldown][] of at least 3 days (`default-days: 3` or higher).
+
+## Problem
+
+Removing external auto-merging was the right call at the time. But it shifted the burden back onto teams who now have to manually review and merge low-risk external updates that CI, security scanning, and test suites already validate.
+
+The original argument in RFC-167 still stands: human review of dependency PRs is largely theatre. Over 90% of engineers weren't following the review guidance anyway, and even those who did couldn't meaningfully catch a determined supply chain attack by reading a changelog.
+
+What's changed since PR #131 is that Dependabot now supports [cooldown][]. Cooldown holds off raising a PR until a package version has been publicly available for a configurable number of days. Malicious or broken releases tend to get flagged within hours. A 3-day window catches the vast majority of these before a PR ever gets opened.
+
+## Proposal
+
+Bring back `update_external_dependencies` in govuk-dependabot-merger. Add one gate: govuk-dependabot-merger checks the repo's `.github/dependabot.yml` for a `cooldown` setting with `default-days` >= 3. If that's missing or too low, external PRs don't get merged and a warning gets logged.
+
+Repos that never set `update_external_dependencies` won't notice any change. The default stays `false`.
+
+### Why 3 days?
+
+Most malicious packages get pulled within hours. Three days is conservative enough to catch stragglers while still being useful. It also matches what we already use on [govuk-dependabot-merger itself][own-dependabot-yml].
+
+### Configuration
+
+Two files need to agree. In `.govuk_dependabot_merger.yml`:
+
+```yaml
+api_version: 2
+defaults:
+  update_external_dependencies: true
+  auto_merge: true
+  allowed_semver_bumps:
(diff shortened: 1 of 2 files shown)


rfc-178-require-dependabot-cooldown-for-external-auto-merging.md
```diff
@@ -0,0 +1,94 @@
...
+
+If a repo opts in but doesn't meet the cooldown requirement, govuk-dependabot-merger blocks external merges and logs a warning explaining why.
+
+### Implementation
+
+Changes to [govuk-dependabot-merger][]:
+
+1. Restore `update_external_dependencies` (reverts [PR #131][pr-131])
+2. Fetch the repo's `.github/dependabot.yml` and check cooldown `default-days` >= 3
+3. Log a warning when cooldown is missing or insufficient
```
leenagupte, line 85: Where will this log show up?

leenagupte (commented):
This sounds like a really good plan. Just one inline question about logging.

rfc-178-require-dependabot-cooldown-for-external-auto-merging.md
sihugh: Could it be added to Dependapanda slack bot?

rfc-178-require-dependabot-cooldown-for-external-auto-merging.md
```diff
@@ -0,0 +1,94 @@
...
+- Repositories MUST explicitly set `update_external_dependencies: true` (default remains `false`)
+
+If a repo opts in but doesn't meet the cooldown requirement, govuk-dependabot-merger blocks external merges and logs a warning explaining why.
+
+### Implementation
+
+Changes to [govuk-dependabot-merger][]:
+
+1. Restore `update_external_dependencies` (reverts [PR #131][pr-131])
+2. Fetch the repo's `.github/dependabot.yml` and check cooldown `default-days` >= 3
```
sihugh, line 84: 👍

sihugh: I think `include` and `exclude` lists within the `cooldown` config might be interesting at some point (such as checking that excluded dependencies are internal somehow), but I think that's outside the scope of this RFC.

sihugh (approved):
I think this is a useful change

rfc-178-require-dependabot-cooldown-for-external-auto-merging.md
AgaDufrat: I think this check could be added to Seal, similar to how we check for [security workflows in the ci](https://github.com/alphagov/seal/blob/f2275678aa1540cd42a9caa4ee266aa258483c31/lib/github_fetcher.rb#L222-L246) and run on a [daily schedule](https://github.com/alphagov/seal/blob/main/.github/workflows/ci_checks.yml).

rfc-178-require-dependabot-cooldown-for-external-auto-merging.md
AgaDufrat: In ~Publishing Platform~ Content APIs we introduced the `cooldown` for all the things for now, accepting that we will need to bump internal gems manually. 
Looks like Dependabot Merger already has logic to check if a dependency [is internal](https://github.com/alphagov/govuk-dependabot-merger/blob/ca7eebcda8df087b915e1387280295ac91885c87/lib/change_set.rb#L2-L10) so it shouldn't be too hard to check the `include` and `exclude` lists.

jasonBirchall: The `include` and `exclude` options are valuable to the service team looking to merge internal dependencies quickly, but I'm not sure how useful they'd be to the dependabot merger. Maybe we should add this to the [recommendation documentation](https://docs.publishing.service.gov.uk/manual/manage-dependencies.html)?

AgaDufrat: How I understand Simon's comment is that Dependabot Merger will check `.github/dependabot.yml` and as long as "cooldown `default-days` >= 3", the check will pass but the config may:
- `exclude` a bunch of external dependencies or
- `include` only a subset of dependencies.

This would leave a bit of a loophole making the cooldown validation incomplete.

AgaDufrat: I agree with "I think that's outside the scope of this RFC", as it more of a check implementation detail.

ChrisBAshton (approved):
LGTM 👍

PR #192, opened 2026-06-16
[RFC-192] Restrict base_paths to a smaller character set

KludgeKML:
## Summary

The validations we run on base paths on GOV.UK are too permissive, making it difficult to easily filter out invalid ones. The systems we currently have for filtering valid requests vary across different apps, and are not in line with the intuition of our developers. We propose restricting _non-redirect_ base paths to a much smaller character set (`a-z 0-9 . / -`). Only 0.013% (131) of our content items would need to be changed to get to complete compliance with this more restrictive set.

[Rendered view](https://github.com/alphagov/govuk-rfcs/blob/174b72ebfcc05ac02bff61628097603e7952bd9c/rfc-192-restrict-base-paths-to-a-smaller-character-set.md)

https://gov-uk.atlassian.net/browse/PNP-10051


maxgds:
This appears to be very sensible and I'm not seeing any obvious issues. Would be good to fix content blocks early rather than when they're more widespread. One thing that puzzles me is 511 characters - can you add some supporting explanation as it seems to come out of nowhere as a need with no supporting evidence.

KludgeKML:
> This appears to be very sensible and I'm not seeing any obvious issues. Would be good to fix content blocks early rather than when they're more widespread. One thing that puzzles me is 511 characters - can you add some supporting explanation as it seems to come out of nowhere as a need with no supporting evidence.

@maxgds Good point. This is an existing requirement (relatively new I think?) that came from the search team. Base paths 512 characters and over were breaking search. I'll find a reference to when it was changed and add an explanation.

(Now added)

brucebolt (commented):
This seems like a very sensible change to make.

I recall issues when we launched the `/euexit` landing page, where redirects had to be manually put in place for `/EUExit`, `/EUexit` etc, so feel like this would stop that occurring in the future.

My only concern would be non-English content: does the `a-z` restriction also include accented or other characters that don't appear in English?  Or would we expect the publishing app to substitute the characters with the nearest equivalent in the 26 characters used in English?  Maybe they already do this?

KludgeKML:
> This seems like a very sensible change to make.
> 
> I recall issues when we launched the `/euexit` landing page, where redirects had to be manually put in place for `/EUExit`, `/EUexit` etc, so feel like this would stop that occurring in the future.
> 
> My only concern would be non-English content: does the `a-z` restriction also include accented or other characters that don't appear in English? Or would we expect the publishing app to substitute the characters with the nearest equivalent in the 26 characters used in English? Maybe they already do this?

@brucebolt Interesting question! I suspect substitution would be reasonable. Given that gov.uk only really supports non-English language _content_ (ie the surrounding page hardware like header / footer is current English only), that urls being restricted to substitute characters is acceptable.

It's worth considering Welsh, since that's a language we're mandated to support. Welsh has some accents (acute, grave, circumflex, and diaeresis), but we can see that the Welsh government site just substitutes the non-accented form in paths (go to https://www.llyw.cymru/ and hover over [Y môr a physgodfeydd](https://www.llyw.cymru/y-mor-a-physgodfeydd) for an example.

That also seems to be the behaviour on Whitehall - I created a news item in integration titled "Y môr a physgodfeydd pàs gẃraidd copïo" (this is not an actual Welsh sentence, I just tacked on a word with every accent), and the resulting path was `/government/news/y-mor-a-physgodfeydd-pas-graidd-copio` I'll have to find time to look through the other publishing apps, but I suspect either they do the same thing or it would be relatively easy to modify them to do the same thing.

I'll add some notes to this effect to the RFC. Thanks!

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
```diff
@@ -0,0 +1,58 @@
...
+
+1) It's hard for frontend apps to deal quickly with obvious penetration attempts like `GET /find-local-council/%20HTTP/1.1%0D%0X-header:%20attempt`. This is clearly not a reasonable URL, but it _might_ be, so frontend apps have to pass it on to content-store (or the GraphQL endpoint), for them to reject it. And
+although content-store _does_ have code to handle requests like these before they get to a database call, it's not clear that's actually the correct behaviour. This URL _is_ valid according to our current validations, so there aren't safe rules to reject it.
+2) The inclusion of characters in both cases means that we can potentially (and do in fact) have content items with base paths that rely on case. Potentially
+a publishing app could write separate content items to gov.uk/example and gov.uk/Example and they would behave differently. Instinctively people tend to assume
+that urls are case-insensitive, and we have previously put code in to [handle people typing with their caps lock on]. But this has led to an even odder situation
```
ChrisBAshton, line 25: TIL!

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
```diff
@@ -0,0 +1,58 @@
...
+in which gov.uk/EXAMPLE, GOV.UK/EXAMPLE, gov.uk/example would all resolve to the same content item, but gov.uk/Example would not.
```
ChrisBAshton, line 26: Do we have data on how many mixed case URLs resolve to different content items? Seems odd that case sensitivity remained semi-supported after that Puppet PR.

ChrisBAshton: EDIT: I see further down "20 of them are invalid only because they include upper case characters" - but how many of those resolve to different content items than their all-lowercase equivalents? (Indeed, how many _have_ all lowercase equivalents?)

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Accents and other languages
+
+One possible concern with the restriction is that we're limiting base paths to english characters. But since GOV.UK is currently a website in English that supports non-English content (ie we can handle non-English content, but the page surrounds are always in English), limiting urls to english characters seems reasonable. The only additional language we're required to support is Welsh, and the written form of Welsh is made entirely of English characters. Welsh does support accents, but it appears to be acceptable in slugs to substitute non-accented versions of the characters (this is what the Welsh government site does). It seems reasonable that we do the same thing (indeed, Whitehall already does this).
```
ChrisBAshton: Worth adding example of Welsh government site here, for completeness

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Proposal
+
+We accept that base_paths MUST NOT be longer than 511 characters, MUST start with a /, and MUST contain only the characters `a-z`, `0-9`, `.`, `-`, and `/`.
+
+The only exception allowed to this is for content items of `schema_name: redirect`
+
+- A validation method is added to [gds-api-adapters] which tests a base path passed to it against the new restriction. This becomes the source of truth.
```
ChrisBAshton, line 57: I like this a lot. We otherwise have competing validation methods across multiple apps - something that's been a continuous problem.

Has some crossover with this ticket about when Publishing API and Whitehall disagree on what a valid redirect URL is: https://gov-uk.atlassian.net/browse/WHIT-2428

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
```diff
@@ -0,0 +1,58 @@
...
+- [gds-api-adapters] methods which are passed a base path use this new method to validate their paths and return GdsApi::InvalidUrl before calling the underlying service.
+- Uses of [gds-api-adapters] are audited, and we ensure that any app calling the upgraded methods handles GdsApi::InvalidUrl appropriately.
```
ChrisBAshton: Seems outside the scope of the RFC tbh

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
```diff
@@ -0,0 +1,58 @@
...
+in which gov.uk/EXAMPLE, GOV.UK/EXAMPLE, gov.uk/example would all resolve to the same content item, but gov.uk/Example would not.
+
+Analysis of items in content store shows that the regex `/^\/$|^(\/[a-z0-9._-]+)+$/` matches the base paths of all but 2126 content items (out of around 1,008,000), so we have a de facto standard which is adhered to by 99.8% of our content items. If we can somehow handle the remaining 2126 gracefully, we could lock down the allowed paths to only the smaller set, allowing frontend apps to safely reject a number of dubious calls they currently have to handle. We would also be able to put simpler rules into nginx that would lowercase all incoming paths, not just those that are entirely in upper case - this might allow us to retire a number of current redirects as well.
+
+One way of graceful handling would be to allow the potentially invalid urls to continue to be used, but only as redirects. Redirects are currently handled entirely within router, so wouldn't requite any special handling in the frontend apps - only short url manager would have to be excluded from restrictions.
```
ChrisBAshton, line 30: Can you go into more detail on the implications for Short URL Manager in the Proposal section?

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
```diff
@@ -0,0 +1,58 @@
...
+a publishing app could write separate content items to gov.uk/example and gov.uk/Example and they would behave differently. Instinctively people tend to assume
+that urls are case-insensitive, and we have previously put code in to [handle people typing with their caps lock on]. But this has led to an even odder situation
+in which gov.uk/EXAMPLE, GOV.UK/EXAMPLE, gov.uk/example would all resolve to the same content item, but gov.uk/Example would not.
+
+Analysis of items in content store shows that the regex `/^\/$|^(\/[a-z0-9._-]+)+$/` matches the base paths of all but 2126 content items (out of around 1,008,000), so we have a de facto standard which is adhered to by 99.8% of our content items. If we can somehow handle the remaining 2126 gracefully, we could lock down the allowed paths to only the smaller set, allowing frontend apps to safely reject a number of dubious calls they currently have to handle. We would also be able to put simpler rules into nginx that would lowercase all incoming paths, not just those that are entirely in upper case - this might allow us to retire a number of current redirects as well.
```
ChrisBAshton: Can you clarify here where Router fits in and how it will continue to support 'invalid' routes provided they are redirects only?

In fact, splitting this out into a separate "Benefits of implementing the RFC" section might be a good idea.

ChrisBAshton (approved):
Sounds good to me 🎉

leenagupte (commented):
I don't have anything extra to add that isn't covered by other comments, but it all sounds very sensible to me too. 🎉 

I agree with Max about the ordering of events. Cleaning up the existing base paths, such as the ones for content blocks, should come first.

young-rob:
Looking like a great idea to me 👍 

Does the order in which we carry out these proposed changes matter, and if so do we need to specify it? e.g. do we need make sure to do the reslugging/base path alterations before we make the nginx changes to prevent 404s?

Also, the regex shown that was used to analyse content store items to get the 99.8% figure include an `_` - has that just snuck into the RFC doc by mistake or do we need to run the regex analysis again?

KludgeKML:
> Does the order in which we carry out these proposed changes matter, and if so do we need to specify it? e.g. do we need make sure to do the reslugging/base path alterations before we make the nginx changes to prevent 404s?

I'll refine the list to the correct order plus explanations of why that order)

> Also, the regex shown that was used to analyse content store items to get the 99.8% figure include an `_` - has that just snuck into the RFC doc by mistake or do we need to run the regex analysis again?

Ah yes, that was just a typo, I'll update it. So that people can double-check my work if necessary, I'm going to add the commands I used to get the details into the JIRA card, plus a spreadsheet of the non-compliant URLs with notes as to why they're non-compliant and (in the case of partial uppercases) whether they could be removed entirely.

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
KludgeKML: I've added some code and a resulting output CSV from running that code against staging to the attached JIRA ticket (wasn't 100% sure if the ticket or the PR was the best place).

We have 451 content items with upper case characters. 430 of them are redirects. 316 are case-only redirects (ie either they redirect directly to the same thing in all lower case, or they redirect to something else, but there's _also_ another redirect all in lower case that redirects to that same something else).

There's exactly 1 clash (ie a non-redirect that isn't the same as the lower-case version. The upper-case version is a `gone`, the lower-case version is a `redirect`. It feels like that's very safe to handle by just deleting the upper-case version once nginx correctly handles all cases.

There are 21 non-redirect, non-contact-block items which include upper case but don't have a lower-case version. They'd have to be reslugged (so that the content item was the lower-case version, but the upper-case version remained as a redirect), but once that was done it would be safe to turn nginx rules to just "rewrite as lower case" and we'd see no ill effect.

I'll cover this when I rework the order of operations.

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
KludgeKML: Done!

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
KludgeKML: Well, maybe - this is just to ensure that we catch breaks at the point we're making the changes. I'll mark it as optional.

KludgeKML:
Thanks for comments all - hopefully the last commit should fill in the blanks and sort out the order of work (and what's optional work)

AgaDufrat:
This is a very sound proposal and well thought through!

I would perhaps elaborate a bit on more the “improved security” in the benefits section, as the benefits of restricting base paths don't seem to be mentioned there:
* Elimination of encoded control characters
* Better protection against path traversal (although I don’t think they could be exploited)

_Also already covered by the second bullet point but it maybe worth calling it out explicitly:_
* Earlier rejection of malformed requests
* Fewer opportunities for inconsistent handling

KludgeKML:
> This is a very sound proposal and well thought through!
> 
> I would perhaps elaborate a bit on more the “improved security” in the benefits section, as the benefits of restricting base paths don't seem to be mentioned there:
> 
> * Elimination of encoded control characters
> * Better protection against path traversal (although I don’t think they could be exploited)
> 
> _Also already covered by the second bullet point but it maybe worth calling it out explicitly:_
> 
> * Earlier rejection of malformed requests
> * Fewer opportunities for inconsistent handling

@AgaDufrat Good suggestions! I've expanded the benefits section slightly to include this.

rfc-192-restrict-base-paths-to-a-smaller-character-set.md
```diff
@@ -0,0 +1,100 @@
...
+
+One way of graceful handling would be to allow the potentially invalid urls to continue to be used, but only as redirects. Redirects are currently handled entirely within router, so wouldn't requite any special handling in the frontend apps - only short url manager would have to be excluded from restrictions.
+
+Fortunately, of the 2126 potentially invalid base paths, all but 131 of them already are redirects, reducing the amount of work significantly. 20 of them are invalid only because they include upper case characters, which means if we allowed nginx to lowercase all paths before passing them through, we could lowercase them without breaking links. 96 are invalid because they include `_`s, which we could reasonably redirect to a path with `-`s.  There are two with both underscores _and_ upper case characters, which we could handle with a combination of the above techniques. Finally there are 13 that are content blocks, which seem to have base_paths of the format `/content-blocks/content_block_<type>/<slug>` - since content blocks are few at the moment, it would perhaps be possible to convert these to use dashes in the second part of the path.
+
```
edavey, line 33: Over in Content Block Manager we've had a good look at how `base_url` is used and we think that we don't currently need to continue to support it. We're planning to withdraw their use and stop publishing blocks to the Content Store. We've drafted an ADR: https://github.com/alphagov/content-block-manager/pull/755 Comments most welcome.

PR #194, opened 2026-07-14
[RFC-194] (WORK IN PROGRESS) Render Govspeak in Frontend

KludgeKML:
It's been a decade since we chose rendering govspeak in the publishing apps, this RFC will explore if that's still the right decision or if there are some benefits to be had by moving rendering to the frontend.

NOTE: this RFC is a lot rawer than the usual ones, I'm opening it up for longer comment/discussion because I suspect there are a great many perspectives on this, and rather than trying to approach everyone independently for synthesis, it's better to just put the idea out in the open and see where the conversation takes us.

[Rendered View](https://github.com/alphagov/govuk-rfcs/blob/fc08f8125738ed4797ca85ce7f583eb99708dd6b/rfc-194-render-govspeak-in-frontend.md)

### DIFF
--- a/rfc-194-render-govspeak-in-frontend.md
+++ b/rfc-194-render-govspeak-in-frontend.md
@@ -1,0 +1,64 @@
+---
+status: proposed
+implementation: proposed
+status_last_reviewed:
+---
+
+# Render Govspeak in Frontend
+
+## Summary
+
+It's been almost a decade since we made the decision (in [RFC-48]) to render [govspeak] in the publishing apps. Recently the concept of rendering it in frontend has come up, and in this RFC we'll look at whether it might make more sense to render govspeak in the frontend apps instead.
+
+## Problem
+
+GOV.UK publishing works in a three-step system. In step one editors create documents using (mainly) govspeak. In step two that raw govspeak is converted into an HTML representation, which is then passed on to the data stores ([publishing-api] and [content-store]). Finally the frontend apps request those HTML representations, which they use to compose the pages rendered to the users.
+
+Pros of the current situation:
+- Efficient: each document of raw govspeak only has to be converted to HTML once, at time of publishing
+- Status Quo: No extra work required
+
+Cons of the current situation:
+- When components in govspeak change, all documents using that component have to be republished
+- It puts a timing burden on component changes, since the republish and the gem change have to be planned to prevent CSS/JS mismatch
+- The app can't easily manipulate HTML (ANECDOTAL - CONFIRM WITH APP TEAM)
+- Parsed HTML is generally larger than raw govspeak
+- Components embedded in pre-parsed govspeak can't have A/B tests applied to them easily
+- LLMs prefer markdown to HTML (ANECDOTAL - CONFIRM WITH CHAT TEAM)
+
+## Things to consider
+
+- Is manipulating Govspeak definitely slower than manipulating HTML?
+- Could we speed up Govspeak -> HTML conversion (Could we realistically replace Kramdown with [Markly](https://github.com/socketry/markly), apparently 40x faster)
+- We have extensions to Markdown to form Govspeak, are these extensions all needed if we're converting closer to time of rendering?
+- Will this affect third parties (App team do not use the rendered HTML, does anyone else?)
+- How do we handle inline images?
+- Can the govspeak component handle this automatically? Can it handle HTML _or_ govspeak automatically?
+- Could we have markdown linters or other support for publishing apps?
+- Would this simplify embedding things that the frontend later has to pick out?
+- Would this simplify analytics?
+
+## Potential Benefits
+
+- We could pass markdown to the mobile app through content api
+- Publishing apps do not have to know about markdown->html conversion
+- It might be easier to handle embed codes in frontend apps, improving the flexibility of
+  various document types (for example, a document could have an embed tag for "postcode
+  finder goes here", allowing editors more latitude in what they put before or after
+  the finder).
+- Single representation of blocks of edited text throughout the publishing system
+- We could pass markdown to LLMs, possibly resulting in better answers
+- We could perform A/B tests on components rendered in govspeak
+
+## Proposal
+
+We move Govspeak rendering out of the publishing stack and into the frontend stack.
+
+Changes that would be required:
+- The govspeak component in the [govuk_publishing_components] gem would need to be enhanced to allow it to accept raw govspeak and convert it to html in-app.
+- Publishing apps would need to stop processing govspeak before passing it to the publishing-api
+
+[content-store]: https://github.com/alphagov/content-store
+[govspeak]: https://github.com/alphagov/govspeak
+[publishing-api]: https://github.com/alphagov/publishing-api
+[RFC-48]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-048-govspeak-rendering-and-embedded-content.md


rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+GOV.UK publishing works in a three-step system. In step one editors create documents using (mainly) govspeak. In step two that raw govspeak is converted into an HTML representation, which is then passed on to the data stores ([publishing-api] and [content-store]). Finally the frontend apps request those HTML representations, which they use to compose the pages rendered to the users.
+
+Pros of the current situation:
+- Efficient: each document of raw govspeak only has to be converted to HTML once, at time of publishing
+- Status Quo: No extra work required
+
+Cons of the current situation:
+- When components in govspeak change, all documents using that component have to be republished
+- It puts a timing burden on component changes, since the republish and the gem change have to be planned to prevent CSS/JS mismatch
+- The app can't easily manipulate HTML (ANECDOTAL - CONFIRM WITH APP TEAM)
```
brucebolt, line 24: This has been something on our team's mind for a while.  We [spiked into using syntax trees to represent content](https://beisgov.sharepoint.com/:w:/r/sites/GOVUK-Pub/_layouts/15/Doc.aspx?sourcedoc=%7B8CFE4317-A1BA-49CE-9BF1-083C0D072E98%7D&file=%5BSPIKE%5D%20Publishing%20Investigation%20into%20Syntax%20Trees.docx&action=default&mobileredirect=true) but this still resulted in an output that was more suited to the web.

Returning raw govspeak in the Content API responses could solve this issue, but they'd need to create a library analogous to govspeak which maintains feature parity.

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+- Parsed HTML is generally larger than raw govspeak
+- Components embedded in pre-parsed govspeak can't have A/B tests applied to them easily
+- LLMs prefer markdown to HTML (ANECDOTAL - CONFIRM WITH CHAT TEAM)
+
+## Things to consider
```
brucebolt, line 29: Another thing to consider: some elements get embedded in publishing apps (e.g. attachments and images).  We'd need to determine whether the content item has enough information to render this govspeak into HTML, or whether additional metadata would need to be added.

brucebolt, line 29: Another thing to consider is existing users of Content API that aren't GOV.UK frontend apps, and how they'll be affected.

That's probably something our team can feed into, based on user research we've done.

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+  the finder).
+- Single representation of blocks of edited text throughout the publishing system
+- We could pass markdown to LLMs, possibly resulting in better answers
+- We could perform A/B tests on components rendered in govspeak
+
+## Proposal
+
+We move Govspeak rendering out of the publishing stack and into the frontend stack.
+
+Changes that would be required:
```
brucebolt, line 57: As mentioned in one of my other comments, there's likely additional information we'd need to add into the content items to allow govspeak to be parsed there.

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+- The app can't easily manipulate HTML (ANECDOTAL - CONFIRM WITH APP TEAM)
+- Parsed HTML is generally larger than raw govspeak
+- Components embedded in pre-parsed govspeak can't have A/B tests applied to them easily
+- LLMs prefer markdown to HTML (ANECDOTAL - CONFIRM WITH CHAT TEAM)
+
+## Things to consider
+
+- Is manipulating Govspeak definitely slower than manipulating HTML?
+- Could we speed up Govspeak -> HTML conversion (Could we realistically replace Kramdown with [Markly](https://github.com/socketry/markly), apparently 40x faster)
+- We have extensions to Markdown to form Govspeak, are these extensions all needed if we're converting closer to time of rendering?
```
brucebolt, line 33: Yes, I think they will be as we have extensions for features that are otherwise missing from markdown.

brucebolt (commented):
On the whole, I'd be supportive of this change.  "Why do we render the govspeak in Publishing API and not the frontend app" is a question I've asked myself many times and not convincingly been able to answer it.

I think there's a lot of implementation details that we'd need to work through in Publishing API and the publishing apps, and also figure out how this affects the insertion of content blocks.

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+
+Pros of the current situation:
+- Efficient: each document of raw govspeak only has to be converted to HTML once, at time of publishing
+- Status Quo: No extra work required
+
+Cons of the current situation:
+- When components in govspeak change, all documents using that component have to be republished
+- It puts a timing burden on component changes, since the republish and the gem change have to be planned to prevent CSS/JS mismatch
+- The app can't easily manipulate HTML (ANECDOTAL - CONFIRM WITH APP TEAM)
+- Parsed HTML is generally larger than raw govspeak
```
yndajas, line 25:
```suggestion
- Rendered HTML is generally larger than raw govspeak
```

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+- Components embedded in pre-parsed govspeak can't have A/B tests applied to them easily
+- LLMs prefer markdown to HTML (ANECDOTAL - CONFIRM WITH CHAT TEAM)
+
+## Things to consider
+
+- Is manipulating Govspeak definitely slower than manipulating HTML?
+- Could we speed up Govspeak -> HTML conversion (Could we realistically replace Kramdown with [Markly](https://github.com/socketry/markly), apparently 40x faster)
```
yndajas, line 32: **issue:** it looks like you're already considering performance, but it might be useful context to know that when working on GraphQL, request-time rendering was one of our big performance hurdles

Previously, the Govspeak that got rendered into HTML by Publishing API was not persisted to the Publishing API database. Instead it was rendered when presenting it downstream to Content Store and persisted only in the Content Store database. A side effect of that was that with GraphQL we had to render Govspeak on every request, and flamegraphs revealed that this was taking up a significant chunk of the overall request processing time

As a result, we decided to cache the rendered Govspeak in Publishing API's database, removing the rendering work from request time

- [ADR](https://github.com/alphagov/publishing-api/blob/main/docs/arch/adr-011-cache-rendered-govspeak.md)
- [implementation PRs](https://github.com/alphagov/publishing-api/pulls?q=is%3Apr+PP-7454+)
- [Jira ticket](https://gov-uk.atlassian.net/browse/PP-7454)

So we'll probably need to be quite confident that this change won't push overall request times beyond acceptable thresholds

nacnudus:
A few unfinished thoughts on this perennial compromise of what to cache, and when.

Cache pros:
- performance
- uniformity (e.g. same Govspeak version for all content items)
- convenience to general users (e.g. Search, Chat) who don't need to customise the render

Cache cons:
- maintenance of the cache is a pain (e.g. republishing content items when a gem changes)
- uniformity (can't customise the render)

Cache position:
- Between the publishing apps and the frontend apps (status quo).
- After the frontend apps. This is Fastly's cache, which isn't really accessible the same way as the Content API.

The Content API currently caches rendered Govspeak, and expanded links. The GraphQL implementation wouldn't have cached expanded links. This RFC wouldn't cache rendered Govspeak.                                                                                                                                                                         

Some users of the Content API have expressed a need for content in both a more semantic form (even more structured than Govspeak) and a more finished form (what the Frontend apps produce, which is cached by Fastly).

The Content API could serve both Govspeak and HTML together - there is already provision for them in the [schemas](https://github.com/alphagov/publishing-api/blob/88967f00046788359d4c00cfe3818ca4b08855c4/content_schemas/formats/shared/definitions/publishing_api_base.jsonnet#L139-L146). Frontend apps could then (dynamically) choose what to use.

Perhaps a new API could be developed for accessing the finished product produced by the frontend apps. That might make it possible to satisfy more of the people, more of the time.

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Problem
+
+GOV.UK publishing works in a three-step system. In step one editors create documents using (mainly) govspeak. In step two that raw govspeak is converted into an HTML representation, which is then passed on to the data stores ([publishing-api] and [content-store]). Finally the frontend apps request those HTML representations, which they use to compose the pages rendered to the users.
+
+Pros of the current situation:
+- Efficient: each document of raw govspeak only has to be converted to HTML once, at time of publishing
+- Status Quo: No extra work required
+
+Cons of the current situation:
```
andysellick, line 21: Another con to possibly add:

`Generated Govspeak markup is harder for frontend developers to change, so blanket rules for styling are applied rather than classes applied to specific elements, reducing the ability to reuse existing Design System styles and introducing inheritance problems that require more CSS rules to fix.`

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+- Single representation of blocks of edited text throughout the publishing system
+- We could pass markdown to LLMs, possibly resulting in better answers
+- We could perform A/B tests on components rendered in govspeak
+
+## Proposal
+
+We move Govspeak rendering out of the publishing stack and into the frontend stack.
+
+Changes that would be required:
+- The govspeak component in the [govuk_publishing_components] gem would need to be enhanced to allow it to accept raw govspeak and convert it to html in-app.
```
andysellick, line 58: Link suggestion: https://components.publishing.service.gov.uk/component-guide/govspeak

rfc-194-render-govspeak-in-frontend.md
leimdorfer: Agree - I think this might become one of the core business reasons for removing the rendering from the content management tools to further downstream. 
Embedded content (e.g. a map or a chart, etc) from authoring tool X might only exist as a reference in CMS Y.

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+- We could pass markdown to the mobile app through content api
+- Publishing apps do not have to know about markdown->html conversion
+- It might be easier to handle embed codes in frontend apps, improving the flexibility of
+  various document types (for example, a document could have an embed tag for "postcode
+  finder goes here", allowing editors more latitude in what they put before or after
+  the finder).
+- Single representation of blocks of edited text throughout the publishing system
+- We could pass markdown to LLMs, possibly resulting in better answers
+- We could perform A/B tests on components rendered in govspeak
+
```
leimdorfer, line 52: Along with the A/B testing there are other potential benefits to offering different parts of the page to different consumers. It opens some possible areas for personalisation within the page and/or associating meta-data to content at a sub-page level (potentially more a benefit of opting for a syntax tree over raw GovSpeak, but better easier than the current rendered html, I believe)

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+status: proposed
+implementation: proposed
+status_last_reviewed:
+---
+
+# Render Govspeak in Frontend
+
+## Summary
+
+It's been almost a decade since we made the decision (in [RFC-48]) to render [govspeak] in the publishing apps. Recently the concept of rendering it in frontend has come up, and in this RFC we'll look at whether it might make more sense to render govspeak in the frontend apps instead.
```
ChrisBAshton, line 11: Is it worth making reference to the new [Tech Strategy slides](https://beisgov.sharepoint.com/:p:/r/sites/GOVUKTechnologyCommunity-TechnologyCommunity/_layouts/15/doc2.aspx?sourcedoc=%7BA20B1F6F-8F10-4E4C-8B8F-CD9E9BFE0E32%7D) here? It shows that other people have done some deep thinking on this too and arrived at a similar conclusion:

> Data Model Coupling: Content is stored as pre-formatted HTML inside JSON blobs instead of raw text, directly blocking seamless ingestion by the app and AI services.
^ p32

> Stop the practice of compiling and storing content as pre-formatted HTML code inside the databases. Instead, transition to a semantic model where content is stored purely as raw text, Markdown, or structured "content blocks," leaving the layout and styling to be handled entirely by the consuming frontend applications.
^ p68

rfc-194-render-govspeak-in-frontend.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Problem
+
+GOV.UK publishing works in a three-step system. In step one editors create documents using (mainly) govspeak. In step two that raw govspeak is converted into an HTML representation, which is then passed on to the data stores ([publishing-api] and [content-store]). Finally the frontend apps request those HTML representations, which they use to compose the pages rendered to the users.
```
ChrisBAshton, line 15: Might be useful to lift and shift some of the background info from https://gov-uk.atlassian.net/wiki/spaces/GOVUK/pages/5689999367/Proposal+to+send+array+of+content+instead+of+details.body#Recap%3A-how-Govspeak-is-rendered into this RFC.

PR #195, opened 2026-07-20
RFC-195: Remove Link Set Links from Publishing API and publishing applications

brucebolt:
[💅 Rendered version](https://github.com/alphagov/govuk-rfcs/blob/rfc-195/rfc-195-remove-link-set-links.md)

Deadline for comments: 15th September 2026.

### DIFF
--- a/rfc-195-remove-link-set-links.md
+++ b/rfc-195-remove-link-set-links.md
@@ -1,0 +1,64 @@
+---
+status: proposed
+implementation: proposed
+status_last_reviewed: 2026-07-20
+---
+
+# Remove link set links from Publishing API and publishing applications
+
+## Summary
+
+Publishing API sits at the heart of the publishing infrastucture. It allows documents to link to others, to enable related links or other similar content to be displayed alongside each other on the public website.
+
+The documents are stored as editions, so it is possible to have both a live edition and a draft edition (which is used to preview the document). Publishing apps send the content as a `PUT content` request, which makes the content appear on the draft site. They then make a `POST publish` request to make that content live.
+
+But there are two ways of adding links:
+
+- One of these are called "edition links", and these work like the rest of the content and are sent in the same payload, so they only go live when the publisher clicks publish.
+- The other are called "link set links". These are associated with the document, not the edition, so go live as soon as a publisher adds them to the document (not when they publish the draft).
+
+## Problem
+
+The existence of two types of links has multiple consequences:
+
+- It's a bad user experience: publishers think they are adding content to a draft document, but it appears live on the site immediately.  This has partly been mitigated by changing the UI in the publishing apps to make this clear, but it's far from ideal.
+- It makes the link expansion code in Publishing API very complicated and difficult to understand. Developers are not keen on making changes because of the complexity. This became a prominent issue when we attempted to implement a GraphQL API, as we needed to support both types of link, again, making this code very complicated.
+- The difference in features between the two types of links makes onboarding difficult as the difference is hard to conceptualise, for both developers and other roles.
+- Edition links were only partly implemented, probably because of the complexity of the code. They don't currently support multi-level links, so you can't represent things like the taxonomy as edition links. This caused issues when the Whitehall team wanted to do that, as they needed to find a way of working around this limitation. See [RFC-188](https://github.com/alphagov/govuk-rfcs/pull/188).
+
+This has been recorded as [tech debt](https://gov-uk.atlassian.net/browse/PTD-164).
+
+Link set links are used by the following publishing applications:
+
+- Collections Publisher
+- Content Tagger
+- HMRC Manuals API
+- Mainstream Publisher
+- Manuals Publisher
+- Search API (limited only to a rake task that publishes static content related to search)
+- Service Manual Publisher
+- Specialist Publisher
+- Travel Advice Publisher
+- Whitehall
+
+## Proposal
+
+We are proposing to remove the usage of link set links from all publishing applications. This will then allow the code for link set links to be removed from Publishing API.
+
+## Work required to support this RFC
+
+### Remove link set links from all publishing applications
+
+TODO: pending a discovery
+
+### Remove link set links from all Publishing API content schemas
+
+TODO: pending a discovery
+
+### Remove endpoints to add link set links to content in Publishing API
+
+TODO: pending a discovery
+
+### Remove link set links from link expansion code in Publishing API
+
+TODO: pending a discovery
--- a/rfc-195-remove-link-set-links.md
+++ b/rfc-195-remove-link-set-links.md
@@ -1,0 +1,322 @@
+---
+status: proposed
+implementation: proposed
+status_last_reviewed: 2026-09-01
+---
+
+# Remove Link Set Links from Publishing API and publishing applications
+
+## Summary
+
+[Publishing API](https://github.com/alphagov/publishing-api) sits at the heart of the publishing infrastructure. It allows documents to link to others, to enable related links or other similar content to be displayed alongside each other on the public website.
+
+Each document can have multiple editions, so it is possible to have both a live edition and a draft edition (which is used to preview the document). Publishing apps send the content as a `PUT content` request, which makes the content appear on the draft site. They then make a `POST publish` request to make that content live.
+
+All links target a Content ID (i.e. all editions and all translations of a document) but they vary in what they link from. There are two different types of link, with a different source:
+
+- One of these is called "Edition Links", and these work like the rest of the content and are sent in the same payload, so they only go live when the publisher clicks publish. Edition Links are only associated with the locale of the edition.
+- The other is called "Link Set Links". These are associated with the Content ID, not the edition, so go live as soon as a publisher adds them (not when they publish the draft). Since these are associated with a Content ID instead of a document, a Link Set Link applies to all locales of the document.
+
+Further details of the two link types are given in [Publishing API's documentation](https://github.com/alphagov/publishing-api/blob/main/docs/link-expansion.md#link-lifecycles).
+
+## Problem
+
+The existence of two types of links has multiple consequences:
+
+- It's a bad user experience: publishers think they are adding content to a draft document, but it appears live on the site immediately. This has partly been mitigated by changing the UI in the publishing apps to make this clear, but it's far from ideal.
+- Some links cannot be added in publishing applications and instead have to be tagged in Content Tagger. This means users have to switch between two applications to set links.
+- It makes the link expansion code in Publishing API very complicated and difficult to understand. Developers are not keen on making changes because of the complexity. This became a prominent issue when we attempted to implement a GraphQL API, as we needed to support both types of link, again, making this code very complicated.
+- The difference in features between the two types of links makes onboarding difficult as the difference is hard to conceptualise, for both developers and other roles.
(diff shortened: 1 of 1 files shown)


leenagupte:
This is really interesting and I love the idea of reducing complexity and links having a common meaning and behaviour.

Some of the things that have come up in discussions are:
1. Bulk tagging in Content Tagger. This hasn't been used in earnest in over 4 years, so I don't know how much of a use case there still is.
2. Smart-answers. The content is hard-coded in the application, so each smart-answer is tagged in Content Tagger. Smart-answers only have "editions" in that publishing-api records each new publish as a new edition, but the application doesn't have sight of those.

Those issues should be referenced at least as a consideration in this RFC.

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,322 @@
...
+
+This analysis (including details of the exact links) has been added to [a document](https://beisgov.sharepoint.com/:x:/r/sites/GOVUK-Pub/Shared%20Documents/Content%20APIs/Workstream_%20Removing%20Link%20Set%20Links%20from%20Publishing%20API/2026-08-17%20Analysis%20of%20links%20used%20in%20schemas.xlsx?d=w505cd6cc06f54c0b9ca7d543a6861961&csf=1&web=1&e=4nPBda).
+
+The Content APIs team will provide a migration guide to teams who own publishing applications, explaining how to migrate away from Link Set Links to Edition Links.
+
+This will include information on the different requests to Publishing API that need to be made and how the user interface may need to be modified to support editions of the parent document.
+
+This will include details and examples on how to:
+
+- Modify the schema to support Edition Links.
```
ChrisBAshton: Can this be done by the Publishing API team? Taking the existing links config as a starting point ([example](https://github.com/alphagov/publishing-api/blob/d007759af2c53ae2411abd38ddcbfdca0abeafe0/content_schemas/formats/statistics_announcement.jsonnet#L61-L64)) and creating an (optional) `edition_links` equivalent, to reduce frictions for teams who want to help get this RFC over the line?

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,322 @@
...
+- Change the publishing applications to send the links as Edition Links, instead of Link Set Links.
```
ChrisBAshton, line 102: Agreed this feels like something for the product teams to do.

For Whitehall, 'legacy' content types such as Statistics Announcements seem to [only send link set links](https://github.com/alphagov/whitehall/blob/822d5047612ae795e0ccbe50683354ea9a288367/app/presenters/publishing_api/statistics_announcement_presenter.rb#L29), but changing it to send edition links is a tiny tweak to the presenter. And then a bulk republish of all the content. Both of which feel like the responsibility of the product teams.

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,322 @@
...
+- Update the publishing application user interface to support multiple editions of a document.
```
ChrisBAshton: This isn't strictly needed, is it? Even if something isn't editionable yet, we can still send edition links in the payload, it's just that it's always one, live edition.

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,322 @@
...
+- Remove the Link Set Links from Publishing API's database.
+- Modify the schema to remove the Link Set Links.
+- Make requests to Publishing API to obtain details of linkable documents that are published by other publishing apps.
```
ChrisBAshton: Not sure I understand what this last bulletpoint is about. Why would I need to obtain details of linkable documents published by other apps...?

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,322 @@
...
+
+The primary motivation behind Content Tagger appears to be to provide a tagging interface for document types whose publishing applications do not have this functionality.
+
+This RFC therefore proposes Content Tagger becoming a taxonomy-management application by retiring Content Tagger's tagging functionality. The publishing applications would get native taxonomy editing (which is already the case in Whitehall). This moves the editorial workflow closer to the content being tagged.
+
+This would create consistency across all publishing applications, by ensuring content is tagged solely in the publishing application, and remove the confusion caused by some publishing applications having their own interface.
+
+As with the other changes needed in publishing applications, the Content APIs team will provide a migration guide to teams who own publishing applications, explaining how to migrate away from Content Tagger and Link Set Links to Edition Links in publishing applications.
```
ChrisBAshton, line 133: I think I'd like to see some more detail here - I don't think there are that many things that need migrating out of Content Tagger? (Things like linked 'mainstream browse pages', etc). It would be good to have a table of the current features in Content Tagger and the proposed new home for them (e.g. is 'mainstream browse pages' now an edition link in Whitehall, and if so, where should Whitehall pull the list of options from?)

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Make publishing applications the authoritative source for their documents' links
+
+Content Tagger manages tags which are represented as Link Set Links of a limited number of link types. These are updated via the "patch links" endpoint in Publishing API. Content Tagger is not aware of editions, and does not publish the content of these documents.
+
+The primary motivation behind Content Tagger appears to be to provide a tagging interface for document types whose publishing applications do not have this functionality.
+
+This RFC therefore proposes Content Tagger becoming a taxonomy-management application by retiring Content Tagger's tagging functionality. The publishing applications would get native taxonomy editing (which is already the case in Whitehall). This moves the editorial workflow closer to the content being tagged.
```
ChrisBAshton: Are you also proposing that publishing apps have their own copy of which taxons their content are tagged to? (Rather than pulling from Publishing API at request time, like we do in Whitehall)?

If so, the problems I ran into in RFC-188 is what happens if the taxonomy tree changes. Content Tagger allows deleting a tag and retagging everything to an ancestor, etc. Keeping local data in sync with the taxonomy tree comes with some risks (worst case scenario being an accidental deletion of all local tag data if we receive a garbled response from upstream!).

ChrisBAshton (dismissed):
Very excited to see this work being prioritised 🎉 nice one.

rfc-195-remove-link-set-links.md
brucebolt: That seems reasonable.  I'll add it to a new paragraph in the same section.

rfc-195-remove-link-set-links.md
brucebolt: You're absolutely right, this is a "nice to have" rather than being essential.  I'll remove this line completely, as it is not essential as part of the migration.

rfc-195-remove-link-set-links.md
brucebolt: This would be so publishing apps can offer publishers a choice of documents to link to, e.g. if a mainstream document were to be tagged to a World Location, then Mainstream Publisher would need to query Publishing API to get a list of all World Locations.

I'll add this example to the document.

rfc-195-remove-link-set-links.md
brucebolt: I've determined which Link Set Links are currently editable in Content Tagger and added that to the RFC, along with a brief sentence about where the publishing apps will get the list of documents from.

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,330 @@
...
+Even if we decide not to carry out the remainder of this RFC, this work should be carried out regardless, since it will have a high impact on other teams' ability to use Edition Links for tagging to the taxonomy.
+
+### Remove Link Set Links from all publishing applications and content schemas
+
+At the time of writing, there are 83 content schemas. This can be broken down based on existing link usage:
+
+- No links: 5
+- Edition Links only: 1
+- Link Set Links only: 0
+- Mixture of Edition Links and Link Set Links: 77
```
graycodes, line 92: Do we know what number of links we have of either kind?

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,330 @@
...
+
+This analysis (including details of the exact links) has been added to [a document](https://beisgov.sharepoint.com/:x:/r/sites/GOVUK-Pub/Shared%20Documents/Content%20APIs/Workstream_%20Removing%20Link%20Set%20Links%20from%20Publishing%20API/2026-08-17%20Analysis%20of%20links%20used%20in%20schemas.xlsx?d=w505cd6cc06f54c0b9ca7d543a6861961&csf=1&web=1&e=4nPBda).
+
+The Content APIs team will provide a migration guide to teams who own publishing applications, explaining how to migrate away from Link Set Links to Edition Links.
```
graycodes, line 96: Should we suggest a date the feature may be retired or some kind of soft deadline?

rfc-195-remove-link-set-links.md
```diff
@@ -0,0 +1,330 @@
...
+
+This means it will not be possible to republish documents from the publishing application to migrate these from Link Set Links to Edition Links.
+
+The only current usage of this link type across GOV.UK is [in `email-alert-api`](https://github.com/alphagov/email-alert-api/blob/132781fa5d801d00f8afe0aefe7ab02b827f9d88/lib/email_alert_criteria.rb#L74). This means users subscribed to alerts on a policy area will not be receiving updates for new documents, as it is no longer possible to tag documents to a policy area. These users will only receive updates for changes to existing documents.
+
+There are currently 25 users subscribed to email alerts for `policy_areas` links, and there have been no matched content changes within the current retention period (one year).
+
+It is therefore proposed that as part of this work, we would contact subscribers of policy areas to inform them this type of subscription has been deprecated. After a short grace period, we will remove all the `policy_areas` links from Publishing API's database and the associated code from Email Alert API.
```
AgaDufrat: _Sorry to be a bit late on this._ No objections to the proposal just a clarifying question: Will this work be completed by the Content API team with review from Navigation team that owns email-alerts-api?

brucebolt: Yes, that's my intention (unless the Navigation team would like to be more involved).  I'll update the text to make this clearer.

rfc-195-remove-link-set-links.md
brucebolt: That's a good idea.  Something I'll have to talk to the team and leads about.

rfc-195-remove-link-set-links.md
brucebolt: Publishing API has a total of 64,318,663 links.

Of these, 59,151,710 are Edition Links and 5,166,953 are Link Set Links.

Breaking down the Edition Links further, 2,166,274 of them are associated with a published edition.  And for Link Set Links, 4,117,109 are associated with a document that has a published edition.

It is the links associated with published editions that will need updating.  It should be safe to discard to Link Set Links that are not associated with a document that has a published edition.

----

For future reference, these are the queries that were used:

All links:
```rb
Link.count
```

Edition Links:
```rb
Link.where(link_set_content_id: nil).count
```

Link Set Links:
```rb
Link.where(edition_id: nil).count
```

Total Edition Links on published editions:
```rb
Link.joins(:edition).where("edition.state": "published").count
```

Total Link Set Links on a document with a published edition:
```rb
Link
  .link_set_links
  .joins(link_set: { documents: :editions })
  .where(editions: { state: "published" })
  .distinct
  .count
```

rfc-195-remove-link-set-links.md
brucebolt: Taxonomy is the part of these changes that I'm least confident about.

If we only had one publishing app, and that app were to also manage the taxonomy, then amending the taxonomy could result in documents being automatically retagged by the app itself.  This will be a lot more difficult with the tagging being done within the publishing apps, but the taxonomy being managed elsewhere.

I'll have a bit more of a think about this, to see what we could do when the taxonomy is updated.

PR #197, opened 2026-09-02
RFC-197: New convention: array of content instead of single 'body'

ChrisBAshton:
[Rendered version](https://github.com/alphagov/govuk-rfcs/blob/rfc-196/rfc-197-array-of-content-in-details.md) 💅

Deadline for comments: **Thursday 17th September**

Jira: https://gov-uk.atlassian.net/browse/WHIT-3923

### DIFF
--- a/rfc-197-array-of-content-in-details.md
+++ b/rfc-197-array-of-content-in-details.md
@@ -1,0 +1,238 @@
+---
+status: proposed
+implementation: proposed
+status_last_reviewed:
+---
+
+# New convention: array of content instead of single 'body' 
+
+## Summary
+
+To date, by convention, the main content of a document is represented as a single `body` property in the `details` hash of its content item. This convention is codified end-to-end, from the publishing app, to Publishing API, to Content Store, to the frontend rendering app, as well as in other downstream consumers such as Search API.
+
+This RFC proposes gradually moving towards a new convention of an 'array' of content as a more flexible means of expressing content and its metadata, to respond to new and emerging requirements.
+
+## Problem
+
+There are emerging requirements that cannot be represented cleanly using a single `details.body` value.
+
+### Multiple sections of content
+
+The new 'mini site' format needs to be able to position a "Featured" section at an arbitrary point in the page content.
+
+With the current representation, the body is a single compiled Govspeak string:
+
+```json
+{
+  "details": {
+    "body": "<div class=\"govspeak\"><p>Some content</p></div>"
+  }
+}
+```
+
+There is no provision for embedding non-body content inside that string of HTML. Under the current architecture, we would need to introduce some sort of placeholder in the Govspeak and have Frontend parse the body to identify and swap out the placeholder.
+
+This is recognised as a problem in the Technology Strategy Playback ([Sharepoint][tech_strategy_sharepoint], [Google][tech_strategy_google]), page 32:
+
+> Rigid Template Layouts: Restrictive publishing formats limit modern design elements (like maps and video), which driving some departments to build disconnected, off-platform sites.
+
+[tech_strategy_sharepoint]: https://beisgov.sharepoint.com/:p:/r/sites/GOVUKTechnologyCommunity-TechnologyCommunity/_layouts/15/doc2.aspx?sourcedoc=%7BA20B1F6F-8F10-4E4C-8B8F-CD9E9BFE0E32%7D
+[tech_strategy_google]: https://docs.google.com/presentation/d/1PkG20xeW1dxeYlgQsW3SPaZLlCceQKVr0RaHsufe36M/edit
+
+### Metadata attached to sections of content
+
+There are also emerging requirements in the Classifications and Systems Metadata space. Whilst the exact use cases are still being finalised, there is a general desire to be able to tag content at a more granular level than the page itself. Use-cases for associating metadata with chunks of content could include:
+
+* designating content as associated with a taxonomy
+* designating content intended for an app
+* designating content intended for AI consumption
+* designating content intended for users in particular nations of the UK
+
+Some [native Govspeak support already exists for devolved content](https://github.com/alphagov/govspeak#devolved-content), but this still requires the entire body to be retrieved and parsed. It's also not a particularly sustainable approach to have to add new Govspeak rules for every thing we might want to be able to tag.
+
+This is also recognised as a problem and recommendation in the Technology Strategy Playback ([Sharepoint][tech_strategy_sharepoint], [Google][tech_strategy_google]), page 64:
+
+> Restructure GOV.UK content and metadata specifically so that it is machine-readable and consumable by third-party Large Language Models (LLMs) and automated agents, rather than just optimising for human web consumption.
+
+## Proposal
+
+Instead of a single `details.body` string of HTML, we will iterate towards `details.content[]`, which will contain an array of objects adhering to a common interface. Each object will include the following properties:
+
+- `type` - the type of the object
+- `value` - the value of the object
+- `metadata` (optional - can be omitted) - a hash of arbitrary metadata associated with the object
+- Additional properties could be allowed in future on a per-object-type basis.
+
+This data structure allows us to solve the content-granularity and content-tagging problems outlined earlier.
+
+### Example
+
+Here is how a content item currently looks:
+
+```json
+{
+  "details": {
+    "body": "<div class=\"govspeak\"><p>Some content</p></div>",
+    "change_history": [
+      {
+        "public_timestamp": "2026-08-10T08:41:00.000+01:00",
+        "note": "First published."
+      }
+    ]
+  }
+}
+```
+
+As a starting point, all we're looking to do is swap the above for this - note that the other fields within `details` (i.e. `change_history`) are unchanged:
+
+```json
+{
+  "details": {
+    "content": [
+      {
+        "type": "html",
+        "value": "<div class=\"govspeak\"><p>Some content</p></div>"
+      }
+    ],
+    "change_history": [
+      {
+        "public_timestamp": "2026-08-10T08:41:00.000+01:00",
+        "note": "First published."
+      }
+    ]
+  }
+}
+```
+
+Going forward, we're free to work towards new object types and any arbitrary metadata, to accommodate future requirements. The following example is out of scope for this RFC, but is included as an indication of what could be possible. As ever, it is up to downstream consumers (such as Frontend) to decide what to do with this information:
+
+```json
+{
+  "details": {
+    "content": [
+      {
+        "type": "html",
+        "value": "<div class=\"govspeak\"><p>Some content</p></div>",
+        "metadata": {
+          "app_friendly": true,
+          "devolved_nations": [
+            { "nation": "wales", "alternative_uri": "https://gov.uk/foo.cy" }
+          ]
+        }
+      },
+      {
+        "type": "featured_section",
+        "ordered_featured_documents": [
+          {
+            "document_type": "Policy paper",
+            "href": "/government/publications/youth-matte..."
+          },
+          ...
+        ]
+      },
+      {
+        "type": "govspeak",
+        "value": "## Heading\n[Link](https://gov.uk/bar)",
+      },
+      {
+        "type": "html/h1",
+        "value": "HEADING LEVEL 1 GOES HERE (just an example)",
+      },
+    ],
+    "change_history": [
+      {
+        "public_timestamp": "2026-08-10T08:41:00.000+01:00",
+        "note": "First published."
+      }
+    ]
+  }
+}
+```
+
+Note the multiple objects in the `details.content` array, the different `type`s and the common use of `value` (where appropriate).
+
+### Implementation
+
+We're going to begin with content types that have [migrated to StandardEdition](https://docs.publishing.service.gov.uk/repos/whitehall/migrating_to_standard_edition.html). At time of writing, these are: Topical Events, News Articles, Case Studies and History Pages. The long term aim in Whitehall is to migrate all content types to StandardEdition, so eventually all content types originating from Whitehall will be represented as `details.content[]`.
+
+Working backwards, we need to build support into the downstream consumers of content.
+
+1. In Frontend, make a change to [ContentItem](https://github.com/alphagov/frontend/blob/d3073a4fd70e304d47aae413a160cbd42ae959eb/app/models/content_item.rb#L14) so that the `body` method grabs the first appropriate item (`type: "html"`) from `details.content` if it exists, otherwise falling back to `details.body`.
+1. In search-api, make a change to [indexable_content_parts](https://github.com/alphagov/search-api/blob/87a43870dc63f1e08d72092cd0eb389a8aaf5bb0/lib/govuk_index/presenters/indexable_content_presenter.rb#L39-L44) to look for `details.content` and to retrieve any objects which have a `value`.
+1. In search-api-v2, make a change to [INDEXABLE_CONTENT_VALUES_JSON_PATHS](https://github.com/alphagov/search-api-v2/blob/e13de24a3b825b7a2c805c011d1693cec435d058/app/models/concerns/publishing_api/content.rb#L15) to include `$.details.content[*]['value']`.
+
+Then in the publishing stack:
+
+1. Update Publishing API schemas for the StandardEdition formats, to accept either a `details.body` string or a `details.content` array of objects.
+1. Make a change to Whitehall's [StandardEditionPresenter](https://github.com/alphagov/whitehall/blob/7a505b1b575e24676405d828e22ad8f32b41f292/app/presenters/publishing_api/standard_edition_presenter.rb#L42-L43) to send `details.content` instead of `details.body`.
+1. Republish all StandardEdition content.
+1. Update Publishing API schemas again to remove support for `details.body` from these formats.
+
+At this point we'd now have the foundations in place, across the stack, to allow for arbitrary placement of additional sections in the page, and arbitrary tagging of said sections.
+
+In terms of governance, we also propose:
+
+1. A shared document in the Developer Docs, describing the different object types that can appear in `details.content[]`. That way, all downstream consumers have a reference they can build against.
+1. Pact tests could be considered further down the line, when more object types are created, but are overkill at this stage.
+
+### Beyond
+
+Where we iterate on the foundation above is out of scope for this RFC. But as an indicator of the kind of value this change can unlock, here are some thoughts below.
+
+On a per-document-type basis, the Whitehall and Patterns & Pages teams will work together to enable dynamic support for unlimited elements in the `details.content` array.
+
+In the short term, we may begin with the new 'mini site' content type, with a view to sending two bodies of content and a Featured section in the middle:
+
+```json
+{
+  "details": {
+    "content": [
+      {
+        "type": "html",
+        "value": "<div class=\"govspeak\"><p>Content above the featured section</p></div>",
+      },
+      {
+        "type": "featured_section",
+        "ordered_featured_documents": [
+          {
+            "document_type": "Policy paper",
+            "href": "/government/publications/youth-matte..."
+          },
+          ...
+        ]
+      },
+      {
+        "type": "html",
+        "value": "<div class=\"govspeak\"><p>Content below the featured section</p></div>",
+      },
+    ],
+  }
+}
+```
+
+We might achieve this in Whitehall in the short term by:
+
+1. Continuing to have a single 'Body' textarea
+1. Defining some sort of `{{Featured}}` placeholder that the publisher can put into the Body
+1. At the point of saving, the StandardEditionPresenter can check for the presence of that placeholder and if it exists, split the Body into two, and inject the Featured section in between, in the `details.content` that is sent to Publishing API
+
+Longer term, one can imagine some sort of [WordPress Gutenberg](https://wordpress.org/gutenberg/)-like WYSIWYG interface where the publisher can create and rearrange structured content, allowing them to inject all manner of special content (such as maps, videos, etc) anywhere in the content body. That interface could also allow for arbitrary tagging of each piece of content.
+
+## Appendix
+
+### Naming
+
+`details.content` was so chosen because this is expected to encapsulate the main content of the page, replacing `details.body` - and there don't appear to be any existing presenters / content items where a `content` property already exists.
+
+`details.content[].type` was so chosen because it's a good name for the 'type' of the thing in the array. We did also consider `content_type`, but apart from the unnecessary repetition of the word 'content', that property name _does_ have an existing meaning in the publishing stack already:
+
+- Content can be sent as an array of hashes each with a `content_type` property, with a value of `text/html` or `text/govspeak`. See [example of sending 'summary' of type Govspeak from Specialist Publisher](https://github.com/alphagov/specialist-publisher/blob/c8bb9d4902e8dd341589ce96018b46a32cccfb91/app/presenters/finder_content_item_presenter.rb#L64-L74)
+- If only an element of `content_type: "text/govspeak"` is sent, [Publishing API renders the Govspeak](https://github.com/alphagov/publishing-api/blob/70c62d85f71205b22a1d1bd467e4330d84b3fbac/lib/govspeak_details_renderer.rb#L27-L35) and saves an additional element to the hash of type `text/html`.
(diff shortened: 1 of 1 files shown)


leenagupte:
I've had a quick read through this and it sounds like a good idea, and a good place to start thinking about breaking up content into sub-page pieces.

I don't have any objection to the suggested naming. It all sounds sensible.

rfc-197-array-of-content-in-details.md
```diff
@@ -0,0 +1,238 @@
...
+
+Some [native Govspeak support already exists for devolved content](https://github.com/alphagov/govspeak#devolved-content), but this still requires the entire body to be retrieved and parsed. It's also not a particularly sustainable approach to have to add new Govspeak rules for every thing we might want to be able to tag.
+
+This is also recognised as a problem and recommendation in the Technology Strategy Playback ([Sharepoint][tech_strategy_sharepoint], [Google][tech_strategy_google]), page 64:
+
+> Restructure GOV.UK content and metadata specifically so that it is machine-readable and consumable by third-party Large Language Models (LLMs) and automated agents, rather than just optimising for human web consumption.
+
+## Proposal
+
+Instead of a single `details.body` string of HTML, we will iterate towards `details.content[]`, which will contain an array of objects adhering to a common interface. Each object will include the following properties:
```
brucebolt, line 59: I know it's explained at the end of the RFC, but is it worth mentioning Publishing API's [ADR-003](https://github.com/alphagov/publishing-api/blob/12a6091765cc90cb499b9b04d810729feb0e9d21/docs/arch/adr-003-representation-for-multiple-content-types.md), which allows the `body` to already contain an array of content types earlier on?

rfc-197-array-of-content-in-details.md
```diff
@@ -0,0 +1,238 @@
...
+}
+```
+
+Note the multiple objects in the `details.content` array, the different `type`s and the common use of `value` (where appropriate).
+
+### Implementation
+
+We're going to begin with content types that have [migrated to StandardEdition](https://docs.publishing.service.gov.uk/repos/whitehall/migrating_to_standard_edition.html). At time of writing, these are: Topical Events, News Articles, Case Studies and History Pages. The long term aim in Whitehall is to migrate all content types to StandardEdition, so eventually all content types originating from Whitehall will be represented as `details.content[]`.
+
+Working backwards, we need to build support into the downstream consumers of content.
```
brucebolt, line 158: I think [content-data-api also consumes this field](https://github.com/alphagov/content-data-api/blob/ad910b9acb047e85fd23b5d747d13e854a325e1f/app/domain/etl/edition/content/parsers/body_content.rb#L4), so would need to be updated.

rfc-197-array-of-content-in-details.md
```diff
@@ -0,0 +1,238 @@
...
+1. Update Publishing API schemas for the StandardEdition formats, to accept either a `details.body` string or a `details.content` array of objects.
+1. Make a change to Whitehall's [StandardEditionPresenter](https://github.com/alphagov/whitehall/blob/7a505b1b575e24676405d828e22ad8f32b41f292/app/presenters/publishing_api/standard_edition_presenter.rb#L42-L43) to send `details.content` instead of `details.body`.
+1. Republish all StandardEdition content.
+1. Update Publishing API schemas again to remove support for `details.body` from these formats.
+
+At this point we'd now have the foundations in place, across the stack, to allow for arbitrary placement of additional sections in the page, and arbitrary tagging of said sections.
+
+In terms of governance, we also propose:
+
+1. A shared document in the Developer Docs, describing the different object types that can appear in `details.content[]`. That way, all downstream consumers have a reference they can build against.
```
brucebolt, line 175: This could be best placed in the Content API documentation, as it may be of interest to external consumers of the API.

We [currently](https://content-api.publishing.service.gov.uk/reference.html#schema-contentitem) don't go into any detail of what can be included in `details`, but I'd hope this can be expanded when our tech writer starts soon.

brucebolt (commented):
This proposal looks good to me.

From a Publishing API perspective the implementation detail seems to cover everything I can think of.

KludgeKML:
Can I check that we're assuming all ordering will be set in Whitehall and the contents array will be ordered in the expected rendering order? (ie if there are 5 content elements in the array and 3 of them are relevant to Frontend, then frontend should render them in the page in the same order they appear in the array?)

rfc-197-array-of-content-in-details.md
```diff
@@ -0,0 +1,238 @@
...
+
+This RFC proposes gradually moving towards a new convention of an 'array' of content as a more flexible means of expressing content and its metadata, to respond to new and emerging requirements.
+
+## Problem
+
+There are emerging requirements that cannot be represented cleanly using a single `details.body` value.
+
+### Multiple sections of content
+
+The new 'mini site' format needs to be able to position a "Featured" section at an arbitrary point in the page content.
```
minglis, line 21: In terms of an MVP version of this, we could in theory hide the array content by processing it into a single body element as it is now on saving to the content store (for example), minimising change in front end for for now. Is this requirement for the "featured" section driving the frontend changes as part of the initial implementation?

KludgeKML: From Frontend's perspective it's a very minor change to support the MVP of handling the array. Honestly I'd rather we did the scaffolding in Frontend than Whitehall, since there's no reasonable way around having to briefly handle both monobody and array in the frontend.

rfc-197-array-of-content-in-details.md
```diff
@@ -0,0 +1,238 @@
...
+        "type": "html",
+        "value": "<div class=\"govspeak\"><p>Some content</p></div>",
+        "metadata": {
+          "app_friendly": true,
+          "devolved_nations": [
+            { "nation": "wales", "alternative_uri": "https://gov.uk/foo.cy" }
+          ]
+        }
+      },
+      {
```
edavey, line 125: I've spent a bit of time thinking through what this means for reusable content (in the form of content blocks owned by Content Block Manager). I think the answer at the moment is "not too much". Content blocks will continue to have been rendered by this point and their metadata will continue to be exposed in their HTML "wrapper" element. I do think there will be a distinction between:

- larger blocks which may naturally form an element in `details.content[]` and
- small inline blocks (wrapped in a `span`) which sit within a larger element in `details.content[]`

I've worked through a couple of examples:

### 1. Some content blocks will be sections

For example, the block of content for the CMA Press Office "contact" details:

```
For media enquiries, contact the CMA press office.

Email

- [press@cma.gov.uk](mailto:press@cma.gov.uk)

Telephone numbers

- Phone: 020 3738 6460 
- Out of hours: 020 3738 6878

Opening hours

Monday to Friday, 9am to 5:30pm. On bank holidays, use our out of hours number.

Find out about call charges
```

as used on [Aldi and Lidl should be subject to...](https://www.integration.publishing.service.gov.uk/api/content/government/news/aldi-and-lidl-should-be-subject-to-the-same-rules-as-major-supermarkets)  may well be a distinct "section" or "element" within the new style `details.content[]` list e.g.:


```json
{  
  "type": "html",  
  "value": "<div class=\"content-block content-block--contact\" data-content-block=\"\" data-document-type=\"contact\" data-content-id=\"2d39a869-aa5f-4594-a317-52e1a8d7d4b4\" data-embed-code=\"{{embed:content_block_contact:contact-the-cma-press-office}}\"><div class=\"vcard\"> <div data-diff-key=\"description\" data-id=\"description\"> <p>For media enquiries, contact the CMA press office.</p> </div> <dl class=\"govuk-!-margin-0 govuk-!-padding-0\"> <div data-diff-key=\"email_addresses-email\" data-block-type=\"email_addresses\" data-block-id=\"email\"> <dt class=\"govuk-!-margin-top-0 govuk-!-margin-right-0 govuk-!-margin-bottom-4 govuk-!-margin-left-0 govuk-!-padding-0 govuk-!-font-size-19 govuk-!-font-weight-bold\"> Email </dt> <dd class=\"govuk-!-margin-top-0 govuk-!-margin-right-0 govuk-!-margin-bottom-6 govuk-!-margin-left-0\"> <ul class=\"govuk-list list-no-bullets govuk-!-margin-top-0 govuk-!-margin-right-0 govuk-!-margin-bottom-6 govuk-!-margin-left-0\"> <li> <a href=\"mailto:press@cma.gov.uk\" class=\"email\"> press@cma.gov.uk </a> </li> </ul> </dd> </div> <div data-diff-key=\"telephones-telephone\" data-block-type=\"telephones\" data-block-id=\"telephone\"> <dt class=\"govuk-!-margin-top-0 govuk-!-margin-right-0 govuk-!-margin-bottom-4 govuk-!-margin-left-0 govuk-!-padding-0 govuk-!-font-size-19 govuk-!-font-weight-bold\"> Telephone numbers </dt> <dd class=\"govuk-!-margin-top-0 govuk-!-margin-right-0 govuk-!-margin-bottom-6 govuk-!-margin-left-0\"> <ul class=\"govuk-list list-no-bullets govuk-!-margin-0\"> <li class=\"govuk-!-margin-bottom-1\"> <span>Phone: </span> <span class=\"tel\">020 3738 6460&nbsp;</span> </li> <li class=\"govuk-!-margin-bottom-1\"> <span>Out of hours: </span> <span class=\"tel\">020 3738 6878</span> </li> </ul> <h3 id=\"opening-hours\">Opening hours</h3> <p>Monday to Friday, 9am to 5:30pm. On <a href=\"https://www.gov.uk/bank-holidays\">bank holidays</a>, use our out of hours number.</p> <p> <a href=\"https://gov.uk/call-charges\"> Find out about call charges </a> </p> </dd> </div> </dl> </div> </div>",  
  "metadata": {  
    "content_block": true,  
    "content_id": "2d39a869-aa5f-4594-a317-52e1a8d7d4b4",  
    "embed_code": "{{embed:content_block_contact:contact-the-cma-press-office}}",  
    "url": "https://content-block-manager.publishing.service.gov.uk/api/blocks/%7B%7Bembed%3Acontent_block_contact%3contact-the-cma-press-office%7D%7D"  
  }  
}
```

In this case would the content block's metadata be "first class" in the list of content elements or would we rely solely on the metadata provided in the wrapper element of the html itself?

```html
<div 
  class="content-block content-block--contact" 
  data-document-type="contact" 
  data-content-id="2d39a869-aa5f-4594-a317-52e1a8d7d4b4" 
  data-embed-code="{{embed:content_block_contact:contact-the-cma-press-office}}"
>
```

Probably that's adequate and a consistent approach is best?


### 2. Some content blocks will be embedded within sections

Inline content blocks will be embedded within a "section" or "element" of the content array.  
  
e.g. the pension value on the [new-state-pension](https://www.integration.publishing.service.gov.uk/api/content/new-state-pension) document:  
  
> £241.30  
  
is shown within the paragraph:  
  
> The full rate of new State Pension is £241.30 a week. Your amount could be different depending on ... 

```html
<p>The full rate of new State Pension is <span class=\"content-block content-block--pension\" data-content-block=\"\" data-document-type=\"pension\" data-content-id=\"c9e6ad4d-ef33-4bba-bc48-37d04e94ee2a\" data-embed-code=\"{{embed:content_block_pension:new-state-pension/rates/full-new-state-pension-amount/amount}}\">£241.30</span> a week. Your amount could be different depending on:</p>
```

In this case the inline content block appears within a section and its metadata is provided on the wrapping element:

```html
<span 
  class="content-block content-block--pension" 
  data-document-type="pension" 
  data-content-id="c9e6ad4d-ef33-4bba-bc48-37d04e94ee2a\" 
  data-embed-code="{{embed:content_block_pension:new-state-pension/rates/full-new-state-pension-amount/amount}}"
>
```

nacnudus:
I don't fully understand the problem statements, given that the publishing stack already supports arbitrary HTML. The proposal seems to assume only frontend apps can assemble content into HTML.

- Would it be a problem for the publishing apps to render the content into HTML, and submit that to the Publishing API? [Publishing API ADR 004](https://github.com/alphagov/publishing-api/blob/12a6091765cc90cb499b9b04d810729feb0e9d21/docs/arch/adr-004-govspeak-and-embedded-content.md) says "frontends should do as little as possible at render time, to reduce latency".
- Is unrendered, JSON-structured content needed in the Publishing API? Possibly the Search API needs to know which content not to index (e.g. stuff not writting for humans).

On the whole I'm attracted to the idea of content being structured in JSON rather than in HTML, because it would be easier to work with in databases. In fact I incline further towards all content being a content block, and content items being composed only of content blocks. But that's probably an extreme position to take. I'm just not sure that this RFC justifies the middle ground yet.

rfc-197-array-of-content-in-details.md
edavey, line 59: Would it be helpful to propose clear terminology for this new concept? I've found myself unsure how to refer to the items in this `details.content[]` list. In the heading above you write about "sections of content". Is `section` the right term for these entities? If so then it might help to reinforce that usage wherever possible e.g.

> ...will contain an array of sections adhering to a common interface. Each section will include...

danhowarthgov:
Thanks for raising, @ChrisBAshton. Dan here, information architect in the Classification Systems and Metadata (CSMD) team. Think this approach definitely has merits. It's obviously a problem to pass through one big unstructured blob of content. We need to be able to break it down so we can use it more flexibly.

I know this RFC is agnostic of how the elements in the array will be created. But also note in the 'Beyond' section on line 219, you suggest this could enable a page-builder approach in Whitehall – like WordPress Gutenberg. I think this could create problems down the line, and constraints that will dictate how publishers behave.

Presumably each content block in the array will be associated with the same `content_id`? Let's take a detailed guide as an example. Say this one, '[Travel to Spain for Work](https://www.gov.uk/guidance/travel-to-spain-for-work)' – which is massive and a clear candidate for being modularised. If this was broken down into various content blocks, they will be sub-parts of the detailed guide. They'll still rely on the whole guide for the context in which they should be understood and used. It still supports a system where the page is king – which we have for now, of course, but presumably this is not what we want in the future.

The alternative might be self-contained component-level content types, that could be compiled into a detailed guide, but are understandable and usable in their own right. In the 'travel to spain for work' guide, there's lots of duplication and repetition throughout the guide and across the [29 other guides in the collection](https://www.gov.uk/government/collections/travelling-to-the-eu-switzerland-norway-iceland-or-liechtenstein-for-work) it belongs to – some prospective content blocks are carbon copies, others would need to support structured variations (e.g. switching out the name of the country). These could be be created and edited in one place, rather than repeated. With this approach, each component would have its own `content_id` and be able to be single-sourced  (or reused) in multiple places. I guess this is partly what the Content Reuse team have been trying to achieve with Content Block Manager – but with a very specific remit to date.

You mentioned the emerging requirements in CSMD (line 44), in what we're referring to as "sub-page content" and content blocks in our investigations. It's too early to say, but I suspect we'll veer towards self-contained component-level content types, over sub-parts. So less Gutenberg, and more like a DITA-based component content management system (CCMS). But that would be on a very small scale to start with.

Sounds like the Patterns and Pages team have been instrumental in proposing this approach, in relation to the Topical Events content type? But to what extent have departmental content teams been consulted, or the content designers in GOV.UK, or people working on content strategy? Because I'd hope this would be driven by a mix of user, publisher and business needs. 

At the very least, I think we ought to do some consequence scanning and work through the risks. I'm still fairly new to GDS and like this democratic approach to proposals – but it definitely excludes non-technical colleagues, who this will directly affect.

I've mentioned it to some of the GOV.UK content leads – and we all have reservations about the page-builder / WYSIWYG possibilities this could pave the way for. They've looked at visual editors before – both drag-and-drop style block-based editing, and WYSIWYG text editors. None have gone any further despite lots of time and effort. I know this is only a speculative future that's not in the scope of this RFC – but it's a logical next step. 

It does feel like this RFC is geared more towards UI flexibility (see case for placing 'featured' items on a page, lines 21 and 184), than improving our ability to manage and publish volumes of content from a single source.

Ultimately, this is about shaping the tools so they can support scalable content operations. So we can manage and publish content in a way that we can create once, publish everywhere. We need an approach to publishing content that's _mise en place_ for meeting user needs and business requirements, not _mise en page_.

rfc-197-array-of-content-in-details.md
danhowarthgov: Agreed. We're increasingly calling them 'content blocks'. So would `details.blockarray` or something like that work? It's much more specific about what it represents than `details.content`.

ChrisBAshton: I'm reluctant to call the array elements "content blocks" as that will cause confusion with:

1. Content Block Manager (which manages content that - for now at least - will be embedded _inside_ the Govspeak inside some of the elements in this array)
2. `StandardEdition.block_content` in Whitehall (the arbitrary hash of content we use to store arbitrary data, replacing the old-style Edition where every new piece of data required defining a dedicated property in the `editions` table). As it stands, `block_content[:body]` will be one of the elements in this array.

I'd be on board with calling the contents of the array "sections" of content, as opposed to objects (vague) or entities (which could imply separate persisted items). I'd avoided the term to not introduce confusion with RFC-182 ('flexible sections') and so as not to tie us down to thinking "every element in the array is a full width 'section' within the main body of the page", but in practice, the first few 'kinds' of section we can envisage _will_ be the full-width of the `<div>` ("Govspeak body", "Featured tiles", "Map"...) so it seems reasonable to lock in that vocabulary now and let it evolve later if need be.

As for the array itself, I think `details.content` works. The intention is for the main contents of the page to be incorporated into the array, and any other `details.*` properties might just act as page-level metadata or legacy data. `details.main_content` might be more expressive, if that's preferable (though arguably more clunky?).

rfc-197-array-of-content-in-details.md
ChrisBAshton: Agree with Keith: the benefit to rolling these minimal changes out across the _entire stack_ is that we can then move quite quickly when we do have a sudden requirement for extra sections or tagging.

rfc-197-array-of-content-in-details.md
ChrisBAshton: Thanks Bruce - good spot. Have added.

rfc-197-array-of-content-in-details.md
ChrisBAshton: Thanks @edavey - we spoke about this in person this week.

I agree - the 'embeddable' content blocks aren't affected here, they're just part of another element in the `details.content[]` array.

'Standalone' content blocks could indeed be their own 'sections' in the array, and their own metadata could be carried over from Content Block Manager as the high-level `metadata` property on the section. I did mention in our chat that there might be a need for Whitehall-level metadata to be merged into it too, so a standalone 'content block' section's `metadata` property could end up being a mixture of CBM metadata (say, `"embed_code": "{{embed:content_block_contact:contact-the-cma-press-office}}"`) and Whitehall metadata (say, `applicable_to_wales: false`). That all sounds doable.

rfc-197-array-of-content-in-details.md
danhowarthgov: Trouble with `details.content` is that 'content' is such a vague and abstract term. It's surely gonna cause problems down the road. An H1 is also content, but would that be passed through in the `details.content` field – or somewhere else?

Shame 'blocks' is out of contention, cos it seems to be industry standard for page-builders – e.g. in [Gutenberg](https://wordpress.org/gutenberg/).

Wagtail CMS, which a bunch of gov depts use – including, shortly, the GDaD framework – calls it '[StreamField](https://docs.wagtail.org/en/latest/topics/streamfield.html)'. Which is specific, if nothing else.

ChrisBAshton:
@nacnudus 

> I don't fully understand the problem statements, given that the publishing stack already supports arbitrary HTML. The proposal seems to assume only frontend apps can assemble content into HTML.
> 
> * Would it be a problem for the publishing apps to render the content into HTML, and submit that to the Publishing API? [Publishing API ADR 004](https://github.com/alphagov/publishing-api/blob/12a6091765cc90cb499b9b04d810729feb0e9d21/docs/arch/adr-004-govspeak-and-embedded-content.md) says "frontends should do as little as possible at render time, to reduce latency".
> * Is unrendered, JSON-structured content needed in the Publishing API? Possibly the Search API needs to know which content not to index (e.g. stuff not writing for humans).
> 
> On the whole I'm attracted to the idea of content being structured in JSON rather than in HTML, because it would be easier to work with in databases. In fact I incline further towards all content being a content block, and content items being composed only of content blocks. But that's probably an extreme position to take. I'm just not sure that this RFC justifies the middle ground yet.

Yes, Whitehall could compile all of the HTML ahead of time and send that to Publishing API, solving the "block of content, 'featured section', then another 'block of content'" problem. What that doesn't really solve is the tagging.

We've been asked to think about how we can apply metadata at a more granular level than page-level, with a view to multi-platform, linking to legislation, etc. It's a bit tricky to articulate, as the problems themselves are still being defined, but it's fair to assume that the alternative - a big blob of HTML, with `data-*` attributes on the relevant parts - probably isn't that friendly for downstream consumers to work with. `details.content[]` is a step in the right direction, setting some basic foundations we can iterate quite quickly when we need to, rather than the current architecture that is quite limiting in what options we have.

We've also been moving in this technical direction for a while, aiming to [remove Govspeak rendering from Whitehall](https://gov-uk.atlassian.net/browse/WHIT-1473), partly for a cleaner, simpler architecture where it's not possible to introduce, say, Whitehall-specific Govspeak rules. This technical simplification is being articulated on the other end of the stack too - draft [RFC-194](https://github.com/alphagov/govuk-rfcs/pull/194) proposes moving all Govspeak rendering to the Frontend (avoids the need to republish documents after a change to Govspeak, etc) - which I'm in favour of.

I'll try to include some of this context in the RFC to strengthen the case for the change 👍

ChrisBAshton:
Thanks for the detailed response, @danhowarthgov.

I'm happy to include in "Beyond" that a DITA-based approach could be followed instead - or I can omit the "Beyond" section entirely if it helps. Ultimately I do want to keep the 'Beyond' section out of scope for this RFC.

From my understanding, I think the move to a `details.content[]` shape actually helps to enable a move to DITA, compared to what we can do today. Whitehall could pull in separate components of content, rather like the 'standalone content blocks' Ed mentions in [this thread](https://github.com/alphagov/govuk-rfcs/pull/197#discussion_r3959682816).

Noted on the reservations about the page-builder approach. I'll loop @leimdorfer in for a product view on Whitehall, but I think he'd agree we wouldn't proceed with WYSIWYG/Gutenberg or DITA approach without user research.

As per your suggestion, I'll add a "Consequences" section to make it clear where we'd be by the end of the RFC's implementation.

danhowarthgov:
> Thanks for the detailed response, @danhowarthgov.
> 
> I'm happy to include in "Beyond" that a DITA-based approach could be followed instead - or I can omit the "Beyond" section entirely if it helps. Ultimately I do want to keep the 'Beyond' section out of scope for this RFC.
> 
> From my understanding, I think the move to a `details.content[]` shape actually helps to enable a move to DITA, compared to what we can do today. Whitehall could pull in separate components of content, rather like the 'standalone content blocks' Ed mentions in [this thread](https://github.com/alphagov/govuk-rfcs/pull/197#discussion_r3959682816).
> 
> Noted on the reservations about the page-builder approach. I'll loop @leimdorfer in for a product view on Whitehall, but I think he'd agree we wouldn't proceed with WYSIWYG/Gutenberg or DITA approach without user research.
> 
> As per your suggestion, I'll add a "Consequences" section to make it clear where we'd be by the end of the RFC's implementation.

Brilliant, thanks @ChrisBAshton. Definitely keep the 'beyond' section in, but be great to have the DITA / CCMS mention as a counterpoint. And will be ideal to add a consequences section. Thank you and have a top weekend!

rfc-197-array-of-content-in-details.md
ChrisBAshton: Good shout. Have pushed up a commit to that effect. 39a0dbcabc4e21064f776faf75808b1a4306256a

rfc-197-array-of-content-in-details.md
ChrisBAshton: Have now pushed up new commits, proposing:

- "Section" - how we refer to each element in the array
- `details.content_sections[]` - the name of the array. It's more specific/searchable than `details.content` and more descriptive than `details.main_content`, and helps reinforce the "section" vocabulary.

KludgeKML: Happy with `content_sections` from a Frontend point of view. Seems like a reasonable choice, and the only thing it would step on would be flexible sections - but they as a concept are being removed now, leaving sections relatively unique.

KludgeKML (approved):
LGTM!

leenagupte (approved):
This all sounds good to me 👍 
Also, `content_sections` sounds like a reasonable compromise on the name of the array.

brucebolt (approved):
I'm happy with the changes following my comments, and the RFC overall.

